PR #100, opened 2014-09-17
Adjust archiving path for Linux

ermshiperete:
On Linux we should use /var/lib instead of /usr/share (which is the
default for CommonApplicationData). Additionally, /var/lib might not
be writable. This change tries to use /var/lib/SIL. If that folder
can't be created, we use a subfolder in ApplicationData instead
(default: $HOME/.config/SIL).


cambell-prince:
This PR has been around for a while now. @neilmayhew or @ermshiperete we should either merge it or close it.  As far as the change goes, LGTM.  But as to whether this change is necessary I shall defer to either of you.

PR #162, opened 2015-01-01
Add strong name project configurations

ddaspit:
This commit adds separate strong name configurations, so that the Palaso libraries can be built with or without strong naming.


JohnThomson:
There seem to be massive changes, most of which I think are probably spurious things done by Visual Studio. Can you explain more the purpose of this? I thought JohnH had already set up a way to build with and without strong names. Also, can you point us to what (significantly) actually changed?

hatton:
> I thought JohnH had already set up a way to build with and without strong
names.

I only did a bandaid: a branch that didn't have the strongname changes.
Seems Damien is making a way to control it at the build level, which is
ideal.

On Mon, Jan 5, 2015 at 8:25 AM, John Thomson <notifications@github.com>
wrote:

> There seem to be massive changes, most of which I think are probably
> spurious things done by Visual Studio. Can you explain more the purpose of
> this? I thought JohnH had already set up a way to build with and without
> strong names. Also, can you point us to what (significantly) actually
> changed?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sillsdev/libpalaso/pull/162#issuecomment-68722705>.
>



-- 

John Hatton

Senior Software Engineer/Program Manager
Language Software Development
SIL International

PR #163, opened 2015-01-07
Fix hintpath for nunit

davyLSDev:


davyLSDev:
I was unable to build on Mac and Linux most recently until I made this change.

PR #166, opened 2015-01-20
BL-911 HACK line appears on terminal

davidmoore1:
Removing console line for resizing
ArtOfReading chooser size


davyLSDev:
LGTM, it is always nice to be able to pull out Debug messages.

PR #168, opened 2015-01-21
Fix file replace

ermshiperete:
Allow to call ReplaceFileWithUserInteractionIfNeeded() if the destination file doesn't exist.

Also two additional changes/fixes that I noticed while debugging this issue.

### DIFF
--- a/Palaso/IO/TempFile.cs
+++ b/Palaso/IO/TempFile.cs
@@ -22,7 +22,7 @@ namespace Palaso.IO
 
 		public TempFile()
 		{
-			 _path = System.IO.Path.GetTempFileName();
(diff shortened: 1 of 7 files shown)


Palaso/IO/TempFile.cs
```diff
@@ -86,8 +86,11 @@ public static TempFile TrackExisting(string path)
 
 		public static TempFile CreateAndGetPathButDontMakeTheFile()
 		{
-			TempFile t = new TempFile();
-			File.Delete(t.Path);
+			// it's safer to use GetRandomFileName here because otherwise we might end up with
```
hatton: I never knew there was a GetRandomFileName!

PR #170, opened 2015-01-22
BL-907: Art of reading image chooser crashing

phillip-hopper:
It seems to be an out-of-memory problem. Displaying thumbnails instead of the full-size image fixes the problem.

### DIFF
--- a/PalasoUIWindowsForms.GeckoFxWebBrowserAdapter/WebThumbnailViewer.cs
+++ b/PalasoUIWindowsForms.GeckoFxWebBrowserAdapter/WebThumbnailViewer.cs
@@ -1,12 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.IO;
 using System.Text;
 using System.Windows.Forms;
 using Palaso.IO;
 using Palaso.UI.WindowsForms.HtmlBrowser;
 
+// ReSharper disable once CheckNamespace
 namespace Palaso.UI.WindowsForms.ImageGallery
 {
 	/// <summary>
@@ -15,9 +17,10 @@ namespace Palaso.UI.WindowsForms.ImageGallery
 	/// </summary>
 	public partial class WebThumbnailViewer : UserControl, IThumbnailViewer
 	{
-		private XWebBrowser _browser;
+		private readonly XWebBrowser _browser;
 		public event EventHandler LoadComplete;
 		private TempFile _htmlFile;
+		private string _tempDirectoryName;
 
 		public event EventHandler SelectedIndexChanged;
 
@@ -25,6 +28,7 @@ namespace Palaso.UI.WindowsForms.ImageGallery
 		{
 			CanLoad = false;
 			ThumbBorderColor = Color.Wheat;
+			ThumbBackgroundColor = Color.White;
 			ThumbNailSize = 95;
 			InitializeComponent();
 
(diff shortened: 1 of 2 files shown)


PalasoUIWindowsForms/ImageGallery/ImageUtilities.cs
```diff
@@ -24,58 +23,48 @@ public static Image GetThumbNail(string imagePath, int destinationWidth, int des
...
 
 			return retBmp;
-#endif
 		}
 
-		private static bool callbackOnAbort()
+		public static Image GetThumbNail(string imagePath, int destinationWidth, int destinationHeight, Color borderColor)
 		{
-			return false;
+			return GetThumbNail(imagePath, destinationWidth, destinationHeight, borderColor, Color.Black);
```
ermshiperete: Wouldn't it be better to use Color.White?

PalasoUIWindowsForms.GeckoFxWebBrowserAdapter/WebThumbnailViewer.cs
```diff
@@ -130,17 +144,30 @@ public void LoadItems(IEnumerable<string> pathList)
...
-				_htmlFile = TempFile.WithExtension("htm");
 			File.WriteAllText(_htmlFile.Path, sb.ToString(), Encoding.UTF8);
 			_browser.Navigate(_htmlFile.Path);
 		}
 
+		//BL-907: Crashes, probably out-of-memory error
+		private string GetThumbnailFile(string fileName)
+		{
+			var cnt = Directory.GetFiles(_tempDirectoryName).Length;
+			var tempFileName = Path.Combine(_tempDirectoryName, "image" + cnt + ".jpg");
```
ermshiperete: Theoretically it's thinkable that this method gets called on multiple threads, so it might be safer to use a random filename.

phillip-hopper: The temp directory name is random, is that sufficient?

PalasoUIWindowsForms/ImageGallery/ImageUtilities.cs
phillip-hopper: Maybe, but I am trying to not change the previous behavior of the method. The reason I added the overload with background color is because the thumbnails were being created with a black background.

PalasoUIWindowsForms.GeckoFxWebBrowserAdapter/WebThumbnailViewer.cs
ermshiperete: I don't think so, because the directory name is stored in a member variable, so multiple threads would use the same random temp directory. This could lead to both threads creating the same file (for a different image).

PalasoUIWindowsForms/ImageGallery/ImageUtilities.cs
ermshiperete: Ahh, ok. Makes sense :-)

phillip-hopper: I added a comment to the code explaining this.

PalasoUIWindowsForms.GeckoFxWebBrowserAdapter/WebThumbnailViewer.cs
phillip-hopper: Updated.

PR #171, opened 2015-01-28
Modify hint path for L10nSharp in the Archiving assembly

jasonleenaylor:
The Archiving dll was referencing a specific L10NSharp found in a packages subfolder which makes me think it was using NuGet (or intending to). However the build on TeamCity doesn't seem to be updating this to the latest which causes a runtime problem for FLEx since the signature of a method was updated.

### DIFF
--- a/SIL.Archiving/SIL.Archiving.csproj
+++ b/SIL.Archiving/SIL.Archiving.csproj
@@ -68,9 +68,6 @@
 	<Reference Include="Ionic.Zip">
 	  <HintPath>..\lib\Ionic.Zip.dll</HintPath>
 	</Reference>
-	<Reference Include="L10NSharp">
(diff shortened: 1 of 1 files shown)


SIL.Archiving/SIL.Archiving.csproj
```diff
@@ -81,6 +78,10 @@
 	<Reference Include="Microsoft.CSharp" />
 	<Reference Include="System.Data" />
 	<Reference Include="System.Xml" />
+    <Reference Include="L10NSharp, Version=1.1.3.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\lib\Debug\L10NSharp.dll</HintPath>
```
papeh: Should this ever by ..\lib\Release?  If not, LGTM :+1:

jasonleenaylor:
Hasso had a good point. I've now addressed it.

PR #172, opened 2015-01-29
Recent change for Linux made AoR slow on Windows

phillip-hopper:


JohnThomson:
Do we still have a speed problem on Linux?

PalasoUIWindowsForms.GeckoFxWebBrowserAdapter/WebThumbnailViewer.cs
```diff
@@ -133,7 +133,16 @@ public void LoadItems(IEnumerable<string> pathList)
...
 				// the data-echo attribute is part of lazy loading for very long lists (see below).
-				sb.AppendLine("<img class='image' src='' data-echo='file://" + new Uri(GetThumbnailFile(path)).AbsolutePath + "'>");
+
+				if (PlatformUtilities.Platform.IsWindows)
+				{
+					sb.AppendLine("<img class='image' src='' data-echo='file://" + htmlPath + "'>");
+				}
+				else
+				{
+					sb.AppendLine("<img class='image' src='' data-echo='file://" + new Uri(GetThumbnailFile(path)).AbsolutePath + "'>");
```
ermshiperete: I'm wondering if we could enhance EnhancedImageServer so that we don't have to render all thumbnails immediately but instead when the browser wants to display one?

JohnThomson: It would be, at best, tricky to apply that idea in this case, because the image server is Bloom code, but this AOR code is shared by other libpalaso clients.

PR #173, opened 2015-01-30
BL-966 AoR instructions are truncated

davidmoore1:
After applying the text, the size of the text
box causes the instructions to be truncated.
The code change adjusts the size leaving more
than enough vertical space to contain the
string, even if it is resized.


ermshiperete:
Do these changes have to be Linux specific, or could we use the same values on Windows as well?

davidmoore1:
Windows does not truncate these fields.  Text is sized correctly there and the adjustment is unnecessary.  I'd prefer not to fix it when it isn't broken.

davidmoore1:
Modified to remove Linux only changes after verifying that the change didn't have a negative effect on the Windows code.

PR #175, opened 2015-02-06
Stop saving metadata back into the original file

hatton:
In BL-1001(bloom) we found that we were actually corrupting someone's original tiff file. Now that we are using taglibsharp instead of the "industry standard" exiftool (which wasn't unicode safe), we realize that it's just too risk to let taglib sharp touch people's original files.

### DIFF
--- a/PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
+++ b/PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
@@ -186,7 +186,8 @@ namespace Palaso.UI.WindowsForms.ImageToolbox
 				throw new NotImplementedException();
 			}
 		}
-			
+#if ifWeEverWantToSaveToOriginalAgain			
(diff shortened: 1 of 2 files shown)


PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
```diff
@@ -186,7 +186,8 @@ private static ImageFormat GetImageFormatForExtension(string fileExtension)
 				throw new NotImplementedException();
 			}
 		}
-			
+#if ifWeEverWantToSaveToOriginalAgain			
+		We're not doing this anymore, too risky. See https://jira.sil.org/browse/BL-1001 
 		/// <summary>
 		/// If you edit the metadata, call this. If it happens to have an actual file associated, it will save it.
 		/// If not (e.g. the image came from a scanner), it won't do anything.
```
JohnThomson: Should we actually delete the method, or change it so it throws?

PR #183, opened 2015-02-10
Show license descriptions in languages other than English

hatton:


PalasoUIWindowsForms.Tests/ClearShare/CreativeCommonsLicenseTests.cs
```diff
@@ -112,6 +112,22 @@ public void HasChanges_CanToggle()
...
+		[Test]
+		public void GetDescription_DontHaveFirstChoice_GivesEnglish()
+		{
+			var l = new CreativeCommonsLicense(true, true, CreativeCommonsLicense.DerivativeRules.DerivativesWithShareAndShareAlike);
+			string usedLanguage;
+			//Assert.AreEqual(l.GetDescription("en"), l.GetDescription(new string[] { "qx", "en" }, out usedLanguage));
+			l.GetDescription(new string[] {"qx", "en"}, out usedLanguage);
+			Assert.AreEqual("en", usedLanguage);
+
+		}
```
JohnThomson: Seems as if there should be a test that it doesn't ALWAYS give English

PalasoUIWindowsForms/ClearShare/CreativeCommonsLicense.cs
```diff
@@ -193,32 +194,73 @@ public override string Token
...
+					return englishText;
+				}
+				//otherwise, see if we have a translation
+				if (LocalizationManager.GetIsStringAvailableForLangId(id, targetLanguage))
+				{
+					idOfLanguageUsed = targetLanguage;
+					return LocalizationManager.GetDynamicStringOrEnglish("Palaso", id, englishText, comment, targetLanguage);
+				}
+			}
+			idOfLanguageUsed = string.Empty;
```
JohnThomson: Wouldn't we generally prefer to return the original English text, if we don't have it in ANY language? Of course maybe 'en' is always in the list of target languages...

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
```diff
@@ -73,9 +76,46 @@ public virtual bool EditingAllowed
...
+			return GetDescription(new string[] {iso6393LanguageCode, "en"}, out idOfLanguageUsed);
+		}
+
+		/// <summary>
+		/// Get a simple, non-legal summary of the license, using the "best" language for which we can find a translation.
+		/// </summary>
+		/// <param name="languagePriorityIds"></param>
+		/// <param name="idOfLanguageUsed">The id of the language we were able to use. Unreliable if we had to use a mix of languages.</param>
+		/// <returns>The description of the license.</returns>
+		public override string GetDescription(IEnumerable<string> languagePriorityIds, out string idOfLanguageUsed)
```
JohnThomson: Looks very like GetComponentOfLicenseInBestLanguage. Can we abstract some of this logic into a base class method?

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
```diff
@@ -109,13 +149,29 @@ public override string ToString()
...
-		public override string GetDescription(string iso639_3LanguageCode)
+		/// <summary>
+		/// Currently, we don't even know the language of custom license strings, so this implementation just gives you the string it has
+		/// </summary>
+		/// <returns></returns>
+		public override string GetDescription(string iso6393LanguageCode)
 		{
 			if (string.IsNullOrEmpty(RightsStatement))
 				return "For permission to reuse, contact the copyright holder.";
 			return "";
```
JohnThomson: Shouldn't this be "return RightsStatement"?

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
```diff
@@ -109,13 +149,29 @@ public override string ToString()
...
 
-		public override string GetDescription(string iso639_3LanguageCode)
+		/// <summary>
+		/// Currently, we don't even know the language of custom license strings, so this implementation just gives you the string it has
+		/// </summary>
+		/// <returns></returns>
+		public override string GetDescription(string iso6393LanguageCode)
 		{
 			if (string.IsNullOrEmpty(RightsStatement))
 				return "For permission to reuse, contact the copyright holder.";
```
JohnThomson: It could make sense not to localize this, since you can't localize a user-supplied rights statement...but OTOH you COULD localize this...

PalasoUIWindowsForms.Tests/ClearShare/CreativeCommonsLicenseTests.cs
JohnThomson: Also perhaps it would be useful to have a test demonstrating (if it is intentional) that you'd better include English in the list, otherwise you may well get the "missing ID" result.

hatton: I agree about the other test idea. Unfortunately I think it would take about a day to modify l10nsharp to be unit-test friendly. Sigh.

PalasoUIWindowsForms/ClearShare/CreativeCommonsLicense.cs
hatton: Yes, that's really up to the client to deal with. Since the client already has to come up with a prioritized list of languages, it seems like it complicates this API to add "and by the way, we always add English".

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
hatton: Consider it done!

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
hatton: Yup!

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
hatton: Done

PalasoUIWindowsForms/ClearShare/LicenseInfo.cs
```diff
@@ -69,13 +71,46 @@ public virtual bool EditingAllowed
...
+		protected virtual string GetBestLicenseTranslation(string idSuffix, string englishText, string comment,
+			IEnumerable<string> languagePriorityIds, out string idOfLanguageUsed)
+		{
+			idSuffix = "MetadataDisplay.Licenses."+idSuffix;
+			foreach (var targetLanguage in languagePriorityIds)
+			{
+				if (targetLanguage == "en")
+				{
+					//do the query to make sure the string is there to be translated someday
+					var unused = LocalizationManager.GetDynamicString("Palaso", idSuffix, englishText, comment);
```
JohnThomson: Is L10NSharp going to do something bad if we ask it for an English translation?

I can imagine someone using the localization mechanism (don't know whether it would work, admittedly) to correct typos in the built-in strings, or adapt them to basic English.

I could also imagine someone using this library in a program basically created in French or Russian, where the fall-back-if-there-is-no-localization actually isn't English.

YAGNI, I guess...but I don't see how it helps to have the special case for English, unless L10NSharp does something silly when asked for English.

PR #184, opened 2015-02-13
Refactor msbuild project files

ermshiperete:
Refactor msbuild project files and enable NuGet package restore (especially for unit tests)

This will make it easier to get builds that work the same on TC, Jenkins, and the developer's machine.


chrisvire:
This will break wesay packaging.  wesay/build/make-source-package.sh calls build.mono.proj explicitly for libpalaso and chorus.

ermshiperete:
Good to know that the SourcePackage target is needed by wesay.  I'll re-submit with that target added and make a note that there are some changes in wesay necessary before build.* gets removed.

I don't think NuGet will affect building wesay: restoring packages doesn't introduce new dependencies and currently affects only the unit tests.

chrisvire:
This works for wesay package building (along with chorus pull request and wesay changes at gerrit).  Wow, that was a lot of work.

chrisvire:
@ermshiperete, can you answer @neilmayhew's question?

ermshiperete:
@neilmayhew wrote:
> I'm still confused about the packages.config files. Some of them bring in both NUnit and NUnit.Runners.Net4, and some don't. When they are both brought in, they're different versions. Maybe if I knew more about NUnit and NUGet it would make more sense, but as it stands I find it very confusing.

Thanks for your persistence :smiley:  This made me re-read how this works in nuget and I found a different solution that requires less changes in the .csproj files: we have to deal with two cases: 1) a developer builds in VS or MD. Both of these tools deal with the package restore, so we don't need special code for that. 2) building from the command line with xbuild. I slightly modified NuGet.targets and Palaso.proj so that before a build we now restore all of the packages of the solution.

I also realized that I was the person who created NUnit.Runners.Net4 :smile:, so I updated that package as well.

(Previously, all the packages.config files that belong to a C# project had both NUnit and NUnit.Runners.Net4 so that when you build a project it also installs nunit-console so that you can run the tests. The solution level packages.config file (.nuget/packages.config) only had NUnit.Runners.Net4 because that's a solution level nuget package (having it in the project's packages.config files was a hack). NUnit and NUnit.Runners.Net4 had different versions because the latter hadn't been updated  because I overlooked the fact that it was me who created it)

ermshiperete:
One question that now comes up is if we should strip down NuGet.targets and remove the things we no longer need.

ermshiperete:
By the way, I haven't tested my latest version on Windows yet. Test should be:

- remove $(RootDir)/packages directory and $(RootDir)/output, $(RootDir)/.nuget/NuGet.exe
- build solution in VS
- remove the above directories again
- build from command line in `build` subdirectory: `msbuild /t:Compile Palaso.proj`

After each build verify that NUnit runner directory got created in $(RootDir)/packages, and that there is a nunit.framework.dll assembly in $(RootDir)/output directory.

ermshiperete:
Turns out we can't have a file NuGet.targets in the .nuget subdirectory because that causes VS to use the old restore style where each project file includes the NuGet.targets file and calls restore at the beginning of the build. Therefore I moved it to build/ and also removed the unnecessary code from it.

Tested on Windows and Linux - works on both command line (msbuild/xbuild) as well as VS/MD.

neilmayhew:
:+1:

chrisvire:
:+1: 

WeSay package builds have been fixed to work with this change.  Chorus was also changed to match.

https://github.com/sillsdev/chorus/pull/37
https://gerrit.lsdev.sil.org/#/c/2888/
https://gerrit.lsdev.sil.org/#/c/2892/

PR #185, opened 2015-02-13
Ignore tests if RAMP is not installed

ermshiperete:
Some tests have the "RampRequired" category. This change
checks if RAMP is installed. If not automatically ignores
the tests that require RAMP.

Change-Id: Iddc2835e129fbc600cc01c4bdb799671ef8cbe45


JohnThomson:
Will SOME automatically-run test (e.g., of RAMP itself) actually verify that these tests don't get broken?

ermshiperete:
Probably not unless some build machine has RAMP installed (and actually not even then because all those tests are in the SkipOnTeamCity category which is excluded on TC). However, this change doesn't make things worse. The benefit of this change is that it's easier for a developer to get the same test results as on TC.

JohnThomson:
Fair enough. I was hoping that RAMP itself was being built on TC.

PR #187, opened 2015-02-19
Null check for default keyboard in WSFontControl (PG-129)

andrew-polk:


andrew-polk:
Thanks, Davy.  I don't have write access to libpalaso.

PR #189, opened 2015-02-24
Change platform-specific properties to support Mac

chrisvire:
* Add Platform.IsMac and Platform.IsUnix (and Unit Tests)
* Change Platform.IsLinux to return true for ONLY Linux (this should
have been Platform.IsUnix)
* Change implementation that used Platform.IsLinux to use
  Platform.IsUnix (not used in chorus, wesay, or fieldworks)
* Change calls to stat to use correct flags for Mac and Linux
* NUnit 2.6.2 added MacOsX to PlatformAttribute's supported
  platforms, however Mono doesn't return MacOsX for
  Environment.OSVersion.Platform so the NUnit implementation
  doesn't work (https://github.com/nunit/nunit/issues/515).
  Therefore UnitTests that run only on Mac have to use a
  compile-time conditional.
  See Palaso.Tests/PlatformUtilities/PlatformTests.cs for
  how system-specific conditionals are defined.

### DIFF
--- a/Palaso.Tests/IO/PathUtilitiesTests.cs
+++ b/Palaso.Tests/IO/PathUtilitiesTests.cs
@@ -17,7 +17,7 @@ namespace Palaso.Tests.IO
 		[TestFixtureSetUp]
 		public void FixtureSetUp()
 		{
-			if (Palaso.PlatformUtilities.Platform.IsLinux)
+			if (Palaso.PlatformUtilities.Platform.IsUnix)
 				TmpAndRootOnDifferentPartitions = StatFile("/tmp") != StatFile("/");
 		}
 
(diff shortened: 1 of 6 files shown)


JohnThomson:
This is scary. Suddenly we have three platforms instead of two. With two options, when there are two ways to do things, it's pretty arbitrary whether you test for Linux, Mono, Unix, or Windows. If we're now trying to support three, seems that at a minimum someone needs to check each usage of each way we test to evaluate whether it should really be one way for Unix and one for everything else, one way for Linux and another for everything else, one way for Windows and one for everything else, one way for Mono and one for everything else (is that the same as IsUnix??) or whether we need three ways of doing it.

You may also need to look for places where we use some kind of #if (typically because the other-platform version does not compile).

davyLSDev:
I like what I see here. All fears aside, I think the distinction between Unices Linux and Mac is a needed one particularly in this spike. Good job on this. (John, rather than us continuing to play fast and loose with our platform definitions, this affords some tightening up of our code. Not a bad thing imho.)

neilmayhew:
@JohnThomson: Usually, the choice will be between Windows and everything else, since everything else is Unix and almost all code will work the same way on all Unix systems. It's very rare that Mac or Linux would be grouped with Windows rather than its Unix companions. Where a distinction between Linux and Mac needs to be made, that would need to be evaluated on a case-by-case basis, and can be added after the code is made to work on Windows and Linux using IsWindows or IsUnix.

It's also very rare that we'd be checking for Mono, since behavioural differences stem almost entirely from the platform rather than the runtime. By definition, Mono is designed to operate the same as .Net, and with Mono taking on more and more code from the open-sourced .Net, the distinction will become even less important as time goes on.

Bottom line: use IsWindows or IsUnix at the outset and then add IsMac or IsLinux afterwards if it becomes necessary.

neilmayhew:
:+1:

PR #190, opened 2015-02-27
Keyman 9: Improve input method change detection (LT-15826)

ermshiperete:
Implement more reliable way of notifying client about input method change.
Depending on the Windows messages doesn't work reliably when input
methods like Keyman or IMEs come into play. These input methods don't
necessarily have a unique language identifier which makes it impossible to
detect the correct input method that is currenlty active.

This change implements notification of input method changes the TSF way
and thus works reliably even with input methods. For older Windows
versions we fall back to Windows messages.

Thanks to Marc Durdin for the analysis of the problem.

### DIFF
--- a/PalasoUIWindowsForms.Tests/Keyboarding/WindowsKeyboardControllerTests.cs
+++ b/PalasoUIWindowsForms.Tests/Keyboarding/WindowsKeyboardControllerTests.cs
@@ -1,4 +1,4 @@
-#if !MONO
+#if !__MonoCS__
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -182,17 +182,6 @@ namespace PalasoUIWindowsForms.Tests.Keyboarding
 			Assert.Greater(keyboards.Count(), 1, "This test requires that the Windows IME has at least two languages installed.");
 
 			Assert.That(keyboards.Select(keyboard => ((KeyboardDescription)keyboard).Engine), Is.All.TypeOf<WinKeyboardAdaptor>());
-		}
-
-		[Test]
-		public void CheckWindowsAssumptions()
-		{
-			// For Windows we expect to have exactly one keyboard adaptor. If we implement
-			// additional ones, e.g. for Keyman, we might need to change some methods, e.g.
-			// ActivateDefaultKeyboard, DefaultForWritingSystem and CreateKeyboardDefinition.
-			Assert.That(KeyboardController.Adaptors.Length, Is.EqualTo(1));
-			Assert.That(KeyboardController.Adaptors.Select(adaptor => adaptor.Type),
-				Has.All.EqualTo(KeyboardType.System));
 		}
 
 		[Test]
--- a/PalasoUIWindowsForms/Keyboarding/Windows/WinKeyboardAdaptor.cs
+++ b/PalasoUIWindowsForms/Keyboarding/Windows/WinKeyboardAdaptor.cs
@@ -1,5 +1,6 @@
-// Copyright (c) 2013 SIL International
+// Copyright (c) 2013-2015 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+
 #if !__MonoCS__
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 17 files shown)


PalasoUIWindowsForms/Keyboarding/Windows/WinKeyboardAdaptor.cs
```diff
@@ -48,14 +49,124 @@ public LayoutName(string layout, string localizedLayout)
...
+				_keyboardAdaptor = keyboardAdaptor;
+			}
+
+			#region ITfLanguageProfileNotifySink Members
+
+			public bool OnLanguageChange(ushort langid)
+			{
+				// In my tests we never hit this method (Windows 8.1). I don't know if the
+				// method signature is wrong or why that is.
+				return true;
```
JohnThomson: It might be helpful (though presumably there's doc of the interface somewhere) to comment what returning true indicates.

PalasoUIWindowsForms/Keyboarding/Windows/WinKeyboardAdaptor.cs
```diff
@@ -48,14 +49,124 @@ public LayoutName(string layout, string localizedLayout)
...
+					profile1.CatId == profile2.CatId &&
+					profile1.HklSubstitute == profile2.HklSubstitute &&
+					profile1.Caps == profile2.Caps &&
+					profile1.Hkl == profile2.Hkl;
+			}
+			#endregion
+
+			#region Fallback if TSF isn't available
+			private void OnWindowsMessageInputLanguageChanged(object sender,
+				InputLanguageChangedEventArgs inputLanguageChangedEventArgs)
```
JohnThomson: It's elegant to have a fall-back, but won't we get the standard windows messages even when TSF IS available? Do we need to do something to deliberately ignore them if we are getting the more reliable notifications?

JohnThomson: Never mind, I see you only register this handler if TSF isn't available. Might be worth enhancing the comment to say that if you do a revision.

PalasoUIWindowsForms.Tests/Keyboarding/WindowsKeyboardControllerTests.cs
```diff
@@ -187,12 +187,15 @@ public void WindowsIME_GetKeyboards_GivesSeveralButOnlyWindowsOnes()
...
 		{
-			// For Windows we expect to have exactly one keyboard adaptor. If we implement
-			// additional ones, e.g. for Keyman, we might need to change some methods, e.g.
-			// ActivateDefaultKeyboard, DefaultForWritingSystem and CreateKeyboardDefinition.
-			Assert.That(KeyboardController.Adaptors.Length, Is.EqualTo(1));
-			Assert.That(KeyboardController.Adaptors.Select(adaptor => adaptor.Type),
-				Has.All.EqualTo(KeyboardType.System));
+			// For Windows we expect to have exactly one system keyboard adaptor and one
+			// other adaptor. If we implement additional ones, we might need to change some
+			// methods, e.g. ActivateDefaultKeyboard, DefaultForWritingSystem and
```
JohnThomson: I haven't yet figured out why this is or what the difference is between them.

PalasoUIWindowsForms/Keyboarding/Windows/WinKeyboardAdaptor.cs
ermshiperete: Done

PalasoUIWindowsForms/Keyboarding/Windows/WinKeyboardAdaptor.cs
ermshiperete: Done

PalasoUIWindowsForms.Tests/Keyboarding/WindowsKeyboardControllerTests.cs
ermshiperete: You're right, this doesn't make so much sense anymore. Removed.

PR #191, opened 2015-02-27
Various cleanup

ermshiperete:
- Update copyright information for keyboarding code
- Cleanup KeyboardController interface
- update version numbers (see commit message for reasons)

### DIFF
--- a/Palaso/Properties/AssemblyInfo.cs
+++ b/Palaso/Properties/AssemblyInfo.cs
@@ -32,8 +32,8 @@ using System.Runtime.InteropServices;
 //
 // You can specify all the values or you can default the Revision and Build Numbers
 // by using the '*' as shown below:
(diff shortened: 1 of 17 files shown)


regnrand:
Palaso and Chorus don't put the version number in those AssemblyInfo files. It is done elsewhere and all assemblies are then stamped with the version number. See:

palaso\build\build.common.proj

For details.

So, while your version number changes do no harm, neither do they do any good.

ermshiperete:
It depends on how you build. If you compile in VS/MD the version numbers from AssemblyInfo get used without being updated. That's the case I'm trying to fix here.

For the official builds (i.e. TC) and for building from the command line I agree that this doesn't make any difference.

regnrand:
Ok

chrisvire:
:+1:

PR #192, opened 2015-02-27
Ignore failing tests on TC

ermshiperete:
For whatever reason the tests that require a readonly file fail
on TC. It turns out that we set the file to be readonly, but
mono on TC reports the file always as being writable which causes
the tests to fail. This change works around this problem by
setting the SkipOnTeamCity category.

Change-Id: I5fb252862373838874e307d6e695b0801af37e13

### DIFF
--- a/Palaso.Lift.Tests/Merging/SynchronicMergerTests.cs
+++ b/Palaso.Lift.Tests/Merging/SynchronicMergerTests.cs
@@ -366,6 +366,7 @@ namespace Palaso.Lift.Tests.Merging
 
 
 		[Test]
+		[Category("SkipOnTeamCity")] // on TC the files don't get set to read-only
 		public void ReadOnlyBaseFile_DoesNothing()
 		{
 			string baseFilePath = Path.Combine(this._directory, _baseLiftFileName);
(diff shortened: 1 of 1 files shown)


papeh:
Is there a SkipOnLinux category that would be better, or is that a FieldWorks category? Otherwise, LGTM.

chrisvire:
You can say:

[Platform(Exclude="Mono")]

ermshiperete:
It only happens on TC and I have no idea why - tried all kinds of things. Maybe it's related to the fact that the build agent runs as root - don't know.

I don't think we want to use [Platform(Exclude="Mono")] since we want to run these tests in general, just not on TC.

It's really strange that the tests suddenly started failing after my changes - I don't think I did anything that would cause these failures. The only real change was the update of NUnit.

papeh:
> Maybe it's related to the fact that the build agent runs as root--don't know [ . . . ] I don't think I did anything that would cause these failures. The only real change was the update of NUnit.

Thanks for clarifying.  Seems strange to me, too. One of the two Windows builds succeeded, so it would be nice if we could continue running these tests on that build, but that may be more trouble than it's worth.

ermshiperete:
Modified the TC build agents so that they don't run as root.

PR #193, opened 2015-03-04
Add UnixName BuildTask

chrisvire:
During the mono build, we need to know whether the platform is
Mac (Darwin) or Linux.  $(OS) is set to Unix for these platforms.

### DIFF
--- a/Palaso.MSBuildTasks/UnixName.cs
+++ b/Palaso.MSBuildTasks/UnixName.cs
@@ -1,0 +1,75 @@
+// Copyright (c) 2014 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+// Parts based on code by MJ Hutchinson http://mjhutchinson.com/journal/2010/01/25/integrating_gtk_application_mac
+
+using System;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+
+namespace Palaso.BuildTasks
+{
+	/// <summary>
+	/// Determines the Unix Name of the operating system executing the build.
+	/// This is useful when determining Mac vs Linux during a build.
+	/// On Mac, the output Value will be "Darwin".
+	/// On Linux, the output Value will be "Linux".
+	/// </summary>
+
+	// This can be used to set DefineConstants during the PreBuild Target.
+	// here is an example build/platform.targets file that can be included
+	// in a CSPROJ file.  SYSTEM_MAC or SYSTEM_LINUX will be defined and
+	// can be used in the C# code for #if conditional compilation.
+	// <?xml version="1.0" encoding="utf-8" ?>
+	// <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	//   <UsingTask TaskName="UnixName" AssemblyFile="Palaso.BuildTasks.dll" />
+	//   <Target Name="BeforeBuild">
+	//     <UnixName>
+	//       <Output TaskParameter="Value" PropertyName="UNIX_NAME" />
+	//     </UnixName>
+	//     <PropertyGroup>
+	//       <DefineConstants Condition="'$(OS)' == 'Unix'">$(DefineConstants);SYSTEM_UNIX</DefineConstants>
+	//       <DefineConstants Condition="'$(UNIX_NAME)' == 'Darwin'">$(DefineConstants);SYSTEM_MAC</DefineConstants>
+	//       <DefineConstants Condition="'$(UNIX_NAME)' == 'Linux'">$(DefineConstants);SYSTEM_LINUX</DefineConstants>
+	//     </PropertyGroup>
+	//   </Target>
+	// </Project>
+	public class UnixName : Task
+	{
+		public override bool Execute()
+		{
+			Value = String.Empty;
+			if (Environment.OSVersion.Platform == PlatformID.Unix)
+			{
+				IntPtr buf = IntPtr.Zero;
+				try
+				{
+					buf = System.Runtime.InteropServices.Marshal.AllocHGlobal(8192);
+					// This is a hacktastic way of getting sysname from uname ()
+					if (uname(buf) == 0)
+						Value = System.Runtime.InteropServices.Marshal.PtrToStringAnsi(buf);
+					else
+						Log.LogError("uname failed");
+				}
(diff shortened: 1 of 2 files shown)


Palaso.MSBuildTasks/UnixName.cs
```diff
@@ -0,0 +1,47 @@
...
+		{
+			if (Environment.OSVersion.Platform == PlatformID.Unix)
+			{
+				IntPtr buf = IntPtr.Zero;
+				try
+				{
+					buf = System.Runtime.InteropServices.Marshal.AllocHGlobal(8192);
+					// This is a hacktastic way of getting sysname from uname ()
+					if (uname(buf) == 0)
+						Value = System.Runtime.InteropServices.Marshal.PtrToStringAnsi(buf);
```
StephenMcConnel: I assume uname() NUL-terminates each of the fields in the output struct.  If so, this looks like a suitable hack to obtain the value of the first field (sysname).

StephenMcConnel:
Looks reasonable to me.  +2 as they say...

Palaso.MSBuildTasks/UnixName.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+
+namespace Palaso.BuildTasks
+{
+	public class UnixName : Task
+	{
+		public override bool Execute()
```
ermshiperete: A comment might be helpful that lists the values that can get returned from this task. And a class comment that shows how to use it might also be helpful.

davyLSDev:
LGTM, this seems to be the way forward with our build issue. (Maybe we should let Neil have a look at it before merging?) But I think it should be fine!

Palaso.MSBuildTasks/UnixName.cs
```diff
@@ -0,0 +1,72 @@
...
+	///   <Target Name="BeforeBuild">
+	/// 	<UnixName>
+	///       <Output TaskParameter="Value" PropertyName="UNIX_NAME" />
+	///     </UnixName>
+	/// 	<PropertyGroup>
+	///       <DefineConstants Condition="'$(UNIX_NAME)' == 'Darwin'">$(DefineConstants);SYSTEM_OSX</DefineConstants>
+	///       <DefineConstants Condition="'$(UNIX_NAME)' == 'Linux'">$(DefineConstants);SYSTEM_LINUX</DefineConstants>
+	/// 	</PropertyGroup>
+	///   </Target>
+	/// </Project>k
```
StephenMcConnel: you do realize that all of the XML inside the <summary> header comment element will be probably be garbled by the popup help windows in the IDEs?  but I'm willing to let that slide.

Palaso.MSBuildTasks/UnixName.cs
```diff
@@ -0,0 +1,72 @@
...
+				}
+				catch
+				{
+					Value = String.Empty;
+				}
+				finally
+				{
+					if (buf != IntPtr.Zero)
+						System.Runtime.InteropServices.Marshal.FreeHGlobal(buf);
+				}
```
neilmayhew: I think it would be good to throw an exception if uname fails, and to rethrow any exception that occurs in the try block. Any of these events should be impossible, but if they do occur it indicates a serious problem of some kind. Just silently setting Value = String.Empty will make it very hard for anyone to diagnose (or even spot) a problem.

Also, if uname() fails, won't Value be Null rather than Empty? It would be better to be consistent.

Palaso.MSBuildTasks/UnixName.cs
```diff
@@ -0,0 +1,72 @@
...
+	/// <?xml version="1.0" encoding="utf-8" ?>
+	/// <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	///   <UsingTask TaskName="UnixName" AssemblyFile="Palaso.BuildTasks.dll" />
+	///   <Target Name="BeforeBuild">
+	/// 	<UnixName>
+	///       <Output TaskParameter="Value" PropertyName="UNIX_NAME" />
+	///     </UnixName>
+	/// 	<PropertyGroup>
+	///       <DefineConstants Condition="'$(UNIX_NAME)' == 'Darwin'">$(DefineConstants);SYSTEM_OSX</DefineConstants>
+	///       <DefineConstants Condition="'$(UNIX_NAME)' == 'Linux'">$(DefineConstants);SYSTEM_LINUX</DefineConstants>
```
neilmayhew: [MSDN](https://msdn.microsoft.com/en-us/library/dd633440.aspx#BKMK_String) says you can call any `String` method on a property. So you could have a single `DefineConstants` without a condition using

    $(UNIX_NAME.ToUpperInvariant())

However, this does require .Net >= 4.

Also, I see you've used "OSX" instead of "DARWIN" which is probably more useful for a non-Mac person reading through the code. However, I think I'd prefer "MAC" to "OSX" since that's even more meaningful and the distinction between classic Mac OS and Mac OS X doesn't exist any more.

Palaso.MSBuildTasks/UnixName.cs
chrisvire: OK. I can change the <summary> to be just around the first paragraph and then use <example> around the rest.  Can I put the XML in a <code> element?

Palaso.MSBuildTasks/UnixName.cs
chrisvire: Sorry, this is straight from the example code (and also is in Palaso.PlatformUtilities.Platform.  Whatever we decide on should be in both places.

Palaso.MSBuildTasks/UnixName.cs
chrisvire: Apple refers to its operating system as OSX and to the hardware as Mac.  It dropped MacOSX a while ago.  So for consistency, I could change Palaso.PlatformUtilities.Platform.IsMac to .IsOSX.  Or we could use DARWIN everywhere. :-)

FYI, SYSTEM_OSX is already being used in Palaso.Tests and I need to fix that place.  

http://www.apple.com/osx/

chrisvire:
Neil had given it a :+1: , but emailed me about fixing the modified newline at the end of the csproj file.  I fixed that and it is ready to go.

PR #194, opened 2015-03-04
Update ibusdotnet.dll

ermshiperete:
The new version of ibusdotnet.dll deals with the
missing InputContext.IsEnabled method.

InputContext.IsEnabled is no longer available in IBus 1.5. The IBus
implementation expects it to always be enabled. This change mimics
this behavior by returning true if the IsEnabled method is not
contained in the interface retrieved by introspection.

This fixes LT-15870.

Change-Id: Ia1647f1b3013bf0c512a966d904a6c10ae869505


papeh:
:+1:

PR #195, opened 2015-03-05
Fix Mac build to use UnixName Custom Build Task

chrisvire:
_system_name and _system_type are not available by default in the
environment.  Use new UnixName Custom Build Task to conditionally
build Mac unit tests instead of depending on _system_name.

Define new build/platform.targets that can be used to create
system-specific definitions for conditional compilation.


Palaso.Tests/PlatformUtilities/PlatformTests.cs
```diff
@@ -4,23 +4,19 @@
...
-// 1) include the following PropertyGroup in the .csproj after all of the DefineConstants
-//    elements.  [Note: _system_name and _system_type are defined by xbuild in later 
-//    versions of Mono.  We tested with Mono 3.12 that comes with Xamarin Studio for 
-//    Mac.  To verify, add /verbosity:diag to xbuild command-line. YMMV.]
+// For code shipped with products, please try to use runtime checks to select 
+// code to run.  There are cases that require compile-time conditionals 
+// (e.g. when testing the runtime or referencing platform specific imported 
+// assemblies like MonoMac.dll).  In this case, do The following:
+// 
+// 1) Include the following at the bollow of the csproj file
```
ermshiperete: bollow?

chrisvire: sigh... bottom.  Not sure how that happened.

chrisvire: Actually, this shouldn't be here anymore.  It is documented in platform.targets.

build/platform.targets
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- 
+Include this file at the end of a csproj file to add system-specific defines 
+for conditional compilation. Example: libpalaso/Palaso.Tests/Palaso.Tests.csproj
+These constants will be defined:
+SYSTEM_UNIX - should be compiled on Linux and Mac
+SYSTEM_MAC - should only be compiled on Mac (e.g. Unit Tests specific to Mac)
+SYSTEM_Linix - should only be compiled on Linux
```
chrisvire: Should be SYSTEM_LINUX

PR #197, opened 2015-03-05
BL-1071: Added method to determine if an input processor is active

phillip-hopper:
Bloom needs this in order to determine if it should activate jquery.longpress or not. KeyMan 9 doesn't play well with jquery.longpress.


StephenMcConnel:
Shouldn't there be a changed file that calls the new IsFormUsingInputProcessor(Form frm) method?

StephenMcConnel:
Looks good to me.  :-)

PR #198, opened 2015-03-05
Add processing for LDML delimiter quote level 1 and 2 continuers

andrew-polk:


SIL.WritingSystems/LdmlDataMapper.cs
```diff
@@ -1063,6 +1076,21 @@ private void WriteDelimitersElement(XElement delimitersElem, WritingSystemDefini
...
 			}
 
+			if (qm1 != null)
+			{
+				var level1ContinuerElem = new XElement(Sil + "quotationContinue", qm1.Continue);
+				specialElem = GetOrCreateSpecialElement(delimitersElem);
+				quotationmarksElem = specialElem.GetOrCreateElement(Sil + "quotation-marks");
+				quotationmarksElem.Add(level1ContinuerElem);
+			}
+			if (qm2 != null)
```
darcywong00: alternateQuotationContinue is an optional element, so don't write this if qm2.Continue is empty

PR #199, opened 2015-03-06
Some refactoring of CommandLineRunnerTests

ermshiperete:
- move to correct location and namespace
- add tests for timeout
- enable test that is working on Linux
- add comment why the other existing test fails on Linux

### DIFF
--- a/Palaso.Tests/CommandLineProcessing/CommandLineRunnerTests.cs
+++ b/Palaso.Tests/CommandLineProcessing/CommandLineRunnerTests.cs
@@ -4,20 +4,19 @@ using Palaso.CommandLineProcessing;
 using Palaso.Progress;
 
 
-namespace Palaso.Tests.Progress
+namespace Palaso.Tests.CommandLineProcessing
 {
 	[TestFixture]
 	public class CommandLineRunnerTests
 	{
+		private const string App = "PalasoUIWindowsForms.TestApp.exe";
 
-		[Test, Category("KnownMonoIssue")]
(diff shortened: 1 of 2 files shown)


Palaso.Tests/CommandLineProcessing/CommandLineRunnerTests.cs
```diff
@@ -0,0 +1,60 @@
...
+		[Test, Category("KnownMonoIssue")]
+		[Platform(Exclude="Linux", Reason = "Test has problems on Mono")]
+		public void CommandWith10Line_CallbackOption_Get10LinesAsynchronously()
+		{
+			var progress = new StringBuilderProgress();
+			int linesReceivedAsynchronously = 0;
+			CommandLineRunner.Run(App, "CommandLineRunnerTest", null, string.Empty, 100,
+				progress, s => ++linesReceivedAsynchronously);
+			// The test fails on Linux because progress gets called 10x for StdOutput plus
+			// 1x for StdError (probably on the closing of the stream), so linesReceivedAsync is 11.
```
papeh: Interesting.  Could a callback that checks `sender` fix this?

ermshiperete: I don't think that's possible because we don't have a sender: The method that gets called takes a single string argument, so we don't know who called us. Or put differently: CommandLineRunner.Run takes a `Action<string>` argument which when called doesn't tell the caller/sender.

PR #200, opened 2015-03-11
Add FileDialogAdapter to PalasoUIWindowsForms.dll

StephenMcConnel:
The code is being moved from SIL.Utils in the FieldWorks project to
make it available to other programs.  This does add references to
gtk-sharp, glib-sharp, and atk-sharp on Linux.


StephenMcConnel:
It might be better to move the entire SilUtils assembly code to Palaso either as a block (add assembly as is) or piecemeal (split it up among existing Palaso assemblies).  This requires a decision before we decide on this particular pull request.

JohnThomson:
Let's hold off on merging this until Steve finishes his research and determines whether it is needed as a part of the full solution (or until we decide to go ahead and move all of SilUtils).

StephenMcConnel:
This isn't needed for the bugfix I was working on.  It may be a good idea, but is part of a bigger issue.

PR #201, opened 2015-03-17
Remove quotationContinue and alternateQuotationContinue before processing

andrew-polk:
This follows the same pattern as the other quotation-marks elements.


papeh:
:+1:

JohnThomson:
I don't know much about this area, but the new removals differ from the first one (and contradict the comment above the block) in that they are not restricted to removing things that are blank or have the narrative attribute.

darcywong00:
Except for the last one, all of these children of sil:quotation-marks should be removed because the writer will repopulate them based on the writing system definition.
```
<sil:quotation-marks>
    <sil:quotation />
    <sil:quotation type="narrative" />
    <sil:quotationContinue />
    <sil:alternateQuotationContinue />
    <sil:quotation type="anythingButBlankOrNarrative" />
</sil:quotation-marks>
```

I'll go ahead and merge this pull request, and then clarify the comment on another commit

PR #202, opened 2015-03-17
Start OpenFileDialog with large icon view (BL-1021)

StephenMcConnel:
Although not explicitly mentioned in the issue, this is part of
making Bloom work the same on both Linux and Windows with
regard to choosing an image file with the file chooser dialog.
(Getting thumbnail images to show is at the Mono library level, this
just uses reflection to get large icons instead of tiny ones.)

### DIFF
--- a/PalasoUIWindowsForms/ImageToolbox/AcquireImageControl.cs
+++ b/PalasoUIWindowsForms/ImageToolbox/AcquireImageControl.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
+using System.Reflection;
 using System.Windows.Forms;
 using L10NSharp;
 using Palaso.IO;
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
It seems that you could write a unit test verifying that the methods you are looking up are present, to avoid the 'silently stops working' problem. It seems you would want to know in our builds at least

PalasoUIWindowsForms/ImageToolbox/AcquireImageControl.cs
```diff
@@ -46,36 +47,10 @@ void _galleryControl_ImageChanged(object sender, EventArgs e)
...
-//			}
-//#else
 #if MONO
 			using (var dlg = new OpenFileDialog())
 #else
-			//The primary thing this OpenFileDialogEx buys us is that with the standard one, there's
+			//The primary thing this OpenFileDialogWithViews buys us is that with the standard one, there's
 			//no way pre-set, what "view" the user gets. With the standard dialog,
 			//we had complaints that a user had to change the view to show icons *each time* they used this.
 			//Now, OpenFileDialogWithViews still doesn't let us read (and thus remember) the selected view.
```
papeh: this comment could be further improved.  how about

    //The primary thing this OpenFileDialogWithViews buys us is the ability to default to large icons.

PR #203, opened 2015-03-18
Don't crash when language associated with Keyman keyboard changes

ermshiperete:
When a user opens the Keyman configuration dialog from the Keyboard tab
of the WS properties dialog and changes the language associated with a
Keyman keyboard we used to crash (LT-16172). This fix is similar the the
one for LT-15333 only that we get a different exception in this case.


papeh:
COMException seems like a broad category to suppress, but I can't think how to improve this.

PR #204, opened 2015-03-18
BL-1104: Out-of-memory errors cropping large images

phillip-hopper:


PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.cs
```diff
@@ -163,24 +158,27 @@ private ListViewItem AddControl(string label, Bitmap bitmap, string imageKey, Sy
...
 
-			if (_inIndexChanging)
+			if (_previousSelectedIndex == _toolListView.SelectedIndices[0])
 				return;
 
 			try
 			{
-				_inIndexChanging = true;
-
-				if (_toolListView.SelectedItems.Count == 0)
```
papeh: if `SelectedItems.Count == 0` is still a possibility (e.g. when there are no items available to select), an exception will be thrown outside the try block.

PalasoUIWindowsForms/ImageToolbox/Cropping/ImageCropper.cs
```diff
@@ -217,45 +242,23 @@ private void ImageCropper_Resize(object sender, EventArgs e)
...
 
-		private float GetImageToCanvasScaleFactor()
+		private float GetImageToCanvasScaleFactor(Image img)
 		{
 			var availArea = new Rectangle(BorderSize, BorderSize, Width - (2*BorderSize), Height - (2*BorderSize));
-			float hProportion = (float) availArea.Width/((float) _originalImage.Width);
-			float vProportion = (float) availArea.Height/((float) _originalImage.Height);
+			float hProportion = (float)availArea.Width / ((float)img.Width);
+			float vProportion = (float)availArea.Height / ((float)img.Height);
 			return Math.Min(hProportion, vProportion);
```
papeh: the above method no longer relies on instance data so could be made static.

PalasoUIWindowsForms/ImageToolbox/Cropping/ImageCropper.cs
```diff
@@ -143,7 +144,32 @@ public PalasoImage Image
...
 				//other code changes the image of this palaso image, at which time the PI disposes of its copy,
 				//so we better keep our own.
-				_originalImage = (Image) value.Image.Clone();
+
+				// save the original in a temp file instead of an Image object to free up memory
+				_savedOriginalImage = TempFile.CreateAndGetPathButDontMakeTheFile();
+				value.Image.Save(_savedOriginalImage.Path, ImageFormat.Png);
+				
+				// make a reasonable sized copy to crop
+				if (value.Image.Width > 1000)
```
papeh: could check height, too.

PalasoUIWindowsForms/ImageToolbox/Cropping/ImageCropper.cs
```diff
@@ -358,33 +359,26 @@ public Image GetCroppedImage()
...
-				CheckForInvalidImage();
-
 				//NB: this worked for tiff and png, but would crash with Out Of Memory for jpegs.
 				//This may be because I closed the stream? THe doc says you have to keep that stream open.
 				//Also, note that this method, too, lost our jpeg encoding:
 				//          return bmp.Clone(selection, _image.PixelFormat);
 				//So now, I first copy it, then clone with the bounds of our crop:
 
-				using (var copy = new Bitmap(_originalImage)) //**** here we lose the jpeg rawimageformat, if it's a jpeg. Grrr.
+				using (var copy = new Bitmap(_savedOriginalImage.Path)) //**** here we lose the jpeg rawimageformat, if it's a jpeg. Grrr.
```
JohnThomson: So the original code ended up copying the original twice, plus the output which might be nearly as big. Both versions have to make a bitmap of the whole original as well as the new bitmap. If the original was a very compressed jpg or sparse line drawing, we may still be using a lot of memory we don't need.
The top answer at http://stackoverflow.com/questions/734930/how-to-crop-an-image-using-c has another approach to copying part of an image into a new bitmap that might be worth considering.

JohnThomson:
One suggestion to consider...otherwise LGTM

PalasoUIWindowsForms/ImageToolbox/Cropping/ImageCropper.cs
phillip-hopper: Line 253 uses the Width and Height properties of the UserControl.

PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.cs
phillip-hopper: Code moved to inside the try-catch block.

PalasoUIWindowsForms/ImageToolbox/Cropping/ImageCropper.cs
phillip-hopper: Now checking both height and width.

PalasoUIWindowsForms/ImageToolbox/Cropping/ImageCropper.cs
phillip-hopper: I couldn't determine if one method is better than the others. It looks like they all call GDI+ to do the actual copying.

papeh:
:+1:

PR #207, opened 2015-03-26
Add checks for Wasta and Cinnamon in PlatformUtilities

StephenMcConnel:
These are needed for fixing https://jira.sil.org/browse/BL-1147.

### DIFF
--- a/Palaso/PlatformUtilities/Platform.cs
+++ b/Palaso/PlatformUtilities/Platform.cs
@@ -11,6 +11,7 @@ namespace Palaso.PlatformUtilities
 		private static readonly string UnixNameLinux = "Linux";
 		private static bool? m_isMono;
 		private static string m_unixName;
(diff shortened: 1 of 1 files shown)


Palaso/PlatformUtilities/Platform.cs
```diff
@@ -76,7 +77,57 @@ private static string UnixName
...
 		}
 
+
+		public static bool IsCinnamon
+		{
+			get
+			{
+				if (!IsUnix)
+					return false;
+				return SessionManager.StartsWith("/usr/bin/cinnamon-session");
```
papeh: `IsUnix && SessionManager.Starts...`

`IsWasta` could be moved up next to this property.

papeh:
You're right, that looks like an interesting way of getting the session manager.  But if you think that's the best way to find out if this `IsCinnamon`, LGTM.

PR #208, opened 2015-03-27
Add mechanism for logging memory state

JohnThomson:


PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -0,0 +1,91 @@
...
+			else if (Platform.IsLinux)
+			{
+				var meminfo = File.ReadAllText("/proc/meminfo");
+				var match = new Regex(@"MemTotal:\s+(\d+) kB").Match(meminfo);
+				if (match.Success)
+				{
+					totalPhysicalMemory = ulong.Parse(match.Groups[1].Value) * 1024;
+				}
+				// So far we have no way to get anything corresponding to Windows's idea of the total virtual memory
+				// the process can use.
```
StephenMcConnel: There are a couple of other lines in /proc/meminfo that may be of interest for getting this value:
SwapTotal:      16729084 kB
and
VmallocTotal:   34359738367 kB
SwapTotal is the total amount of physical swap memory (on disk).  This plus the MemTotal value might be what you want since I think it's the limit of what can actually be allocated by the kernel.
VmallocTotal gives the theoretical virtual memory size as far as I can tell from the sketchy descriptions I've seen.  (I could be wrong.)  Since addresses don't have to be contiguous, theoretically any address in that area could be allocated, but probably very few systems (if any) would be set up to allow all of those addresses to be allocated.  (if that's even possible for any CPU chip in production.)

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -0,0 +1,91 @@
...
+			if (danger && okToBotherUser && !s_warningShown)
+			{
+				s_warningShown = true;
+				var warning = LocalizationManager.GetString("MemoryWarning",
+					"Unfortunately, {0} is starting to get short of memory, and may soon slow down or experience other problems. We recommend that you quit and restart it when convenient.");
+				var caption = LocalizationManager.GetString("Warning", "Warning");
+				MessageBox.Show(null, string.Format(warning, Application.ProductName), caption, MessageBoxButtons.OK,
+					MessageBoxIcon.Warning);
+			}
+			return danger;
```
StephenMcConnel: would it be worth telling the user when/if the memory situation has improved so that he's no longer in danger?  does that even ever happen?

StephenMcConnel:
I have a couple of suggestions to consider, but nothing that should hold up check-in necessarily. :+1:

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
JohnThomson: I think not. Ideally the user should be able to ignore memory issues...we just handle it. Let's wait and see whether someone asks for such a thing, at least.

PR #209, opened 2015-03-31
Fix recent memory check work for Linux

StephenMcConnel:
Also allow higher limit for 64-bit processes with ample physical
memory.  (Currently only Linux supports 64-bit processes for our
code as far as I know.)

### DIFF
--- a/PalasoUIWindowsForms/Reporting/MemoryManagement.cs
+++ b/PalasoUIWindowsForms/Reporting/MemoryManagement.cs
@@ -37,7 +37,9 @@ namespace Palaso.UI.WindowsForms.Reporting
 		public static bool CheckMemory(bool minor, string eventDescription, bool okToBotherUser)
 		{
 			var heapMem = GC.GetTotalMemory(true); // first, as it may reduce other numbers
+			var is64BitProcess = IntPtr.Size == 8; // according to MSDN
 			long memorySize64;
+			long workingSet64;
 			string message;
 			ulong totalPhysicalMemory = 0;
 			string totalVirtualMemory = "unknown";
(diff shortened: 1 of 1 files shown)


PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -37,7 +37,9 @@ public static class MemoryManagement
 		public static bool CheckMemory(bool minor, string eventDescription, bool okToBotherUser)
 		{
 			var heapMem = GC.GetTotalMemory(true); // first, as it may reduce other numbers
+			var is64BitProcess = IntPtr.Size == 8; // according to MSDN
 			long memorySize64;
```
jasonleenaylor: These variable names seem to indicate that they are used only when working with a 64bit process but that doesn't seem to be the case. Is the 64 after memorySize and workingSet correct or helpful?

StephenMcConnel: The 32-bit variants of the properties in the Process class are all marked obsolete.  I suppose retaining the 64 on the names merely reminds us that these are long (or ulong) variables instead of int (or uint) variables.  Since the names mimic the names of the Process properties, you might argue that they're helpful reminders that we've accessed the recommended properties.  In any case, I would argue that the 64 is correct.  Helpful?  I suppose opinions could vary.
Note that the PAE extensions to the chip architecture means that even a 32-bit OS with 32-bit processes can have more than 4GB of RAM available for use even if any one process can only directly access 2GB.

JohnThomson: The trailing 64 for property names in properties of Process makes them longs; there are (deprecated) properties with the shorter names that return ints. I think it's at least somewhat helpful for a variable that is mainly intended to hold the result of something like proc.memorySize64 to be called memorySize64. But could go either way.

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -52,30 +54,49 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
...
 			if (minor)
 				Logger.WriteMinorEvent(message);
 			else
 				Logger.WriteEvent(message);
-			var danger = memorySize64 > 1000000000;
+			long safelimit = 1000000000;
+			if (is64BitProcess && totalPhysicalMemory >= 8192000000L)
+				safelimit = 2000000000;
+			bool danger = false;
+			if (Platform.IsWindows)
```
JohnThomson: This seems to deserve a comment explaining that we've found memorySize64 on Linux to be useless as a measure of RAM in use by the process.

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -52,30 +54,49 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
...
+					totalSwapMemory = ulong.Parse(match2.Groups [1].Value) * 1024;
+				var availableMemory = totalPhysicalMemory + totalSwapMemory;
+				if (is64BitProcess)
+				{
+					totalVirtualMemory = (availableMemory / 1024).ToString ("N0");
+				}
+				else
+				{
+					ulong twoGB = 2147483648L;
+					totalVirtualMemory = ((availableMemory > twoGB ? twoGB : availableMemory) / 1024).ToString("N0");
```
JohnThomson: Do you know for sure that this is the most VM a 32-bit .NET process can address on Linux? Anything up to 4G is conceivable.

StephenMcConnel: As far as I can tell, this is correct.  Some of the comments I've seen on the web indicate that a 32-bit process cannot allocate more than about 1.4GB in a 32-bit process without hitting "Out Of Memory" errors.

PR #210, opened 2015-04-08
Exclude Microsoft.VisualBasic.Devices on Linux

jasonleenaylor:
* Add copyright notice
* Remove unused usings
* Exclude the Microsoft.VisualBasic.Devices using on Linux because
  Mono aggressively pursues every using at runtime and has problems
  if it can't find them, even if the code doesn't need them. Excluding
  avoids the need for every Linux package using palaso to have access
  to it.

### DIFF
--- a/PalasoUIWindowsForms/PalasoUIWindowsForms.csproj
+++ b/PalasoUIWindowsForms/PalasoUIWindowsForms.csproj
@@ -287,7 +287,7 @@
     <None Include="..\lib\common\Enchant.Net.dll.config">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <Reference Include="Microsoft.VisualBasic" />
+    <Reference Include="Microsoft.VisualBasic" Condition="'$(OS)'=='Windows_NT'"/>
     <Reference Include="sysglobl" />
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
--- a/PalasoUIWindowsForms/Reporting/MemoryManagement.cs
+++ b/PalasoUIWindowsForms/Reporting/MemoryManagement.cs
@@ -1,16 +1,17 @@
-﻿using System;
-using System.Collections.Generic;
+﻿// Copyright (c) 2014 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+using System;
 using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Reflection;
-using System.Text;
-using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using L10NSharp;
-using Microsoft.VisualBasic.Devices;
 using Palaso.PlatformUtilities;
 using Palaso.Reporting;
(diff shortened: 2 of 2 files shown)


jasonleenaylor:
@StephenMcConnel Could you check this over?

PalasoUIWindowsForms/PalasoUIWindowsForms.csproj
```diff
@@ -287,7 +287,7 @@
     <None Include="..\lib\common\Enchant.Net.dll.config">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <Reference Include="Microsoft.VisualBasic" />
+    <Reference Include="Microsoft.VisualBasic" Condition=" '$(Configuration)' != 'DebugMono' And '$(Configuration)' != 'ReleaseMono'"/>
```
StephenMcConnel: The condition is usually OS based, not configuration based.  I don't remember the exact syntax off hand.

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -1,16 +1,16 @@
...
-using System.Linq;
-using System.Reflection;
-using System.Text;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using L10NSharp;
-using Microsoft.VisualBasic.Devices;
 using Palaso.PlatformUtilities;
 using Palaso.Reporting;
+#if !MONO // This is not needed on Linux, and can causes extra dependancy trouble if it is compiled in
```
StephenMcConnel: generally prefer __MonoCS__ to MONO

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -45,9 +45,11 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
 			string totalVirtualMemory = "unknown";
 			if (Platform.IsWindows)
 			{
+#if !MONO
```
StephenMcConnel: generally prefer __MonoCS__ to MONO.

I suppose we could dispense with the using statement altogether and attempt reflection here.  (with a comment explaining the issue)  if we do that, it's possible that we don't need the reference in the csproj file, but I'm not sure about that.

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -43,14 +44,12 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
 			string message;
 			ulong totalPhysicalMemory = 0;
 			string totalVirtualMemory = "unknown";
-			if (Platform.IsWindows)
-			{
+
+#if !__MonoCS__ // There are completely different dependencies and code per platform for finding the memory information
 				var computerInfo = new Computer().Info;
 				totalPhysicalMemory = computerInfo.TotalPhysicalMemory;
```
StephenMcConnel: One could fully specify Computer here and avoid the need for the using statement above.  Or one could leave well enough alone.  :-)

StephenMcConnel:
Looks Good Enough To Me.  :+1:

PR #211, opened 2015-04-16
Skip ByHand tests on Team City

phillip-hopper:
These tests, marked "ByHand," are failing most of the time on Team City.

### DIFF
--- a/build/Palaso.proj
+++ b/build/Palaso.proj
@@ -106,7 +106,7 @@
 
 		<NUnitTeamCity
 			Assemblies="@(TestAssemblies)"
-			ExcludeCategory="SkipOnTeamCity$(ExtraExcludeCategories)"
+			ExcludeCategory="SkipOnTeamCity;ByHand$(ExtraExcludeCategories)"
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
Why don't we set "ByHand" as a category of tests to skip in the build instead?

jasonleenaylor:
I can't imagine that we would want to run any ByHand tests on our continuous build server. I think you can add this in the NUnit task in the .proj file(s)

JohnThomson:
I tend to agree that 'ByHand' tests are unlikely to be ones we want to run on a build server. OTOH I can see some benefit in having a single, consistent way of marking all the tests that are NOT run automatically. There do not seem to be so many that putting both flags on is very expensive.

jasonleenaylor:
John, I see your point but I think there are already multiple test categories being skipped by the TeamCity NUnit build tasks. If we want to aim for consistency then we should have it only skip 'SkipOnTeamCity' and add that to all the rest of the tests that are currently skipped a different way.

phillip-hopper:
Pull request updated

PR #214, opened 2015-04-24
Support customization of where Registration settings are stored

JohnThomson:
Also switches Registration settings to be persisted using our
cross-platform class. The first release using this will not upgrade
old registration settings. The only current client is Bloom, which
is OK with this.
Also adds a better algorithm (modify time of config file) for finding
previous configurations to upgrade from.


PalasoUIWindowsForms.Tests/SettingsProviderTests.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace PalasoUIWindowsForms.Tests
+{
+	[TestFixture]
+	public class SettingsProviderTests
+	{
+		[Test]
+		public void SettingsFolderWithNoConfig_SortsAfterOneWithConfig()
+		{
+			// We do two iterations like this because we can't count on deleting the file to really make it gone at once.
+			for (var which = 0; which < 1; which++)
```
jasonleenaylor: This comment made me expect to see a file deletion. I understand the problem you are avoiding but maybe you can re-word this to indicate that we are NOT deleting files, but switching which folder does not have a config.

PalasoUIWindowsForms.Tests/SettingsProviderTests.cs
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		/// <summary>
+		/// This test was primarily created to test that we can override the default location for the RegistrationSettingsProvider.
+		/// However it's also quite a useful test of the whole business of reading settings.
+		/// </summary>
+		[Test]
+		public void CanOverrideDefaultLocation()
+		{
+			RegistrationSettingsProvider.SetProductName("BloomUnitTest");
```
jasonleenaylor: I know Bloom was the whole point of this, and I'm not overly bothered by this, but if you have to update this file anyhow you might change the name to 'FlowerBeta' or something less obviously focused on a single product. (Keeping product names out of even the tests and comments is helpful to avoid leading junior developers into the trap of putting product specific code into a generic library)

jasonleenaylor:
Other than a few minor gripes about comments and names in the test class it looks good to me.

Palaso/Settings/CrossPlatformSettingsProvider.cs
```diff
@@ -14,17 +14,27 @@ namespace Palaso.Settings
 	/// </summary>
 	public class CrossPlatformSettingsProvider : SettingsProvider, IApplicationSettingsProvider
 	{
+		//Public only for unit testing. I can't get InternalsVisibleTo to work, possibly because of strong naming.
+		public const string UserConfigFileName = "user.config";
```
ermshiperete: Can't we make this (and the other members that are currently public for unit test purposes) protected and then derive a class (in the unit tests) that exposes the members we want to test?

regnrand: The other option is to make them internal and allow the test project to access internal stuff of the main assembly.

JohnThomson: I tried InternalsVisibleTo and cannot figure out why it doesn't work. Probably something to do with strong naming. Eberhard's idea should, though.

regnrand: Odd, since it is used in half dozen assemblies in Palaso already.

JohnThomson: Yes, including this one! Though I haven't found an instance of a test that depends on it. Both Resharper and full compilation report the methods as inaccessible.

PR #215, opened 2015-04-27
Tweak CheckMemory method to provide more information

StephenMcConnel:
Also try to clarify between process and system memory measurements.

### DIFF
--- a/PalasoUIWindowsForms/Reporting/MemoryManagement.cs
+++ b/PalasoUIWindowsForms/Reporting/MemoryManagement.cs
@@ -2,6 +2,7 @@
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 using System;
 using System.Diagnostics;
+using System.Text;
 using System.Windows.Forms;
 using L10NSharp;
 using Palaso.PlatformUtilities;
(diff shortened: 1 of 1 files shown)


PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -72,25 +74,37 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
 					totalVirtualMemory = ((availableMemory > twoGB ? twoGB : availableMemory) / 1024).ToString("N0");
 				}
 #endif
+			long virtualMemory64;
+			long peakVirtualMemory64;
+			long peakWorkingSet64;
```
papeh: I would have put the comments for these at their declaration (and declaration at assignment, since they're not used outside the using block)

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -100,7 +114,9 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
 			// In Windows/.Net, Process.PrivateMemorySize64 seems to give a reasonable value for current
 			// memory usage.  In Linux/Mono, Process.PrivateMemorySize64 gives what seems to be a virtual
 			// memory limit or some some value.  In that context, Process.WorkingSet64 appears to be the
```
papeh: some sum?

PalasoUIWindowsForms/Reporting/MemoryManagement.cs
```diff
@@ -100,7 +114,9 @@ public static bool CheckMemory(bool minor, string eventDescription, bool okToBot
...
-			// best bet for approximating how much memory the process is currently using.
+			// best bet for approximating how much memory the process is currently using.  (I'm not sure
+			// but what that would be the better value for Windows as well, but we'll leave it as it is
+			// for the time being.)
```
papeh: How about: I'm not sure *whether* that would be . . .

papeh:
:+1:

PR #216, opened 2015-04-28
Update to use ICU 54

ddaspit:
Update libpalaso to use ICU 54. Updated csproj files to copy over new ICU DLLs. Fixed some unit tests.

Please, do not merge this pull request. I (DD) will need to do the merge. I have not updated the buildupdate scripts yet, because the build dependencies have not been updated yet. Once this pull request has been approved. I will update the build dependencies, generate the buildupdate scripts, update this pull request, and perform the merge.


lib/common/icu.net.dll.config
```diff
@@ -1,9 +0,0 @@
-<?xml version="1.0"?>
-<configuration>
-  <dllmap dll="icuin40.dll" target="libicui18n.so.40"/>
-  <dllmap dll="icuuc40.dll" target="libicuuc.so.40"/>
-  <dllmap dll="icuin48.dll" target="libicui18n.so.48"/>
-  <dllmap dll="icuuc48.dll" target="libicuuc.so.48"/>
-  <dllmap dll="icuin50.dll" target="libicui18n.so.50"/>
-  <dllmap dll="icuuc50.dll" target="libicuuc.so.50"/>
-<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0,Profile=Client"/></startup></configuration>
```
JohnThomson: Looks like something to support Linux; I'm not clear why we would no longer need it, but I don't anything about how the Linux version ships.

JohnThomson:
It's a potentially breaking change that surrogates out of order throws. If the clients of our method expect it to just produce something garbed-looking, maybe we should preserve that behavior? I don't know, I haven't looked at clients. But if some bad data is out there, going from an oddity in the output to a program crash is fairly serious.

JohnThomson:
Slightly surprising that the old DLLs are removed, but no new ones appear to be added to take their places. I suppose you've figured out a new way to get them shipped as needed?

JohnThomson:
LGTM if none of my comments need action.

StephenMcConnel:
This looks okay if we're building and shipping ICU54 for libpalaso even on Linux.  If libpalaso still depends on the system icu library, then something needs to be in the .config file to map to it, and the internal code of icu.net needs to still deal with the possible versions it may encounter.

ermshiperete:
One thing to be aware of is that currently the Palaso test app seems to be broken on Trusty because buildupdate.mono.sh downloads icu.net for ICU 48 (whereas Trusty comes with 52). It would be great if this could be improved with the switch to ICU54 so that it works in all cases.

lib/common/icu.net.dll.config
ddaspit: This file is no longer being used. We now use the buildupdate script to pull down icu.net.dll.config to the lib/DebugMono and lib/ReleaseMono directories.

ddaspit:
John, that is a good point about the surrogate change. I don't know if will be an issue or not. ICU probably sees this as a bug fix. It most likely should have never allowed out of order surrogates, since it is invalid Unicode data. Regarding the deleted DLLs, they will now downloaded in the buildupdate scripts. I just can't generate the buildupdate scripts until I update the build dependencies on TeamCity, so they aren't currently included in this pull request. I will update them right before I merge in the pull request.

ddaspit:
Eberhard, I understand the issue with the Palaso test app being broken on Trusty. I don't really have a good solution for the problem. Currently, the version of ICU that ICU.NET is built against is determined at compile-time. We have a build configuration for ICU.NET to build against ICU 52 on Linux. The buildupdate script is generated using the build dependencies specified on TeamCity. The Linux buildupdate script is generated from the "palaso-precise64-master Continuous" build, which has a dependency on the "icucil-precise64-Continuous" build, which builds against ICU 48. It is setup this way most likely because it is building on Precise, which uses ICU 48. Do we have a Trusty build agent on TeamCity? If we did, we could create a new configuration called "palaso-trusty64-master Continuous" that has a dependency on a new ICU.NET configuration called "icucil-trusty64-icu52 Continuous". Unfortunately, I don't think we have a Trusty build agent yet.

ddaspit:
I have updated the buildupdate scripts.

I am going to cross my fingers and hope that the surrogate issue doesn't cause any projects any problems.

I am going to defer the Trusty issue until a later date. It is unclear at this time as to the best way to fix this issue.

PR #217, opened 2015-04-29
New try at custom registration settings location.

JohnThomson:
Also switches Registration settings to be persisted using our
cross-platform class. The first release using this will not upgrade
old registration settings. The only current client is Bloom, which
is OK with this.
Also handles a problem with saving two sets of settings
in the same file using our cross-platform provider.
Also adds a better algorithm (modify time of config file) for finding
previous configurations to upgrade from.


jasonleenaylor:
I think the commit comment could use some improvement. Since the previous commit with these changes was reverted it should note all the changes not just the differences from the last try. Also I don't think it includes the unit tests which were in the reverted commit, and it probably should.

JohnThomson:
Actually all the new unit tests from the previous attempt are present. I just didn't revert the adding of the new file, so they don't show up in this PR.
I've updated the commit message.

PR #219, opened 2015-04-30
Prevent overwriting an existing LDML file with an empty one

tombogle:
To prevent overwriting of an existing LDML file with an empty one in the event event of a failed write, make a backup first and put it back if the write fails.

### DIFF
--- a/Palaso/WritingSystems/LdmlDataMapper.cs
+++ b/Palaso/WritingSystems/LdmlDataMapper.cs
@@ -493,11 +493,33 @@ namespace Palaso.WritingSystems
 					readerSettings.ProhibitDtd = false;
 					reader = XmlReader.Create(oldFile, readerSettings);
 				}
-				using (var writer = XmlWriter.Create(filePath, CanonicalXmlSettings.CreateXmlWriterSettings()))
+				
+				string backupFilePath = null;
+				if (File.Exists(filePath))
 				{
-					writer.WriteStartDocument();
-					WriteLdml(writer, reader, ws);
-					writer.Close();
+					try
+					{
+						backupFilePath = Path.ChangeExtension(Path.GetTempFileName(), Path.GetExtension(filePath));
+						File.Copy(filePath, backupFilePath);
+					}
+					catch (Exception)
+					{
+						backupFilePath = null;
+					}
+				}
+				try
+				{
+					using (var writer = XmlWriter.Create(filePath, CanonicalXmlSettings.CreateXmlWriterSettings()))
+					{
+						writer.WriteStartDocument();
+						WriteLdml(writer, reader, ws);
+					}
+				}
+				catch (Exception)
+				{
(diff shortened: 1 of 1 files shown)


Palaso/WritingSystems/LdmlDataMapper.cs
```diff
@@ -493,11 +493,33 @@ public void Write(string filePath, WritingSystemDefinition ws, Stream oldFile, W
...
+				
+				string backupFilePath = null;
+				if (File.Exists(filePath))
 				{
-					writer.WriteStartDocument();
-					WriteLdml(writer, reader, ws);
-					writer.Close();
+					try
+					{
+						backupFilePath = Path.ChangeExtension(Path.GetTempFileName(), Path.GetExtension(filePath));
```
JohnThomson: There's a method on TempFile to get one with a particular extension. Using this would also make sure that the backup gets deleted.

tombogle: I did intend to delete the backup file at the end and not leave it around, so that was a good catch. I’m not so sure about the merits of using TempFile. For one thing, there is the (very slight) performance hit of creating the file and then overwriting it. In addition, since the backup file might not exist, we can’t just use a using to ensure that it is disposed, so a manual call to Dispose would be required to ensure that it got disposed at the correct time. So I’m leaning toward just calling File.Delete at the appropriate time.

tombogle:
Yes, but Path.GetTempFileName doesn't create an empty file (which is part of why I didn't want to use TempFile).

Palaso/WritingSystems/LdmlDataMapper.cs
```diff
@@ -493,12 +493,37 @@ public void Write(string filePath, WritingSystemDefinition ws, Stream oldFile, W
...
+					using (var writer = XmlWriter.Create(filePath, CanonicalXmlSettings.CreateXmlWriterSettings()))
+					{
+						writer.WriteStartDocument();
+						WriteLdml(writer, reader, ws);
+					}
+				}
+				catch (Exception)
+				{
+					if (backupFilePath != null)
+						File.Copy(backupFilePath, filePath);
```
jasonleenaylor: (Sorry for any confusion about my previous comment which I just accidentally deleted)
Tom, I meant my comment about overwriting on copy to refer to the copy from the backupFilePath to the filePath.
If the XmlWriter crashed while writing it could be leaving an empty or corrupt file at the filePath location and the File.Copy of the backupFilePath would fail and throw an exception in that situation unless you tell it that it can overwrite the file, right?

tombogle:
Thanks, Jason. That makes sense. Unfortunately, I'm making these changes in master, but my interest is to the benefit of our software which is building off the new Writing Systems branch. So I don't have an easy way to test to make sure my code works correctly. It's not the sort of code that is easy to write a clean unit test for.

PR #220, opened 2015-05-04
Share DBL bundle code from Protoscript Generator (PG-214)

andrew-polk:


ddaspit:
Looks fine to me. The only thing that I would be wary of is checking in binary files. I noticed that you are going to check in a zip file. Obviously, it is not a hard and fast rule and you know better than me whether it applies in this case, but I just wanted to make sure that you had thought about alternatives to checking in the zip file.

PR #221, opened 2015-05-04
Fix a memory leak in ImageToolboxControl

StephenMcConnel:
Note that setting ImageInfo.Disposed causes ImageInfo.Dispose() to return
immediately without freeing its resources.  Dispose() sets Disposed
when it is finished freeing everything.

### DIFF
--- a/PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.Designer.cs
+++ b/PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.Designer.cs
@@ -10,29 +10,6 @@ namespace Palaso.UI.WindowsForms.ImageToolbox
 		/// </summary>
 		private System.ComponentModel.IContainer components = null;
 
-		/// <summary>
-		/// Clean up any resources being used.
-		/// </summary>
-		/// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
-		protected override void Dispose(bool disposing)
-		{
-			if (disposing)
-			{
-				_toolImages.Dispose();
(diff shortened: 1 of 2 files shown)


hatton:
Since this supposed to be an automatically generated file that undergoes no hand editing, what can you tell us? Was this problem introduced by hand editing, or automatically? Is there any chance of it coming back automatically? Can things be changed so that no by-hand material is even in there (that's normally the requirement).

StephenMcConnel:
The  InitializeComponents() method and the list of member component variables are the only sections modified by Designer after the initial creation.  The Dispose() method is created initially, but it is allowed (and maybe even expected) to be edited by hand.  This problem was introduced by hand editing.

hatton:
> The Dispose() method is created initially, but it is allowed (and maybe even expected) to be edited by hand. This problem was introduced by hand editing.

Hmm. Can you point me to a writeup of that?  I for one, have always done my disposing in the non-generated part. In most cases, this is a matter of adding the object you want to get disposed to the objects "Components" property. E.g. http://stackoverflow.com/a/5069426/723299

jasonleenaylor:
LGTM

hatton:
Dispose in these things appears to be a complicated issue: http://stackoverflow.com/questions/167602/how-do-i-add-dispose-functionality-to-a-c-sharp-usercontrol

But I'll merge this.

tombogle:
As the one comment says, you can safely modify the generated Dispose method (and optionally move it to the non-Designer file). Although it is initially generated, it is never modified by Designer, which is why it’s not inside the

#region Windows Form Designer generated code

PR #223, opened 2015-05-13
BL-2169: Stop leaking image files

phillip-hopper:
This was leaking thousands of image files in Bloom.

### DIFF
--- a/PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
+++ b/PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
@@ -7,14 +7,14 @@ using System.IO;
 using Palaso.Code;
 using Palaso.IO;
 using Palaso.UI.WindowsForms.ClearShare;
-using System.Linq;
 
 namespace Palaso.UI.WindowsForms.ImageToolbox
 {
 	public class PalasoImage : IDisposable
-
 	{
 		private Metadata _metadata;
+
+		private string _tempFilePath;
 
 		public Metadata Metadata
 		{
@@ -216,7 +216,7 @@ namespace Palaso.UI.WindowsForms.ImageToolbox
 			Metadata.Write(_pathForSavingMetadataChanges);
 		}
 
-		private static Image LoadImageWithoutLocking(string path)
+		private static Image LoadImageWithoutLocking(string path, out string tempPath)
 		{
 			/*          1) Naïve approach:  locks until the image is dispose of some day, which is counter-intuitive to me
 							  return Image.FromFile(path);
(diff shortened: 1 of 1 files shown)


PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
```diff
@@ -348,6 +351,11 @@ protected virtual void Dispose(bool disposing)
 					Image.Dispose();
 					Image = null;
 				}
+
+				if (!string.IsNullOrEmpty(LeakedFilePath))
+				{
+					File.Delete(LeakedFilePath);
```
hatton: Perhaps swallow any exceptions on this when not in #Debug?

PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
```diff
@@ -238,20 +238,23 @@ private static Image LoadImageWithoutLocking(string path)
...
 			var i = new PalasoImage
 			{
-				Image = LoadImageWithoutLocking(path),
+				Image = LoadImageWithoutLocking(path, out tempPath),
 				FileName = Path.GetFileName(path),
 				_originalFilePath = path,
 				_pathForSavingMetadataChanges = path,
-				Metadata = Metadata.FromFile(path)
+				Metadata = Metadata.FromFile(path),
+				LeakedFilePath = tempPath
```
hatton: It's not going to "Leaked" anymore, right? So how about "_tempFilePath"

PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
```diff
@@ -7,15 +7,15 @@
...
-using System.Linq;
 
 namespace Palaso.UI.WindowsForms.ImageToolbox
 {
 	public class PalasoImage : IDisposable
-
 	{
 		private Metadata _metadata;
 
+		private string LeakedFilePath { get; set; }
```
hatton: Any reason to be a property instead of a member?

jasonleenaylor: I'm not a fan of the variable name. It does follow from the local variable that you are exposing through the out parameter, but could you name it TemporaryFilePath or something else?

PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
phillip-hopper: Done.

PalasoUIWindowsForms/ImageToolbox/PalasoImage.cs
phillip-hopper: Done.

hatton:
:+1:

PR #228, opened 2015-05-28
BL-2153: Allow caller to get values of physical and virtual memory

phillip-hopper:


JohnThomson:
There was no easy way to make a unit test for the old code, but it seems the new version could have one...at least that it gets some values back, even if we have no way to tell whether they are correct.

phillip-hopper:
Added unit tests.

PR #235, opened 2015-06-17
PG-239: Made LanguageIso return a default if language is not specified.

tombogle:


SIL.DblBundle/Bundle.cs
```diff
@@ -96,9 +103,17 @@ protected Bundle(string pathToZippedBundle)
 		public string Id { get { return m_dblMetadata.Id; } }
 
 		/// <summary>
-		/// 3-letter ISO 639-2 code for the language of the DBL bundle
+		/// 3-letter ISO 639-2 code for the language of the DBL bundle. If the metadata's language does
+		/// not identify an ISO 639-2 code, this returns "qaa", which is the code indicating "unknown".
```
andrew-polk: After your change, this no longer returns qaa by default.

SIL.DblBundle/Bundle.cs
```diff
@@ -81,6 +81,13 @@ protected Bundle(string pathToZippedBundle)
 
 		#region Public properties
 		/// <summary>
+		/// Typically, this should be set to SIL.WritingSystems.WellKnownSubtags.UnlistedLanguage by the client.
+		/// We would have probably just hard-coded to that, but we didn't want to have to refernce SIL.WritingSystems
```
andrew-polk: refernce -> reference

darcywong00:
Ignore the noise from Bob.  It is building using the wrong dependencies in this case (its expecting the PR to be against master in effect).

darcywong00:
Added the redundant dependency on the TC master build config so now it handles PRs on the SILWritingSystems branch ok.

PR #239, opened 2015-06-26
Fixed output folders for builds other than Debug. Also tweaked a few …

tombogle:
…things to improve interface to better serve needs of Glyssen and HearThis when working with text bundles.


papeh:
:+1:

PR #240, opened 2015-06-26
[Keyboarding] Fix keyboard switching in WS setup dialog in test app

ermshiperete:
After closing the WS setup dialog we shouldn't shutdown the
keyboard controller until we exit the app. Otherwise automatic
keyboard switching will stop working when the WS setup dialog is
opened anew.

NOTE: when running the test app on Trusty you'll have to manually
copy the icu.net.dlls for ICU52 from TC - the build currently copies
the ICU48 dlls which cause an immediate crash trying to open the
ws setup dialog.
Change-Id: Ib325f464841695c17aa7593dc080f12d9b0e3d07

### DIFF
--- a/PalasoUIWindowsForms.TestApp/TestAppForm.cs
+++ b/PalasoUIWindowsForms.TestApp/TestAppForm.cs
@@ -31,6 +31,8 @@ namespace PalasoUIWindowsForms.TestApp
 	/// ----------------------------------------------------------------------------------------
 	public partial class TestAppForm : Form
 	{
+		private bool _KeyboardControllerInitialized;
+
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Default c'tor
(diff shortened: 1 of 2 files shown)


PalasoUIWindowsForms.TestApp/TestAppForm.cs
```diff
@@ -72,25 +74,18 @@ private void OnWritingSystemSetupDialogClicked(object sender, EventArgs e)
...
 			{
-				throw;
+				KeyboardController.Initialize();
+				_KeyboardControllerInitialized = true;
 			}
-			finally
+			var wsRepo = LdmlInFolderWritingSystemRepository.Initialize(tempPath, onMigration, onLoadProblem);
+			using (var dialog = new WritingSystemSetupDialog(wsRepo))
 			{
-				KeyboardController.Shutdown();
```
jasonleenaylor: So there is already a KeyboardController.Shutdown() call in the app exit code?

ermshiperete: I moved the KeyboardController.Shutdown() to the Dispose() method.

jasonleenaylor: Sorry, that diff was suppressed so I missed it.

PR #247, opened 2015-07-07
bug/WS-334 Fix ImageTooboxControl for repeated selection of crop control

darcywong00:
https://jira.sil.org/browse/WS-334 describes the issue in more detail.

The ImageCropper is not designed to work without an image (there is no facility at the moment to clear the image).  It seems that this isn't a desirable feature for this class anyway, so we ensure that in the context of the ImageToolboxControl this doesn't happen.

There are other fixes to the ImageToolboxControl to improve the transitioning selection between two controls.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/247)


PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.cs
```diff
@@ -218,11 +214,9 @@ private void listView1_SelectedIndexChanged(object sender, EventArgs e)
 
 				if (_currentControl != null)
 				{
-					GetImageFromCurrentControl();
-
 					_panelForControls.Controls.Remove(_currentControl);
 					((IImageToolboxControl) _currentControl).ImageChanged -= new EventHandler(imageToolboxControl_ImageChanged);
-					_currentControl.Dispose();
```
papeh: Whose job is it now to dispose _currentControl?

Otherwise, LGTM :+1:

cambell-prince: We explicitly set our reference to null below.  If this is the last one out the door then the finalizer calls dispose.  It is not clear that this *is* the last one to finish, so calling Dispose explicitly is unwise IMO.  Especially considering that there's no unmanaged resources to dispose of.  I see why the Dispose was added, but in theory if there isn't a reference loop (which should be fixed with weak refs) then the managed wrappers should actually dispose of the unmanaged resources that they are managing.  There doesn't seem to be any reason to not allow the finalizer to run.  Its not that expensive an op in this case.

cambell-prince:
FYI, I pushed this code from Darcy's machine so don't really want to review it myself.  LGTM :-)  But someone from the Bloom team might like to take a look @StephenMcConnel @JohnThomson

PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.cs
JohnThomson: I don't have the big picture entirely. At least in this method, _currentControl comes from running a function which is the tag of the list item. Are these functions something the ImageCropper initializes, or something its client initializes? Either way, it should be documented either that each call of one of these functions creates a new control which the ImageCropper will dispose, or that the client that supplies the function will be responsible to dispose of it. I lean towards the former, since otherwise the client has no way to know when it can dispose the control. I'm not happy with leaving disposable objects to be disposed by their finalizers; in general, there is no way of knowing what important resources might need to be released in a timely way. Also code-checking tools complain about it. But I don't know how feasible this is or how far out into the client community such a change might affect things.

hatton:
@cambell-prince I see JT added a comment. Do you have an answering comment? Thx.

JohnThomson:
Bloom is currently experiencing a crash to do with the disposal of the image cropper...IIRC, something to do with not having permission to delete some temp files. Not sure if there is a connection, but it feels suspicious. Not sure whether this PR would help or make it worse or is unrelated.

cambell-prince:
I'm not convinced that I've got a good understanding of what'd disposed where in the ImageCropper.  Prior to this PR the currentControl called Dispose, but the object did not get cleaned up and was used later (with its resources already disposed).  Now the control owns the child controls and all are Disposed as the dialog closes.  At least that's the intention.

PalasoUIWindowsForms/ImageToolbox/ImageToolboxControl.cs
JohnThomson: Looking further, AddControl is private and is used only in OnLoad, being passed a function which creates an AcquireImageControl. There seems no reason for this complexity. If we always make one of those when an image is selected, why not just let the select code know it directly? Much more readable, and it would be unambiguous that we ought to dispose of _currentControl if it isn't null.

That leaves the question, why is it a problem to dispose the control, if indeed that was causing problems? Maybe the AcquireImageControl is disposing something it shouldn't?

JohnThomson:
I think someone needs to get a good understanding of what's going on here. We shouldn't be content to just say, if we do this non-standard thing and choose not to dispose of this object it seems to solve this problem.

hatton:
I'd sure like to see this whole thing dumped and replaced with an HTML implementation. One immediate benefit would be these crashes; a future one would be that we could cheaply add the simple image-touch up tools that have been requested, since those are readily available as canvas-based javascript components.

hatton:
Closing this as it seems dead.

PR #254, opened 2015-07-14
Re-add previous dllmaps

chrisvire:
When PalasoUIWindowsForms was changed to SIL.Windows.Forms,
the dllmap entries were removed from the app.config.  At least
the entries for libgtk and libgdk are needed for WeSay packages
for Linux to build.  Why were these removed?


chrisvire:
@ddaspit, @StephenMcConnel, @cambell-prince, @darcywong00: I created this PR so that we can choose the correct set of dllmap entries that should be included.  WeSay needs libgdk and libgtk for the packaging to complete.

PR #255, opened 2015-07-22
Fix display of meta-information in ImageToolboxControl

StephenMcConnel:
This fixes https://silbloom.myjetbrains.com/youtrack/issue/BL-2354.


papeh:
LOKTM.

PR #257, opened 2015-07-29
BL-2415: Use ProductName setting from config file, if present

phillip-hopper:
Added CoreProductName to SettingsProtectionSingleton so the
Application.ProductName can be overridden. The specific example that
prompted this change is Bloom, where the Application.ProductName for
is 'BloomAlpha' for alpha releases and 'BloomBeta' for beta releases,
but we wish to display just 'Bloom' to the user in the
SettingsProtectionDialog.

### DIFF
--- a/PalasoUIWindowsForms/SettingProtection/SettingsProtection.cs
+++ b/PalasoUIWindowsForms/SettingProtection/SettingsProtection.cs
@@ -1,4 +1,5 @@
-﻿using System.Drawing;
+﻿using System.Configuration;
+using System.Drawing;
 using System.Windows.Forms;
 
 namespace Palaso.UI.WindowsForms.SettingProtection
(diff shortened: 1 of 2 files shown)


papeh:
This does change the API, but not in a breaking way. However, if you don't foresee the new property being used for anything other than the factory password, you may wish to make it `internal`.
:+1:

JohnThomson:
Not following Papeh's comment. I don't see how it is used for factory password; rather the initial use is in prompting for the settings protection password, if I understand correctly.

JohnThomson:
I find it confusing that an application will now have two things called ProductName...the regular one (from Application.ProductName, which I think comes from one of several properties in AssemblyInfo.cs) and the new one from a key in App.Config. All the more so since we're not  changing (and I think don't want to change) what Application.ProductName returns, just what is displayed instead of it in (so far) one place.
PreferedProductName is also rather vague (and mis-spelled; should be preferred)....at this point we 'prefer' it only for these two purposes (generating the factory password and displaying the message asking for the password).
How about GenericProductName or BaseProductName or CoreProductName or DisplayProductName (depending somewhat on how else you think it might be used), and make it the same in app.config as in SettingsProtection?
Since the basic thought seems to be 'the real name of the product itself, without any adornments like Beta or Alpha', I lean towards CoreProductName...what do you think?

PalasoUIWindowsForms/SettingProtection/SettingsProtection.cs
```diff
@@ -62,8 +63,27 @@ public static SettingsProtectionSettings Settings
 
 		public static string FactoryPassword
 		{
-			get { return  (Application.ProductName.Substring(0, 1) + "7" + Application.ProductName.Substring(1, Application.ProductName.Length-1)).ToLower(); }
+			get
+			{
+				var productName = PreferredProductName;
+				return  (productName.Substring(0, 1) + "7" + productName.Substring(1, productName.Length-1)).ToLower();
```
ermshiperete: The second argument could be omitted from the second Substring call - this would make it easier to read:

    return  (productName.Substring(0, 1) + "7" + productName.Substring(1).ToLower();

phillip-hopper:
Updated.

JohnThomson:
Phil. looks good now, except unfortunately the change I requested made the commit comment out of date.

phillip-hopper:
Commit message updated.

PR #259, opened 2015-07-29
Keyboarding

ermshiperete:
Various fixes and improvements to keyboarding code (see individual commits for details)

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/259)

### DIFF
--- a/Palaso.Tests/Code/IClonableGenericTests.cs
+++ b/Palaso.Tests/Code/IClonableGenericTests.cs
@@ -22,19 +22,33 @@ namespace Palaso.Tests.Code
 
 		protected class ValuesToSet
 		{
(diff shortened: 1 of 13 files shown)


jasonleenaylor:
While my specific comment about the confusing Test class with no tests was addressed I still find it very surprising to have Test classes where all the unit tests are actually living in the base class. I won't hold back this commit for that reason, I just think that the test files could use a comment explaining why they are actually useful. It seems that all the extensions are doing is providing different sets of test data to the base class in a way that isn't commonly used in our codebase.

ermshiperete:
Reviewed 10 of 12 files at r1, 3 of 3 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/259)

PR #260, opened 2015-07-31
Expose standard properties for analytics use.

StephenMcConnel:
See https://silbloom.myjetbrains.com/youtrack/issue/BL-2418 for the
justification of this added functionality.


papeh:
:+1:

PR #261, opened 2015-08-05
Update Build and BuildUpdate scripts

papeh:


GetAndBuildThis.bat
```diff
@@ -1,28 +1,31 @@
...
 ) ELSE (
 	set BUILD_CONFIG=%1
 )
 
-if not "%VS120COMNTOOLS%" == "" (
-	echo Setting up Visual Studio Pro 2013 Tools...
-	@call "%VS120COMNTOOLS%vsvars32.bat"
-	goto build
-)
+IF "%2"=="" (
```
jasonleenaylor: A comment about the second argument would be nice

PR #263, opened 2015-08-12
Add desktop environment to error report

ermshiperete:
On Linux this change adds the currently running desktop environment (Unity, gnome shell etc) to the error report.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/263)

### DIFF
--- a/Palaso.Tests/reporting/ErrorReportTests.cs
+++ b/Palaso.Tests/reporting/ErrorReportTests.cs
@@ -1,11 +1,11 @@
-﻿
+﻿// Copyright (c) 2015 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading;
 using NUnit.Framework;
+#if !__MonoCS__
 using Palaso.Email;
+#endif
 using Palaso.Reporting;
 
 namespace Palaso.Tests.reporting
@@ -13,14 +13,17 @@ namespace Palaso.Tests.reporting
 	[TestFixture]
 	public class ErrorReportTests
 	{
-		[Test, Ignore("by hand only")]
-		public void Test()
+		[Test]
+		[Ignore("by hand only")]
+		public void ReportNonFatalException()
 		{
 			ErrorReport.EmailAddress = "pretend@8ksdfj83jls8.com";
 			ErrorReport.ReportNonFatalException(new ApplicationException("testing"));
 		}
-#if !MONO
-		[Test, Ignore("by hand only")]
+
+#if !__MonoCS__
+		[Test]
+		[Ignore("by hand only")]
 		public void TestSendEmail()
 		{
 			MAPI x = new MAPI();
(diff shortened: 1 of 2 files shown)


Palaso.Tests/reporting/ErrorReportTests.cs
```diff
@@ -28,5 +31,69 @@ public void TestSendEmail()
 			x.SendMailDirect("test", "testbody");
 		}
 #endif
+
+		[Test]
+		[Platform(Include = "Linux", Reason = "Linux specific test")]
+		public void Properties_ContainDesktopEnvironment()
```
papeh: having this test next to the other Linux-specific tests would be nice

Palaso.Tests/reporting/ErrorReportTests.cs
```diff
@@ -28,5 +31,69 @@ public void TestSendEmail()
...
+
+		[Test]
+		[Platform(Include = "Linux", Reason = "Linux specific test")]
+		public void GetStandardProperties_ContainDesktopEnvironment()
+		{
+			// SUT
+			var props = ErrorReport.GetStandardProperties();
+
+			// Verify
+			Assert.That(props.Keys, Has.Member("DesktopEnvironment"));
```
papeh: Do we need a corresponding test that this property is not returned on Windows?

Palaso/Reporting/ErrorReport.cs
```diff
@@ -416,24 +419,61 @@ public static string GetOperatingSystemLabel()
...
-			foreach (var version in list)
-			{
-				if(version.Match(System.Environment.OSVersion))
-					return version.Label + " " + Environment.OSVersion.ServicePack;
+				var list = new List<Version>();
+				list.Add(new Version(PlatformID.Win32NT,0,5, "Windows 2000"));
+				list.Add(new Version(PlatformID.Win32NT, 1, 5, "Windows XP"));
+				list.Add(new Version(PlatformID.Win32NT, 0, 6, "Vista"));
+				list.Add(new Version(PlatformID.Win32NT, 1, 6, "Windows 7"));
+				list.Add(new Version(PlatformID.Win32NT, 2, 6, "Windows 8"));
```
papeh: this will need to handle Windows 10 eventually.

ermshiperete:
Reviewed 2 of 2 files at r1.

---

<sup>**[Palaso.Tests/reporting/ErrorReportTests.cs, line 37 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/263#-JwWZZXu7mn5J6VjZSPM-r1-37)** ([raw file](https://github.com/sillsdev/libpalaso/blob/ce2e9089674008f6718ab47c68e297e3f7d0cef5/Palaso.Tests/reporting/ErrorReportTests.cs#L37)):</sup>
Done.

---

<sup>**[Palaso.Tests/reporting/ErrorReportTests.cs, line 67 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/263#-JwWZZXu7mn5J6VjZSPM-r1-67)** ([raw file](https://github.com/sillsdev/libpalaso/blob/ce2e9089674008f6718ab47c68e297e3f7d0cef5/Palaso.Tests/reporting/ErrorReportTests.cs#L67)):</sup>
Done.

---

<sup>**[Palaso/Reporting/ErrorReport.cs, line 434 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/263#-JwWZZXvAqyuhVTvtYyr-r1-434)** ([raw file](https://github.com/sillsdev/libpalaso/blob/ce2e9089674008f6718ab47c68e297e3f7d0cef5/Palaso/Reporting/ErrorReport.cs#L434)):</sup>
Good point. Also Windows 8.1 was missing. I added both (plus a comment why I think we won't see those values :smiley: )

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/263)

Palaso/Reporting/ErrorReport.cs
```diff
@@ -416,24 +419,69 @@ public static string GetOperatingSystemLabel()
...
-					return version.Label + " " + Environment.OSVersion.ServicePack;
+				// https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832%28v=vs.85%29.aspx
+				var list = new List<Version>();
+				list.Add(new Version(PlatformID.Win32NT, 5, 0, "Windows 2000"));
+				list.Add(new Version(PlatformID.Win32NT, 5, 1, "Windows XP"));
+				list.Add(new Version(PlatformID.Win32NT, 6, 0, "Vista"));
+				list.Add(new Version(PlatformID.Win32NT, 6, 1, "Windows 7"));
+				list.Add(new Version(PlatformID.Win32NT, 6, 2, "Windows 8"));
+
+				// Note: Windows might not report Windows 8.1 or 10 unless the app has
```
hatton: On Windows nowadays, we need to do something like this, instead:

var osClass = new System.Management.ManagementClass("Win32_OperatingSystem");
var oSProperties = new Dictionary<string, string>();
foreach(var queryObj in osClass.GetInstances())
{
	foreach(var prop in queryObj.Properties)
	{
		if(prop.Value!=null)
			oSProperties.Add(prop.Name,prop.Value.ToString());
    }
}

now we have oSProperties["Version"] && oSProperties["Caption"], and this will work even for new versions that you didn't know about when you made the manifest for the installed version of the app.

PR #264, opened 2015-08-14
Check bit rate before trying to play a sound

papeh:
If bit rate is 0, defer to the OS.


papeh:
This was meant to solve a problem on Linux; unfortunately, this file `AudioIrrKlangSession.cs` is apparently not compiled on Linux.  So this may not be necessary, since we have the desired behaviour on Windows already.

jasonleenaylor:
LGTM

PR #266, opened 2015-08-19
Escape quotes and backslashes passed to xdg-email

marksvc:
* Fixes LT-16454.

### DIFF
--- a/Palaso.Tests/Email/LinuxEmailProviderTests.cs
+++ b/Palaso.Tests/Email/LinuxEmailProviderTests.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using NUnit.Framework;
+using Palaso.Email;
+
+namespace Palaso.Tests.Email
+{
+	[TestFixture]
+	public class LinuxEmailProviderTests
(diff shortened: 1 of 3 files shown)


marksvc:
Thank you. I updated the comment to explain where groups of backslashes are lost. It isn't necessarily a problem because it's escaping the string for the Linux shell (though that's technically part of the problem inside xdg-email itself): half of the backslashes are lost by Mono, as Process.Start here doesn't seem to be following bash shell rules regarding backslashes inside single-quotes. A more appropriate name might be EscapeStringForMonoAndXdgEmail, but I will change it to EscapeString for the time being. 

It's not a problem of adding additional quoting, so adding quotes before calling SendMessage isn't going to help. And although the strings could be escaped before calling SendMessage, since SendMessage is part of the IEmailProvider, and other email providers don't have these same troubles, I would think it problematic to do the escaping before SendMessage() is called. But we could talk more about this if desired.

marksvc:
ermshiperete: Thank you. I will send and update using that.

PR #267, opened 2015-08-21
Add missing dispose calls to fix program hangs on Linux

jasonleenaylor:
* Dispose of temporary bitmaps as well as the Disabled grey
  scale bitmaps in BitmapButton
* Dispose of SettingsProtectionHelper components and timer

### DIFF
--- a/PalasoUIWindowsForms/SettingProtection/SettingsProtectionHelper.cs
+++ b/PalasoUIWindowsForms/SettingProtection/SettingsProtectionHelper.cs
@@ -21,6 +21,7 @@ namespace Palaso.UI.WindowsForms.SettingProtection
 	public partial class SettingsProtectionHelper : Component, IExtenderProvider
 	{
 		private readonly Dictionary<Component, bool> _controlIsUnderSettingsProtection;
+		private bool _isDisposed;
 
 		public bool CanExtend(object extendee)
 		{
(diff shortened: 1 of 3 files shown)


PalasoUIWindowsForms/SettingProtection/SettingsProtectionHelper.cs
```diff
@@ -154,6 +154,9 @@ public ISite Site
 		private bool _isDisposed = false;
 		public void Dispose()
 		{
+			base.Dispose();
```
papeh: good catch.  Should this method not be `override`?

PalasoUIWindowsForms/Widgets/BitmapButton.cs
```diff
@@ -746,31 +748,33 @@ private void paint_ImageBorder(System.Drawing.Graphics g, System.Drawing.Rectang
...
+					//
+					// draw line 1
+					//
+					border_Contract(1, ref pts);
+					g.DrawLines(pen0, pts);
+					//
+					// release resources
+					//
+					pen1.Dispose();
+					pen0.Dispose();
```
papeh: `using(pen0)`
`using(pen1)`

PalasoUIWindowsForms/SettingProtection/SettingsProtectionHelper.cs
jasonleenaylor: Actually, there should never have been a Dispose() here in the first place. Thanks for catching this, apparently I was trying to rush my change in before the weekend.

PalasoUIWindowsForms/Widgets/BitmapButton.cs
jasonleenaylor: done

jasonleenaylor:
Windows test failure seems unrelated. Looks like a fragile test.

PR #269, opened 2015-08-24
Keyboarding

ermshiperete:
- Refactor keyboard adaptors
- Fix keyboard switching on Wasta 14
- Remove Cinnamon-specific code
- Refactor keyboard setup program detection

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/269)


PalasoUIWindowsForms/Keyboarding/Linux/CombinedIbusKeyboardSwitchingAdaptor.cs
```diff
@@ -0,0 +1,239 @@
...
+		public override IKeyboardDefinition DefaultKeyboard
+		{
+			get
+			{
+				if (_defaultKeyboard == null)
+				{
+					var desired = String.Format("xkb:{0}:", DefaultLayout);
+					if (!String.IsNullOrEmpty (DefaultVariant))
+						desired = String.Format ("xkb:{0}\\{1}:", DefaultLayout, DefaultVariant);
+					var pattern = String.Format("[^A-Za-z]{0}[^A-Za-z]|^{0}[^A-Za-z]|.*[^A-Za-z]{0}$",
```
jasonleenaylor: So the backup pattern doesn't use the variant, is that intended?

jasonleenaylor:
I only had one question, overall the refactoring looks very nice. Looks good to me after addressing my question.

ermshiperete:
<sup>**[PalasoUIWindowsForms/Keyboarding/Linux/CombinedIbusKeyboardSwitchingAdaptor.cs, line 203 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/269#-JxYhixWIOtZyXpmoOOn-r3-203)** ([raw file](https://github.com/sillsdev/libpalaso/blob/0674907f581ba9303a8b4c4761f4bbb28bf40d65/PalasoUIWindowsForms/Keyboarding/Linux/CombinedIbusKeyboardSwitchingAdaptor.cs#L203)):</sup>
That's a good question, I don't know. That was existing code...

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/269)

ermshiperete:
Review status: 0 of 41 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[PalasoUIWindowsForms/Keyboarding/Linux/CombinedIbusKeyboardSwitchingAdaptor.cs, line 203 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/269#-JxYhixWIOtZyXpmoOOn-r3-203)** ([raw file](https://github.com/sillsdev/libpalaso/blob/0674907f581ba9303a8b4c4761f4bbb28bf40d65/PalasoUIWindowsForms/Keyboarding/Linux/CombinedIbusKeyboardSwitchingAdaptor.cs#L203)):</sup>
Unfortunately there are no tests for this code, so it's hard to understand what it tries to achieve. I'm not sure if we can come into situation in real life where regex would match: that would have to be a case where we use combined keyboards, but the xkb keyboard that the OS reports doesn't match one of the keyboards installed in ibus, and one of the ibus reported xkb keyboards wouldn't start with *'xkb:'*. I can't see how this would happen. And it's even harder to imagine how the variant would then come into play.

I wrote some tests for this code that I'll submit in a separate PR, and I added a comment to the code as well.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/269)

PR #272, opened 2015-08-28
Don't set ibus keyboard again when it's already set (LT-16179)

ermshiperete:
When using the IPA KMFL keyboard we get a commit from IBus after
every character. At least on Wasta 14 this causes a keyboard
activation. The previous code that didn't check if the keyboard
was already set causes the IPA keyboard to lose its context thus
preventing the IPA keyboard from working properly. This change
removes the override of IBusKeyboardAlreadySet() method and uses
the base implementation instead.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/272)


papeh:
Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/272)

PR #273, opened 2015-08-28
Add multilingual search index to ArtOfReadingImageCollection

StephenMcConnel:
This does not change the GUI at all, and does not require a change to calling ImageToolboxDialog if you're happy with English lookup or don't have access to the new Art of Reading multilingual index.  It does provide a way to pass in the language code of the language you'd like to use for image searches in Art of Reading.


papeh:
:eyes: :ok_hand: to me

PR #275, opened 2015-08-31
Don't execute IbusKeyboardSwitchingAdaptor.HandleGotFocus twice

ermshiperete:
When using the keyboarding code in FW HandleGotFocus got called
twice because of some specific requirements of FW on Linux. This
change checks the control.Focus state and returns if this doesn't
tell us that we have focus.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/275)


papeh:
Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/275)

PR #279, opened 2015-09-02
Add GUI selection of index language for Art of Reading chooser

StephenMcConnel:
This implements https://silbloom.myjetbrains.com/youtrack/issue/BL-2481.

### DIFF
--- a/PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
+++ b/PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
@@ -1,0 +1,59 @@
+﻿using System.Windows.Forms;
+
+namespace Palaso.UI.WindowsForms.Extensions
+{
+	public static class ToolStripExtensions
+	{
+		public static void SizeTextRectangleToText(this ToolStripItemTextRenderEventArgs args)
+		{
+			var textSize = args.Graphics.MeasureString(args.Text, args.TextFont);
+			const int padding = 2;
+
+			var rc = args.TextRectangle;
+			var changed = false;
+
+			// adjust the rectangle to fit the calculated text size
+			if (rc.Width < textSize.Width + padding)
+			{
+				var diffX = (int)System.Math.Ceiling(textSize.Width + padding - rc.Width);
+				rc.X -= diffX / 2;
+				rc.Width += diffX;
+				changed = true;
+			}
+
+			if (rc.Height < textSize.Height + padding)
+			{
+				var diffY = (int)System.Math.Ceiling(textSize.Height + padding - rc.Height);
+				rc.Y -= diffY / 2;
+				rc.Height += diffY;
+				changed = true;
+			}
+
+			// if nothing changed, return now
+			if (!changed) return;
+
+			args.TextRectangle = rc;
+		}
+	}
+
+	/// <summary>
+	/// Use the extension to ToolStripItemTextRenderEventArgs to implement a renderer for
+	/// ToolStrip objects that provides absolutely no border on any edge.
+	/// </summary>
+	/// <remarks>
+	/// This code originated in the Bloom project, but proved to be useful for the
+	/// ArtOfReadingChooser control in PalasoUIWindowsForms.
+	/// </remarks>
+	public class NoBorderToolStripRenderer : ToolStripProfessionalRenderer
+	{
+		protected override void OnRenderToolStripBorder(ToolStripRenderEventArgs e) { }
+
+		protected override void OnRenderItemText(ToolStripItemTextRenderEventArgs e)
+		{
+			// Without this, the ToolStrip renderer leaves some border artifacts
+			// even when the style is set to "no border".
+			e.SizeTextRectangleToText();
+			base.OnRenderItemText(e);
+		}
+	}
+}
--- a/PalasoUIWindowsForms/ImageGallery/ArtOfReadingImageCollection.cs
+++ b/PalasoUIWindowsForms/ImageGallery/ArtOfReadingImageCollection.cs
@@ -5,6 +5,7 @@ using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Windows.Forms;
 using Palaso.Extensions;
 using Palaso.IO;
 using Palaso.Linq;
@@ -18,7 +19,8 @@ namespace Palaso.UI.WindowsForms.ImageGallery
 		private Dictionary<string, string> _partialPathToWordsIndex;
 		private static readonly object _padlock = new object();
 		public string SearchLanguage { get; set; }
-
+		private List<string> _indexLanguages;
+		public IEnumerable<string> IndexLanguageIds { get { return _indexLanguages; } }
 
 		public ArtOfReadingImageCollection()
 		{
@@ -60,23 +62,17 @@ namespace Palaso.UI.WindowsForms.ImageGallery
 		/// <returns>number of index lines successfully loaded</returns>
 		public int LoadMultilingualIndex(string pathToIndexFile)
 		{
-			var defaultLang = "en";
+			const string defaultLang = "en";
 			using (var f = File.OpenText(pathToIndexFile))
 			{
-				// The file starts with a line that looks like the following:
-				//order	filename	artist	country	en	id	fr	es	ar	hi	bn	pt	th	sw	zh
-				var header = f.ReadLine();
-				var columns = header.Split(new char[]{'\t'});
-				if (columns.Length < 5 ||
-					columns[0] != "order" ||
-					columns[1] != "filename" ||
-					columns[2] != "artist" ||
-					columns[3] != "country")
-				{
+				var columns = GetColumnHeadersIfValid(f);
+				if (columns == null)
 					return 0;
-				}
 				int desiredColumn = -1;
 				int defaultColumn = -1;
+				// The first four columns are fixed metadata.  The remaining columns contain
+				// search words for different languages, one language per column.  The header
+				// contains the ISO language codes.
 				for (int i = 4; i < columns.Length; ++i)
 				{
 					if (columns[i] == SearchLanguage)
@@ -276,11 +272,6 @@ namespace Palaso.UI.WindowsForms.ImageGallery
 
 		public static string TryToGetRootImageCatalogPath()
 		{
-			//look for the cd/dvd
-/* retire this            var cdPath = TryToGetPathToCollectionOnCd();
-			if (!string.IsNullOrEmpty(cdPath))
-				return cdPath;
-*/
 			var distributedWithApp = FileLocator.GetDirectoryDistributedWithApplication(true,"Art Of Reading", "images");
 			if(!string.IsNullOrEmpty(distributedWithApp) && Directory.Exists(distributedWithApp))
 				return distributedWithApp;
@@ -321,23 +312,6 @@ namespace Palaso.UI.WindowsForms.ImageGallery
 			return null;
 		}
 
-		private static string TryToGetPathToCollectionOnCd()
-		{
-			//look for CD
-			foreach (var drive in DriveInfo.GetDrives())
-			{
-				try
-				{
-					if(drive.IsReady && drive.VolumeLabel.Contains("Art Of Reading"))
-						return Path.Combine(drive.RootDirectory.FullName, "images");
-				}
-				catch (Exception)
-				{
-				}
-			}
-			return null;
-		}
-
 		public static bool DoNotFindArtOfReading_Test = false;
 
 		public static IImageCollection FromStandardLocations()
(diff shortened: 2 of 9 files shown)


PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+			var textSize = args.Graphics.MeasureString(args.Text, args.TextFont);
+			const int padding = 2;
+
+			var rc = args.TextRectangle;
+			var changed = false;
+
+			// adjust the rectangle to fit the calculated text size
+			if (rc.Width < textSize.Width + padding)
+			{
+				var diffX = (int)System.Math.Ceiling(textSize.Width + 2 - rc.Width);
```
JohnThomson: Should the '2' in this line be padding?

PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+				rc.X -= diffX / 2;
+				rc.Width += diffX;
+				changed = true;
+			}
+
+			if (rc.Height < textSize.Height + padding)
+			{
+				var diffY = (int)System.Math.Ceiling(textSize.Height + 2 - rc.Height);
+				rc.Y -= diffY / 2;
```
JohnThomson: Here too?

PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+	/// This code originated in the Bloom project, but proved to be useful for the
+	/// ArtOfReadingChooser control in PalasoUIWindowsForms.
+	/// </remarks>
+	public class NoBorderToolStripRenderer : ToolStripProfessionalRenderer
+	{
+		protected override void OnRenderToolStripBorder(ToolStripRenderEventArgs e) { }
+
+		protected override void OnRenderItemText(ToolStripItemTextRenderEventArgs e)
+		{
+			// this is needed, especially on Linux
```
JohnThomson: It would be helpful to say why...what goes wrong on each platform without it.

PalasoUIWindowsForms/ImageGallery/ArtOfReadingImageCollection.cs
```diff
@@ -357,18 +338,48 @@ public static IImageCollection FromStandardLocations(string lang)
...
+			{
+				using (var f = File.OpenText(pathToIndexFile))
+				{
+					// The file starts with a line that looks like the following:
+					//order	filename	artist	country	en	id	fr	es	ar	hi	bn	pt	th	sw	zh
+					var header = f.ReadLine();
+					if (String.IsNullOrEmpty(header))
+						return;
+					var columns = header.Split (new char[]{ '\t' });
+					if (columns.Length < 5 ||
```
JohnThomson: Comment "Require the four fixed columns and at least one language"

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -77,33 +86,32 @@ private void _searchButton_Click(object sender, EventArgs e)
...
 					bool foundExactMatches;
-					IEnumerable<object> results = _imageCollection.GetMatchingPictures(_searchTermsBox.Text, out foundExactMatches);
-					if (results.Count() == 0)
-					{
-						_messageLabel.Visible = true;
-						_searchResultStats.Text = "Found no matching images".Localize("ImageToolbox.NoMatchingImages");
-					}
-					else
+					// (avoid enumerating the returned IEnumerable<object> more than once by copying to a List.)
+					var results = new List<object>(_imageCollection.GetMatchingPictures(_searchTermsBox.Text, out foundExactMatches));
```
JohnThomson: Consider using Linq ToList(). Is <object> the best possible signature?

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -77,33 +86,32 @@ private void _searchButton_Click(object sender, EventArgs e)
...
+					if (results.Any())
 					{
 						_messageLabel.Visible = false;
+						betterLinkLabel1.Visible = false;
```
JohnThomson: This variable needs a meaningful name. There is no clue here why we would want to hide it.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -77,33 +86,32 @@ private void _searchButton_Click(object sender, EventArgs e)
...
+						var fmt = "Found {0} images".Localize("ImageToolbox.MatchingImages", "The {0} will be replaced by the number of matching images");
 						if (!foundExactMatches)
-							_searchResultStats.Text += string.Format(" with names close to {0}.", _searchTermsBox.Text);
+							fmt = "Found {0} images with names close to {1}".Localize("ImageToolbox.AlmostMatchingImages", "The {0} will be replaced by the number of images found.  The {1} will be replaced with the search string.");
+						_searchResultStats.Text = string.Format(fmt, results.Count, _searchTermsBox.Text);
+					}
+					else
+					{
+						_messageLabel.Visible = true;
+						if (PlatformUtilities.Platform.IsWindows && !_searchLanguageMenu.Visible)
```
JohnThomson: Without knowing what betterLinkLabel1 contains, this is hard to evaluate, but it seems odd that it gets hidden on all platforms but only shown on Windows. Is it an optional component that doesn't WORK on Linux? A comment might be appropriate.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -190,24 +198,25 @@ private void ArtOfReadingChooser_Load(object sender, EventArgs e)
 				//label1.Visible = _searchTermsBox.Visible = _searchButton.Visible = _thumbnailViewer.Visible = false;
 				_messageLabel.Visible = true;
 				_messageLabel.Font = new Font(SystemFonts.DialogFont.FontFamily, 10);
-				_messageLabel.Text = @"This computer doesn't appear to have the 'International Illustrations: the Art Of Reading' gallery installed yet.";
+				_messageLabel.BackColor = Color.White;
```
JohnThomson: Why are we having to mess with BackColor? This would normally be in designer...I don't see that it is ever set to anything but white.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -218,10 +227,98 @@ private void ArtOfReadingChooser_Load(object sender, EventArgs e)
 		   // _searchTermsBox.Text = @"flower";
 			_searchButton_Click(this,null);
 #endif
-
+			var searchControl = _searchTermsBox.TextBox.TheControl;
+			if (searchControl is TextBox)
+				searchControl.MinimumSize = _searchTermsBox.Size;
```
JohnThomson: It's not obvious why this isn't done in designer or why it might not be a text box. A comment along the lines, "On Linux the search control contains an HTML control rather than just a text box." would be a beginning.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -218,10 +227,98 @@ private void ArtOfReadingChooser_Load(object sender, EventArgs e)
...
 			_messageLabel.BackColor = Color.White;
 		}
 
+		private void SetMessageLabelText()
+		{
+			var msg = "In the box above, type what you are searching for, then press ENTER.".Localize("ImageToolbox.EnterSearchTerms");
+			var searchLang = "English + Indonesian";
+			if (_searchLanguageMenu.Visible)
+				searchLang = _searchLanguageMenu.ToolTipText;
```
JohnThomson: Why is the search language stored in the menu's tooltip?? and when and why would the menu not be visible??

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -218,10 +227,98 @@ private void ArtOfReadingChooser_Load(object sender, EventArgs e)
...
+				betterLinkLabel1.Visible = true;
+				betterLinkLabel1.BackColor = Color.White;
+				betterLinkLabel1.Location = new Point(betterLinkLabel1.Left, _messageLabel.Bottom + 4);
+			}
+			_messageLabel.Text = msg;
+		}
+
+		protected class LanguageChoice
+		{
+			static readonly List<string> _languageNamesRecognizableByOtherLatinScriptReaders = new List<string> { "en", "fr", "es", "it", "tpi", "pt", "id" };
```
JohnThomson: These feel like language IDs, not names.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -218,10 +227,98 @@ private void ArtOfReadingChooser_Load(object sender, EventArgs e)
...
+			if (item != null)
+			{
+				var lang = item.Tag as LanguageChoice;
+				if (lang != null && SearchLanguage != lang.Id)
+				{
+					_searchLanguageMenu.Text = lang.NativeName;
+					_searchLanguageMenu.ToolTipText = lang.ToString();
+					SearchLanguage = lang.Id;
+					_imageCollection.ReloadImageIndex(lang.Id);
+					SetMessageLabelText();		// Update with new language name.
```
JohnThomson: To make this work it seems something would have to update the menu's tooltip?? Would it be clearer to just pass the LanguageChoice object to SetMessageLabelText?

PalasoUIWindowsForms/Widgets/TextInputBox.cs
```diff
@@ -22,6 +22,8 @@ public class TextInputBox : UserControl
 		public TextInputBox()
 		{
 			_inputBox = CreateTextBox();
+			if (_inputBox.TheControl is TextBox)
+				_inputBox.TheControl.Dock = DockStyle.Fill;	// Make the real inner box match the size of the outer virtual box.
 			this.Controls.Add(_inputBox.TheControl);
```
JohnThomson: Could we handle the minimumSize thing here, too...or maybe in an override of OnSizeChanged?

PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
StephenMcConnel: looks like it should be.

PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
StephenMcConnel: yes.

PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
StephenMcConnel: I don't know why Linux was thought to be a special problem.  It solves a minor rendering glitch on Windows as well.  (Somebody else wrote this code and left this cryptic message.)  I'll at least make the comment less cryptic.

PalasoUIWindowsForms/Extensions/ToolStripExtensions.cs
JohnThomson: You're proposing this change...you're supposed to know!

PalasoUIWindowsForms/ImageGallery/ArtOfReadingImageCollection.cs
StephenMcConnel: ok

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: It's going through an interface that already has one IEnumerable<> member.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: This was a pre-existing variable.  But I've renamed it to _downloadArtOfReadingLink

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: You're right, it's not Windows specific.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: ok.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: ok

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: The menu's Text contains only the native name of the language.  The tooltip contains the native name + the English name in parenthesis if it is in a nonRoman script or otherwise unlikely to be guessable by a literate European.

StephenMcConnel: commented.  The menu is visible only when the multilingual index is available.  We're deliberately allowing people not to upload a 300MB installer to get the new index.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: renamed to idsOfRecognizableLanguages

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
StephenMcConnel: The tooltip is being updated if you look.  We don't have the LanguageChoice object in both places where SetMessageLabelText() is called.  And if we're using the old index, we never have one of those.  I think the current code is working okay.

PalasoUIWindowsForms/Widgets/TextInputBox.cs
StephenMcConnel: moved the minimum size setting to an OnSizeChanged override.  We don't know the size in the constructor.

PalasoUIWindowsForms/ImageToolbox/ArtOfReadingChooser.cs
```diff
@@ -218,8 +228,98 @@ private void ArtOfReadingChooser_Load(object sender, EventArgs e)
...
+				_downloadInstallerLink.BackColor = Color.White;
+				_downloadInstallerLink.Location = new Point(_downloadInstallerLink.Left, _messageLabel.Bottom + 4);
+			}
+			_messageLabel.Text = msg;
+		}
+
+		protected class LanguageChoice
+		{
+			static readonly List<string> idsOfRecognizableLanguages = new List<string> { "en", "fr", "es", "it", "tpi", "pt", "id" };
+			private readonly CultureInfo _info;
```
JohnThomson: I find it slightly surprising that this list is distinctly different (e.g., no Arabic) from the list we advertise AoR now supports.

PR #281, opened 2015-09-03
Use Safe String Formatting when formatting unknown strings

papeh:


papeh:
Tried writing a unit test, but in the test context, the line that crashes is never actually run.

PalasoUIWindowsForms/Progress/LogBox.cs
```diff
@@ -287,7 +287,7 @@ private void Write(Color color, FontStyle style, string msg, params object[] arg
 						rtfBoxForDelegate.SelectionColor = color;
 						rtfBoxForDelegate.SelectionFont = fnt;
 #endif
-						rtfBoxForDelegate.AppendText(string.Format(msg + Environment.NewLine, args));
+						rtfBoxForDelegate.AppendText(GenericProgress.SafeFormat(msg + Environment.NewLine, args));
```
hatton: Is there an advantage to using this wrapper instead of the string extension FormatWithErrorStringInsteadOfException? I would go so far as to say that this method isn't enough about progress that GenericProgress should be offering it as a public method.

papeh: I agree.  When I pick this up again, I could gladly remove the GenericProgress method entirely (with appropriate fanfare on the email list)

PR #282, opened 2015-09-04
Fix problem with keyboard setup link in Wasta 14

ermshiperete:
On Wasta 14 without the parens we returned "keyboard" as arguments
which broke the keyboard setup link.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/282)


StephenMcConnel:
I'm not sure why the parentheses are needed here, but LGTM.  :+1:

PR #283, opened 2015-09-07
Allow creation of a file lock from a file path

ddaspit:
Add new method to create a SimpleFileLock from a file path. This allows applications to specify where the lock file will be created.


ermshiperete:
LGTM :+1:

ddaspit:
Thanks for the quick review.

PR #287, opened 2015-09-12
Fix a momentary black box artifact when querying  Art Of Reading

hatton:
Hiding the viewer during search prevents a momentary black screen on first query (and only the first).

The related Bloom issue: https://silbloom.myjetbrains.com/youtrack/issue/BL-2729


StephenMcConnel:
Looks good to me.  :+1:   I figured it would probably be a simple fix.  :-)

PR #288, opened 2015-09-14
Replace SharpZipLib with Ionic.DotNetZip in DblBundle extraction code

andrew-polk:


papeh:
:+1:

ermshiperete:
Does this work on Linux? I think in other places we had to go the opposite direction in order for it to work on Linux. Or what's the reason behind this change?

andrew-polk:
It does work on Linux.  My initial implementation used DotNetZip because libpalaso was already using DotNetZip in the archiving code.  However, when I pushed my original code, it *did* fail on linux.  So... I exchanged it for SharpZipLib.  My intention was to go back and change the previous archiving code to use SharpZipLib so there would only be one zip library in use in libpalaso.  But when I realized Paratext used DotNetZip (including the linux version), I decided to take another shot at getting DotNetZip to work for linux.  I essentially pared down the zip code from Paratext to the few lines I needed, and it worked on linux.  So now I am changing this code from SharpZipLib to DotNetZip so we will once again have only one zip library in use in libpalaso.

tombogle:
No need to check for directory's existence before creating it. Directory.CreateDirectory already handles this.
The code to unzip a zip file with its directory structure shuld be moved to a place in the IO namespace in the core DLL, so it doesn't have to be duplicated in each place where it is needed.

ermshiperete:
If it works on Linux than I'm OK with this change :+1: 

The only reservation I have against using Ionic.Zip from ../lib is that it's unclear where this version came from, which makes it harder to maintain in case we're running into problems that might be caused by the zip library.

PR #289, opened 2015-09-16
API: Remove public string-formatting methods from Progress

papeh:
Use Safe String Formatting when formatting unknown strings

### DIFF
--- a/Palaso/Progress/Progress.cs
+++ b/Palaso/Progress/Progress.cs
@@ -421,13 +421,8 @@ namespace Palaso.Progress
 		public SynchronizationContext SyncContext { get; set; }
 		public void WriteStatus(string message, params object[] args)
 		{
-#if MONO
-			Console.Write("                          ".Substring(0, indent*2));
-			Console.WriteLine(GenericProgress.SafeFormat(message, args));
-#else
 			Console.Write("                          ".Substring(0, indent * 2));
-			Console.WriteLine(GenericProgress.SafeFormat(message, args));
-#endif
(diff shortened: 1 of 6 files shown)


Palaso/Progress/Progress.cs
```diff
@@ -608,15 +603,7 @@ public bool WasCancelled
...
 		{
-			LastError = LastWarning =LastStatus = string.Empty;
-		}
-	}
-
-	public static class SafeFormatString
-	{
-		public static string EscapeThingsThatLookLikeArguments(this string format, params object[] args)
-		{
-			return format.Replace("{", @"\{");
```
papeh: not sure this would have worked anyway; MSDN doc says, for String.Format, escape `{` and `}` with `{{` and `}}`.

PalasoUIWindowsForms/Progress/LogBox.cs
```diff
@@ -287,7 +287,7 @@ private void Write(Color color, FontStyle style, string msg, params object[] arg
 						rtfBoxForDelegate.SelectionColor = color;
 						rtfBoxForDelegate.SelectionFont = fnt;
 #endif
-						rtfBoxForDelegate.AppendText(string.Format(msg + Environment.NewLine, args));
+						rtfBoxForDelegate.AppendText(msg.FormatWithErrorStringInsteadOfException(args) + Environment.NewLine);
```
papeh: No unit tests because `SafeInvoke` prevents this line from being executed before the test exists.

PR #290, opened 2015-09-28
Fix backspacing with Korean IBus keyboard

ermshiperete:
When IBus didn't handle the key we should let the control deal with
the keypress. This fixes the problem when typing with Korean Hangul
keyboard it was not possible to backspace past the beginning of the
pre-edit. This change makes it possible, although there is still a
slight problem in that when backspacing over the first character
in the pre-edit the display doesn't get updated, but when backspacing
again the first character in the pre-edit and the character before
that get deleted (as it should be, but the visual indication is
wrong).

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/290)

### DIFF
--- a/PalasoUIWindowsForms/Keyboarding/Linux/IbusKeyboardSwitchingAdaptor.cs
+++ b/PalasoUIWindowsForms/Keyboarding/Linux/IbusKeyboardSwitchingAdaptor.cs
@@ -178,7 +178,6 @@ namespace Palaso.UI.WindowsForms.Keyboarding.Linux
 				e.Control.MouseDown += HandleMouseDown;
 				e.Control.PreviewKeyDown += HandlePreviewKeyDown;
 				e.Control.KeyPress += HandleKeyPress;
-				e.Control.KeyDown += HandleKeyDown;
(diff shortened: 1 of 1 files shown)


papeh:
This should probably be handled with events--perhaps subscribing `OnKeyDown` only if IBus has handled Preview KeyDown, and immediately unsubscribing `OnKeyDown` on its completion.


---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/290)

ermshiperete:
Excellent suggestion. Done.


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/290)

jasonleenaylor:
Reviewed 1 of 1 files at r3.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/290)

PR #292, opened 2015-09-30
API: Use NuGet package instead of copied MarkdownSharp.cs

jasonleenaylor:
* The MarkdownSharp copied from google code was an out of date
  version with some buggy behavior regarding lists
* Delete our copied code file and use the NuGet package for
  MarkdownDeep which is a fork of the original source project

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/292)


jasonleenaylor:
I marked it as API since I'm removing a public class. I know of no uses outside of the PalasoUIWindowsForms assembly.

papeh:
:+1:

marksvc:
LGTM

PR #293, opened 2015-10-01
Betterlabel

marksvc:
I would be glad to hear if anyone has a better name for CanSelectText

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/293)

### DIFF
--- a/PalasoUIWindowsForms/Widgets/BetterLabel.cs
+++ b/PalasoUIWindowsForms/Widgets/BetterLabel.cs
@@ -13,43 +13,66 @@ namespace Palaso.UI.WindowsForms.Widgets
 	{
 		private Brush _textBrush;
 		private Brush _backgroundBrush;
+		private bool _isTextSelectable;
 
 		public BetterLabel()
 		{
 			InitializeComponent();
 			ReadOnly = true;
-			Enabled = false;
+			IsTextSelectable = false;
 			ForeColor = SystemColors.ControlText;
 			SetStyle(ControlStyles.UserPaint,true);
 			_backgroundBrush = new SolidBrush(BackColor);
 			_textBrush = new SolidBrush(ForeColor);
-#if __MonoCS__
-			// These settings (and the overrides below) don't keep the BetterLabel from getting focus in Mono,
-			// but they do keep it from showing highlighted selections within the label.
-			Enabled = true;
-			// These may not work any better than setting UserPaint true, but ...
-			SetStyle(ControlStyles.UserMouse, true);
-			SetStyle(ControlStyles.Selectable | ControlStyles.StandardClick | ControlStyles.StandardDoubleClick, false);
-#endif
+		}
+
+		/// <summary>
+		/// Should the label allow a user to select and copy the text, such as from an error message?
+		/// </summary>
+		public bool IsTextSelectable
+		{
+			get
+			{
+				return _isTextSelectable;
+			}
+			set
+			{
+				_isTextSelectable = value;
+
+				// Always Enabled on Mono so text is black not grey.
+				if (Environment.OSVersion.Platform == PlatformID.Unix)
+				{
+					Enabled = true;
+					return;
+				}
+
+				Enabled = value;
+			}
 		}
 
 #if __MonoCS__
 		protected override void OnMouseDown(MouseEventArgs args)
 		{
-			// ignore the mouse totally.
+			if (IsTextSelectable)
+				base.OnMouseDown(args);
+			// Or ignore the mouse totally.
 		}
 		protected override void OnMouseMove(MouseEventArgs args)
 		{
-			// ignore the mouse totally.
+			if (IsTextSelectable)
+				base.OnMouseMove(args);
+			// Or ignore the mouse totally.
 		}
 		protected override void OnMouseUp(MouseEventArgs args)
 		{
-			// ignore the mouse totally.
+			if (IsTextSelectable)
+				base.OnMouseUp(args);
+			// Or ignore the mouse totally.
 		}
 #endif
 
 		/// <summary>
-		/// we custom draw so that we can be ReadOnly without being necessarily grey
+		/// Custom draw to be ReadOnly without being necessarily grey.
 		/// </summary>
 		/// <param name="e"></param>
 		/// <remarks>
@@ -57,6 +80,7 @@ namespace Palaso.UI.WindowsForms.Widgets
 		/// (A Mono comment claims that MS/.Net doesn't call OnPaint, which it does,
 		/// and it's unclear how to fix the Mono code reliably.)  Mono also seems to
 		/// ignore ControlStyles settings almost totally.
+		/// The text is black in Mono if Enabled is true.
 		/// </remarks>
 		protected override void OnPaint(PaintEventArgs e)
 		{
--- a/PalasoUIWindowsForms/Widgets/BetterLabel.cs
+++ b/PalasoUIWindowsForms/Widgets/BetterLabel.cs
@@ -5,7 +5,7 @@ using System.Windows.Forms;
 namespace Palaso.UI.WindowsForms.Widgets
 {
 	/// <summary>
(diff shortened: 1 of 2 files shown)


papeh:
looks good; see comments.


---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 8 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-ZGdCdkIjIKrRjfnEQ)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L8)):</sup>
.NET (all caps)


---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 44 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-ZHLpMeq7R_vc6DnIw)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L44)):</sup>
How about the following, to avoid an unnecessary set?  I can also think of a few cases when a client would want to specifically disable a label.

    if(value || !IsUnix)
        Enabled=value;


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/293)

ermshiperete:
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 78 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-aSnNDxXQ8_tyMmCIq)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L78)):</sup>
Isn't `Enabled` always `true` in Mono?


---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 130 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-aT9FJEmKx9aWCQYSO)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L130)):</sup>
I find this comment a little hard to understand. How about something like the following: 

    //This may require the use of mono-sil which fixes MeasureString to calculate the correct value to be tall enough


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/293)

marksvc:
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 8 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-ZGdCdkIjIKrRjfnEQ)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L8)):</sup>
Done.


---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 44 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-ZHLpMeq7R_vc6DnIw)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L44)):</sup>
That's clever, and something like this may work too:

    Enabled = value || Environment.OSVersion.Platform == PlatformID.Unix;

But I think we can avoid the unnecessary set while being more readable by doing it this way:

    if (Environment.OSVersion.Platform == PlatformID.Unix)
    {
        Enabled = true;
        return;
    }

    Enabled = value;


> I can also think of a few cases when a client would want to specifically disable a label.

And they still could if wanted, unless they set CanSelectText after setting betterLabel.Enabled = false;

If someone wants a label that is not Enabled but where the user can select text, or if the above limitation is a pain, a solution can probably be devised (perhaps such as by overriding Enabled and doing special handling). I can work on that now if you think that's important to do at this time.


---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 78 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-aSnNDxXQ8_tyMmCIq)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L78)):</sup>
Apparently not. Or rather, it appears that it can be set to false.

In the BetterLabel constructor I just put:

			Console.WriteLine("Enabled? {0}",Enabled);
			Enabled=false;
			Console.WriteLine("Enabled? {0}",Enabled);
			Enabled=true;
			Console.WriteLine("Enabled? {0}",Enabled);

And the output is:

Enabled? True
Enabled? False
Enabled? True


---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 130 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-aT9FJEmKx9aWCQYSO)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L130)):</sup>
Thank you.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/293)

marksvc:
I am changing CanSelectText to IsTextSelectable. 

I am not able to test this patch on Windows currently, but it's progress.


---

Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/293)

marksvc:
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 78 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-aSnNDxXQ8_tyMmCIq)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L78)):</sup>
Oh, maybe you meant in the context of the way this BetterLabel class is being written to work. Yes; kind of. Enabled should always be true in Mono in this context, unless a client sets Enabled to false.  The comment about the text being black in Mono if Enabled is true was to add to the discussion regarding how to make the text the right colour on various platforms.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/293)

papeh:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[PalasoUIWindowsForms/Widgets/BetterLabel.cs, line 44 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/293#-K-ZHLpMeq7R_vc6DnIw)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5271afc07e655da60d2184b7ac8043c377afe639/PalasoUIWindowsForms/Widgets/BetterLabel.cs#L44)):</sup>
I like the clever one-liner, but agree that your choice may be more scrutable.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/293)

PR #295, opened 2015-10-09
Retrieve Windows version information from a different api

jasonleenaylor:
* Uses the version information from a network api when Windows
  reports that it is "Windows 8"
* This will allow applications to report useful version information
  on versions of windows which are released after the application is built.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/295)

### DIFF
--- a/Palaso/Reporting/ErrorReport.cs
+++ b/Palaso/Reporting/ErrorReport.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Collections.Specialized;
 using System.ComponentModel;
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
I've considered using the new method for all versions, but proving that the dll reports correctly on all old versions was not something I wanted to verify.

ermshiperete:
:+1:  LGTM


---

Reviewed 2 of 2 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/295)

PR #296, opened 2015-10-13
Add build task for generating release artifacts

jasonleenaylor:
The task uses a markdown file and a version number as input
* Optionally stamp the markdown file with latest version and release
* Generate html file (or element) from the markdown file
* Generate new changelog entry from the latest markdown entry

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/296)

### DIFF
--- a/BuildTaskTests/GenerateReleaseArtifactsTests.cs
+++ b/BuildTaskTests/GenerateReleaseArtifactsTests.cs
@@ -1,0 +1,298 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.Build.Framework;
+using NUnit.Framework;
+using Palaso.BuildTasks;
+using Palaso.TestUtilities;
+
+namespace Palaso.BuildTask.Tests
+{
+	[TestFixture]
+	internal class GenerateReleaseArtifactsTests
+	{
+		#region TestHelper classes
+
+		private sealed class MockEngine : IBuildEngine
+		{
+			public List<string> LoggedMessages = new List<string>();
+
+			public void LogErrorEvent(BuildErrorEventArgs e)
+			{
+				LoggedMessages.Add(e.Message);
+			}
+
+			public void LogWarningEvent(BuildWarningEventArgs e)
+			{
+				LoggedMessages.Add(e.Message);
+			}
+
+			public void LogMessageEvent(BuildMessageEventArgs e)
+			{
+				LoggedMessages.Add(e.Message);
+			}
+
+			public void LogCustomEvent(CustomBuildEventArgs e)
+			{
+				LoggedMessages.Add(e.Message);
+			}
+
+			public bool BuildProjectFile(string projectFileName, string[] targetNames, IDictionary globalProperties,
+				IDictionary targetOutputs)
+			{
+				throw new NotImplementedException();
+			}
+
+			public bool ContinueOnError { get; private set; }
+			public int LineNumberOfTaskNode { get; private set; }
+			public int ColumnNumberOfTaskNode { get; private set; }
+			public string ProjectFileOfTaskNode { get; private set; }
+		}
+
+		/// <summary>
+		/// This class is implemented to avoid a dependency on Palaso (which isn't strictly circular, but sure feels like it)
+		/// The TempFile class that lives in Palaso.IO is a more robust and generally preferred implementation.
+		/// </summary>
+		private sealed class TwoTempFilesForTest : IDisposable
+		{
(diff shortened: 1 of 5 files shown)


jasonleenaylor:
FLExBridge would be a user of all the capabilities added here and we may start using them in FLEx as well, but it seems that the changelog entry creation could prove useful to several other projects.

ermshiperete:
Great work! A few questions and suggestions.


---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

<sup>**[BuildTaskTests/GenerateReleaseArtifactsTests.cs, line 31 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/296#-K0ZJF2uRbVIAF8K4Fm9)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5112b79bff0ca05b16893b4772e488fb6b3363ea/BuildTaskTests/GenerateReleaseArtifactsTests.cs#L31)):</sup>
According to [MSDN](https://msdn.microsoft.com/en-us/library/system.io.file.delete%28v=vs.110%29.aspx) it wouldn't be necessary to check for the file existence, but it doesn't hurt:

> If the file to be deleted does not exist, no exception is thrown.


---

<sup>**[BuildTaskTests/GenerateReleaseArtifactsTests.cs, line 160 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/296#-K0ZPOgmAGGtyc5-FLvB)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5112b79bff0ca05b16893b4772e488fb6b3363ea/BuildTaskTests/GenerateReleaseArtifactsTests.cs#L160)):</sup>
I would prefer if we could check the full content of the first line. That would document how the version number and date gets set in the MD file.


---

<sup>**[BuildTaskTests/GenerateReleaseArtifactsTests.cs, line 178 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/296#-K0ZKl8gTuyWpSxMyDol)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5112b79bff0ca05b16893b4772e488fb6b3363ea/BuildTaskTests/GenerateReleaseArtifactsTests.cs#L178)):</sup>
Would it be better to check that the line is still "## DEV_VERSION_NUMBER: DEV_RELEASE_DATE"?


---

<sup>**[BuildTaskTests/GenerateReleaseArtifactsTests.cs, line 204 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/296#-K0ZLBMxL1UFZdve5s3B)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5112b79bff0ca05b16893b4772e488fb6b3363ea/BuildTaskTests/GenerateReleaseArtifactsTests.cs#L204)):</sup>
You might also want to check the new lines/first line - that would make it clearer what the task actually puts in.


---

<sup>**[BuildTaskTests/GenerateReleaseArtifactsTests.cs, line 245 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/296#-K0ZO6h3SLbV32ZR7EEm)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5112b79bff0ca05b16893b4772e488fb6b3363ea/BuildTaskTests/GenerateReleaseArtifactsTests.cs#L245)):</sup>
I think it would be better to use `Is.StringStarting` here as well - it needs to be a 1st level list. When we use `Is.StringContaining` it could be any level.

It might also be good to test the beginning of the text for each line, e.g. `Is.StringStarting("  * Update French")`. This would make it clearer how things get converted.


---

<sup>**[Palaso.MSBuildTasks/GenerateReleaseArtifacts.cs, line 120 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/296#-K0ZNp1E7Qkd2f2pFTWs)** ([raw file](https://github.com/sillsdev/libpalaso/blob/5112b79bff0ca05b16893b4772e488fb6b3363ea/Palaso.MSBuildTasks/GenerateReleaseArtifacts.cs#L120)):</sup>
It might be good to enhance the unit test above with a 3rd level list, just for documentation purposes.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/296)

papeh:
Reviewed 4 of 6 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/296)

PR #303, opened 2015-10-28
Add Detach() to TempFile

hatton:
[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/303)


StephenMcConnel:
Looks good to me, as they say.  :+1:

papeh:
LGTM, although Detaching makes the file not temporary.  Is there any reason File wouldn't work?

PR #304, opened 2015-10-30
Msg on denied file writes now includes read only & file permissions

hatton:
[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/304)


papeh:
:+1:

PR #307, opened 2015-11-03
Fix ThumbnailViewer (BL-2901)

gmartin7:
* reflection was referencing the old namespace system

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/307)

### DIFF
--- a/SIL.Windows.Forms/ImageGallery/ThumbnailViewer.cs
+++ b/SIL.Windows.Forms/ImageGallery/ThumbnailViewer.cs
@@ -41,13 +41,13 @@ namespace SIL.Windows.Forms.ImageGallery
 			if (UseWebViewer)
 			{
 				var path = Path.Combine(Path.GetDirectoryName(new Uri(Assembly.GetExecutingAssembly().CodeBase).LocalPath),
-					"PalasoUIWindowsForms.GeckoBrowserAdapter.dll");
+					"SIL.Windows.Forms.GeckoBrowserAdapter.dll");
 				if (File.Exists(path))
 				{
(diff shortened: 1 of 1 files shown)


SIL.Windows.Forms/ImageGallery/ThumbnailViewer.cs
```diff
@@ -41,13 +41,13 @@ public IThumbnailViewer CreateViewer()
...
 				var path = Path.Combine(Path.GetDirectoryName(new Uri(Assembly.GetExecutingAssembly().CodeBase).LocalPath),
-					"PalasoUIWindowsForms.GeckoBrowserAdapter.dll");
+					"SIL.Windows.Forms.GeckoBrowserAdapter.dll");
 				if (File.Exists(path))
 				{
 					var assembly = Assembly.LoadFile(path);
 					if (assembly != null)
 					{
-						var browser = assembly.GetType("Palaso.UI.WindowsForms.ImageGallery.WebThumbnailViewer");
+						var browser = assembly.GetType("SIL.Windows.Forms.GeckoBrowserAdapter.WebThumbnailViewer");
```
JohnThomson: Good catch...but why do we want to change from ImageGallery to GeckoBrowserAdapter??

gmartin7: I dunno. I assumed Damien and company just put WebThumbnailViewer over there for their own reasons. (WebThumbnailViewer IS a GeckoFx control, as far as I understand things). I didn't move it, I just figured out where it had gotten to. :)

PR #310, opened 2015-11-05
Reset writing system properties when reading an LDML file

ddaspit:
If an existing writing system definition is reused when loading an LDML file, we want to make sure that any properties that are not set during reading are reset to their defaults, otherwise we can run into issues.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/310)


cambell-prince:
LGTM.  We should let NRSI know that we are still waiting on having live work as well as staging does.


---

Reviewed 3 of 3 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/310)

ddaspit:
Thanks for the review, Cambell. I think you are the only one in my timezone that can review pull requests. I was waiting for Martin to get back to talk to him about the SLDR, but I guess I will go ahead and email Sharon C. She can probably help.

PR #318, opened 2015-11-13
Get implicit scripts from SLDR alltags.txt

ddaspit:
* parse alltags.txt to get implicit scripts, SLDR-to-Palaso tag mapping, and available LDML files
* only download alltags.txt if it is newer than the embedded alltags.txt
* update alltags.txt to latest version
* update IetfLanguageTag to use implicit scripts from alltags.txt
* add OfflineSldrAttribute to make it easier for unit tests to use SLDR in offline mode
* remove LanguageSubtag.ImplicitScriptCode
* SIL-compatible normalization is no longer necessary
* synchronize access to SLDR cache directory

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/318)


JohnThomson:
Reviewed 37 of 39 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

_[SIL.TestUtilities/OfflineSldrAttribute.cs, line 8 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LQUDNg3yRvOVphVQh:-K3LQUDNg3yRvOVphVQi:1799771169) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.TestUtilities/OfflineSldrAttribute.cs#L8)):_
Needs a class comment (don't assume reader knows what Sldr stands for or how a temp folder can stand in for a missing online resource)

---

_[SIL.TestUtilities/OfflineSldrAttribute.cs, line 29 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LR3Z28Ium1bD8v3Ul:-K3LR3Z28Ium1bD8v3Um:-352374139) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.TestUtilities/OfflineSldrAttribute.cs#L29)):_
I suppose this is defined by ITestAction but its significance isn't obvious so it might be worth a comment.

---

_[SIL.WritingSystems.Tests/IetfLanguageTagTests.cs, line 281 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LS4uLhCgjcnhFbvRx:-K3LS4uLhCgjcnhFbvRy:65015438) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L281)):_
This test sort of covers the case, but its name doesn't indicate it because it tests something else as well. I expected to see a test like Canonicalize_NonImplicitScript_DoesNotSuppressesScript. Are there other variations we should consider, like not suppressing it when the (non-implicit) script is the only tag after Language? Are there special cases worth testing, like where the only tag for a language is one with an implicit script, or where there is an implicit script but some other tag does exist? I don't know the exact state of the scripts you have tried, so these may be covered.

---

_[SIL.WritingSystems.Tests/LanguageLookupTests.cs, line 6 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LTIRwC07ObOw4ZeQu:-K3LTIRwC07ObOw4ZeQv:1803381446) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems.Tests/LanguageLookupTests.cs#L6)):_
Should this have the new OfflineSldr attribute?

---

_[SIL.WritingSystems.Tests/LdmlInFolderWritingSystemRepositoryTests.cs, line 607 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LToYVbhapu4M-bvUO:-K3LToYVbhapu4M-bvUP:769373055) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems.Tests/LdmlInFolderWritingSystemRepositoryTests.cs#L607)):_
I suppose you actually expect this to pass even on Linux? If not, might need an attribute to stop it running there.

---

_[SIL.WritingSystems.Tests/SldrTests.cs, line 41 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LUt4hlFtGnWdH3KfN:-K3LUt4hlFtGnWdH3KfO:-2077354425) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems.Tests/SldrTests.cs#L41)):_
I find this surprising, since the new attribute class sets it false on Dispose. I would expect both to either set it to the same default state, or both set it back to what it was before the test started. If the difference is intentional, I think it should be explained in a comment.

---

_[SIL.WritingSystems.Tests/StandardSubtagsTests.cs, line 36 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LWJpkzOmvRFxtra5g:-K3LWJpkzOmvRFxtra5h:1733752148) ([raw file](https://github.com/sillsdev/libpalaso/blob/13a48a58667515b8c37c43000dbaded029e5112e/SIL.WritingSystems.Tests/StandardSubtagsTests.cs#L36)):_
It's not useful to explicitly test that it does NOT have one?

---

_[SIL.WritingSystems/Sldr.cs, line 44 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3L__uQSfUDbj_ZJEhJ:-K3L__uQSfUDbj_ZJEhK:-1411415442) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L44)):_
I know it's not new in this change, but this is a significant class that seems to be used all over with a very cryptic name. It badly needs a good class comment.

---

_[SIL.WritingSystems/Sldr.cs, line 90 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3L_xF3IpVmPf_UGGcG:-K3L_xF3IpVmPf_UGGcH:1347360021) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L90)):_
Feels like accessing or modifying this should be synchronized if you are protecting the cache based on it.

---

_[SIL.WritingSystems/Sldr.cs, line 274 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LdKhToYczucV4KtUs:-K3LdKhToYczucV4KtUt:-2134261784) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L274)):_
Inside this, we may lock the SldrCacheMutex. Is there any way another thread could lock that first and then need the SyncRoot and form a deadlock?
Should we comment (or better, enforce somehow) that one of these locks may only be claimed by a thread that already has the other?

---

_[SIL.WritingSystems/Sldr.cs, line 306 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3Lfl69ofoGxZaBlj78:-K3Lfl69ofoGxZaBlj79:1665457152) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L306)):_
Shouldn't we wait to delete it until we successfully download a new one? Or reconstruct it somehow if we don't succeed? I guess not...this branch says that the cached version is older than the one built into the program. How can that happen? Someone installed a newer version? I think a comment is called for.

---

_[SIL.WritingSystems/Sldr.cs, line 313 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LfDfFBcAaJLBv9XkR:-K3LfDfFBcAaJLBv9XkS:1741634672) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L313)):_
This block deserves a comment (or better, its own method). I _think_ it's just testing whether there has been a commit to some online version of something since the built-in version?

---

_[SIL.WritingSystems/Sldr.cs, line 372 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LgjF-W1JBOpIH-5qP:-K3LgjF-W1JBOpIH-5qQ:-1338285264) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L372)):_
This somewhat convoluted bit of code to parse a line would be a nice candidate for a method, which would also make testing easier. It would also be easier to follow with a comment that gave several examples of what a line might hold and what each part represents and what results are wanted for each example.

---

_[SIL.WritingSystems/Sldr.cs, line 419 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3Lhh2nN_kY4mUccKT8:-K3Lhh2nN_kY4mUccKT9:1600786352) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/Sldr.cs#L419)):_
Please comment to explain why we don't replace the existing tag if it has an implicit script code

---

_[SIL.WritingSystems/SldrLanguageTagInfo.cs, line 3 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3Li5xcVorwsQLxvDDc:-K3Li5xcVorwsQLxvDDd:326959158) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/SldrLanguageTagInfo.cs#L3)):_
I think even a class this simple deserves a comment, especially when public.

---

_[SIL.WritingSystems/SldrLanguageTagInfo.cs, line 18 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LiQGXeMTWHihe8ome:-K3LiQGXeMTWHihe8omf:-1536325884) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/SldrLanguageTagInfo.cs#L18)):_
Either here or below, explain how and in what kinds of cases LanguageTag is different from SldrLanguageTag

---

_[SIL.WritingSystems/SldrLanguageTagInfo.cs, line 23 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LicbCjDQ3vDrhUS6h:-K3LicbCjDQ3vDrhUS6i:-1401147094) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/SldrLanguageTagInfo.cs#L23)):_
Will this be null or empty if the instance has none? Is there any distinction between a tag that has no script code (explicit or otherwise) and one that has only an implicit one? Maybe worth saying clearly (if true) that it will be null/empty for something like en-Cyrl that has an explicit script code.

---

_[SIL.WritingSystems/SldrLanguageTagInfo.cs, line 33 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LjGmPjWbUpLG6PKcp:-K3LjGmPjWbUpLG6PKcq:-1618781436) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/SldrLanguageTagInfo.cs#L33)):_
Deserves a comment. I have no idea in what case or sense one of these objects might be unavailable. A dead language? Invalid code? We have no data about it? We're offline and couldn't download the info?

---

_[SIL.WritingSystems/WritingSystemDefinition.cs, line 431 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3LjlgLS21l3zHRs-7U:-K3LjlgLS21l3zHRs-7V:400740636) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/WritingSystemDefinition.cs#L431)):_
I think it would be helpful, in the presence of all this stuff about implicit scripts, that any property returning a Script should have a comment indicating clearly what it will return for a WritingSystem whose Script is implicit (i.e., the implicit script or some empty value or...?)

---

_[SIL.WritingSystems/WritingSystemDefinition.cs, line 791 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/318#-K3Ll5jT848TlMF98Fm_:-K3Ll5jUFL9BN-rarueA:-2042085251) ([raw file](https://github.com/sillsdev/libpalaso/blob/b1d4d374089eb676731a27cbd8c16b14e655a00e/SIL.WritingSystems/WritingSystemDefinition.cs#L791)):_
Is this Set method something inherited? I can't find it here. So it's hard to figure out what this is doing.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/318)_

PR #322, opened 2015-11-19
Update SimpleFileLock to generate lock files that can be read by older versions of Palaso

ddaspit:
* generate lock file contents using JSON.NET instead of object serialization
* add unit tests for SimpleFileLock

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/322)

### DIFF
--- a/SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs
+++ b/SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs
@@ -1,0 +1,164 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using NUnit.Framework;
+using SIL.IO.FileLock;
+using SIL.IO.FileLock.FileSys;
+
+namespace SIL.Tests.IO.FileLock
+{
+	[TestFixture]
+	public class SimpleFileLockTests
+	{
+		private static readonly string LockPath = Path.Combine(Path.GetTempPath(), "SimpleFileLockTests");
+
+		[SetUp]
+		public void SetUp()
+		{
+			CleanupLockFile();
+		}
+
+		[TearDown]
+		public void TearDown()
+		{
+			CleanupLockFile();
+		}
+
+		private static void CleanupLockFile()
+		{
+			if (File.Exists(LockPath))
+				File.Delete(LockPath);
+		}
+
+		[Test]
+		public void TryAcquireLock_Unlocked_ReturnsTrue()
+		{
+			SimpleFileLock fileLock = SimpleFileLock.Create("SimpleFileLockTests");
+			Assert.That(fileLock.TryAcquireLock(), Is.True);
+			fileLock.ReleaseLock();
+		}
+
+		[Test]
+		public void TryAcquireLock_LockedByActiveProcess_ReturnsFalse()
+		{
+			LockIO.WriteLock(LockPath, new FileLockContent
+			{
+				PID = 0,
+				ProcessName = Process.GetProcessById(0).ProcessName,
+				Timestamp = DateTime.Now.Ticks
+			});
+			SimpleFileLock fileLock = SimpleFileLock.Create("SimpleFileLockTests");
+			Assert.That(fileLock.TryAcquireLock(), Is.False);
+		}
+
+		[Test]
+		public void TryAcquireLock_LockedByDeadProcess_ReturnsTrue()
+		{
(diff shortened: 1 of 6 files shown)


ermshiperete:
Great to see tests for this! One test that's missing is that we can acquire two locks with different names - don't know how important that is, but it would test the `lockName` parameter of `SimpleLockFile.Create`.

---

Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

_[SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs, line 42 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/322#-K3TjQ4dgaHXsTTggIFE:-K3TjQ4dgaHXsTTggIFF:1424590335) ([raw file](https://github.com/sillsdev/libpalaso/blob/008ac8074659fabf26d59d329c34c01862f38661/SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs#L42)):_
*Active* is ambiguous: I first read it to be the currently running process instead of any running process. But I can't think of anything better...

---

_[SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs, line 60 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/322#-K3TimbxFFm64RnKx0gx:-K3TimbxFFm64RnKx0gy:-1239931031) ([raw file](https://github.com/sillsdev/libpalaso/blob/008ac8074659fabf26d59d329c34c01862f38661/SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs#L60)):_
There's the theoretical slight chance that this test might fail because there just happens to be a process named "Stuff" that happens to have PID 9999. Maybe use `Path.GetRandomFileName()` for the process name instead?

---

_[SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs, line 78 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/322#-K3TiJNlKnty_Y6tvfyB:-K3TiJNlKnty_Y6tvfyC:-588983856) ([raw file](https://github.com/sillsdev/libpalaso/blob/008ac8074659fabf26d59d329c34c01862f38661/SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs#L78)):_
I'm not clear what we test here. That we don't delete the lock file if it's still opened? Then a slightly clearer name for the test might be `TryAcquireLock_FileStillOpened_ReturnsFalse`.

---

_[SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs, line 154 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/322#-K3Tfh-cP47-sJPsr1rX:-K3Tfh-dAIUihCwXHefR:1051201513) ([raw file](https://github.com/sillsdev/libpalaso/blob/008ac8074659fabf26d59d329c34c01862f38661/SIL.Core.Tests/IO/FileLock/SimpleFileLockTests.cs#L154)):_
I'm not clear what we test here. That we don't delete the lock file if it's still opened? Then a slightly clearer name for the test might be `ReleaseLock_FileStillOpened_LockFileExists`.

---

_[SIL.Core/IO/FileLock/FileSys/LockIO.cs, line 54 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/322#-K3TdiRAuXQUR-LbzQFh:-K3TdiRAuXQUR-LbzQFi:1238312958) ([raw file](https://github.com/sillsdev/libpalaso/blob/008ac8074659fabf26d59d329c34c01862f38661/SIL.Core/IO/FileLock/FileSys/LockIO.cs#L54)):_
forward or backward compatibility?

It might help for the understanding to add a comment somewhere that explains the problem why we need this (if we have two versions of an app that should work on the same lock, and one version of the app uses an older version of the SIL.Core/Palaso library).

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/322#-:-K3Tk6Z0Rz6gbPicjbF9:-540323630)_

ermshiperete:
Reviewed 6 of 6 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/322)_

ermshiperete:
Reviewed 1 of 1 files at r2.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/322)_

PR #324, opened 2015-11-25
[v2.6] Fix crash from getting active keyboard

ermshiperete:
Under some circumstances ActiveKeyboard can't detect the currently
active keyboard and used to return null which caused a crash. Now
we return KeyboardDescriptionZero instead.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/324)


hatton:
The Zero name feels awkward, but that's not introduced in this PR, so... LGTM

PR #325, opened 2015-11-25
Fix crash from getting active keyboard

ermshiperete:
Under some circumstances ActiveKeyboard can't detect the currently
active keyboard and used to return null which caused a crash. Now
we return KeyboardController.NullKeyboard instead.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/325)

### DIFF
--- a/SIL.Windows.Forms.Keyboarding/KeyboardController.cs
+++ b/SIL.Windows.Forms.Keyboarding/KeyboardController.cs
@@ -492,7 +492,7 @@ namespace SIL.Windows.Forms.Keyboarding
 				if (_activeKeyboard == null)
 				{
 					_activeKeyboard = Adaptors[KeyboardAdaptorType.System].SwitchingAdaptor.ActiveKeyboard;
-					if (_activeKeyboard == null)
+					if (_activeKeyboard == null || _activeKeyboard == NullKeyboard)
 					{
 						try
 						{
--- a/SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs
+++ b/SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs
@@ -858,13 +858,15 @@ namespace SIL.Windows.Forms.Keyboarding.Windows
 				{
 					var profile = ProfileMgr.GetActiveProfile(Guids.TfcatTipKeyboard);
 					return Keyboard.Controller.AvailableKeyboards.OfType<WinKeyboardDescription>()
-						.FirstOrDefault(winKeybd => InputProcessorProfilesEqual(profile, winKeybd.InputProcessorProfile));
+						.FirstOrDefault(winKeybd => InputProcessorProfilesEqual(profile, winKeybd.InputProcessorProfile)) ??
+						KeyboardController.NullKeyboard;
 				}
 
 				// Probably Windows XP where we don't have ProfileMgr
 				var lang = ProcessorProfiles.GetCurrentLanguage();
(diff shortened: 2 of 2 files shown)


hatton:
LGTM. But has a failing test on TC/Linux.

ddaspit:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs, line 862 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/325#-K4W_jwpW5F67Nj4n5IE:-K4W_jwqQlbUZ-PsAmZh:1990638867) ([raw file](https://github.com/sillsdev/libpalaso/blob/46f7d8c2c1b2c09f10f6581f2b9fc663f606fd0d/SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs#L862)):*
Wouldn't this bypass the logic in KeyboardController. ActiveKeyboard that tries to get the keyboard for the current input language if the adaptor returns null for ActiveKeyboard? Is that logic unnecessary for Windows keyboards?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/325)*

ermshiperete:
I don't understand why the tests keep failing. My change only changed a Windows-only file that doesn't get compiled on Linux, so it can't have anything to do with the test failures!?

Further investigation that the tests also fail on the master branch. Reason is some flaky behaviour in dbus-sharp. I started updating dbus-sharp to a newer version, but that introduces new issues that need to be fixed first.

---

---

*[SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs, line 862 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/325#-K4W_jwpW5F67Nj4n5IE:-K4XGEZNQXP8m6Y9cE8H:2004150295) ([raw file](https://github.com/sillsdev/libpalaso/blob/46f7d8c2c1b2c09f10f6581f2b9fc663f606fd0d/SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs#L862)):*
Good point. I changed the code in `KeyboardController.ActiveKeyboard`.

The reason that it's better to return `NullKeyboard` instead of `null` is that we not always go through `KeyboardController.ActiveKeyboard`, e.g. the `WinKeyboardAdapter` calls `ActiveKeyboard` several times. Using `NullKeyboard` omits the need to check for `null` everywhere.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/325#-:-K41iW6tL9bK6DjGOuGU:1400357038)*

ddaspit:
The change in KeyboardController deals with my question, so your changes look good to me. 

It looks like their is something wrong with the Linux build agent.  I don't think it is related to these changes. The Chorus build on Linux has also been having problems as well. A lot of the builds fail with an out of memory exception.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/325#-:-K4_c0xBu3nIpi6YABVJ:-305719794)*

ddaspit:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/325)*

ddaspit:
Cambell and I fixed the Linux build agent. It was configured to have too little memory. It seems to be running better now.

PR #327, opened 2015-12-01
More plumbing to let client of ImageToolBox read SearchLanguage

hatton:
[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/327)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs
+++ b/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs
@@ -22,6 +22,8 @@ namespace SIL.Windows.Forms.ImageToolbox
 		private Control _currentControl;
 		private PalasoImage _imageInfo;
 		private ListViewItem _cropToolListItem;
+		private string _incomingSearchLanguage;
+		private AcquireImageControl _acquireImageControl;
 
 		public ImageToolboxControl()
 		{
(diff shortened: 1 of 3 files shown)


JohnThomson:
Reviewed 3 of 3 files at r1.

---

*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 128 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/327#-K4Y6r3LBz3RCBvF9_j-:-K4Y6r3LBz3RCBvF9_j0:2079614976) ([raw file](https://github.com/sillsdev/libpalaso/blob/046235c4abce5e7a600ebd5a4ed2dbf3ee3f3281/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L128)):*
Since you say "at some point" we add the acquireImageControl, should we set its searchLanguage if it already exists?
I imagine it's a tricky decision: if the user has done something that causes it to be created and then chosen a language, we wouldn't want to change it; but if they do something and then cancel, we don't want to lose the value we set (which currently will be forgotten, if the control is created before we set this).
Of course, normal usage is probably to set it before the event that causes the control to be created, so maybe it doesn't matter. If that's the expectation, you could encourage it using Assert(_acquireImageControl == null).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/327)*

hatton:

---

*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 128 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/327#-K4Y6r3LBz3RCBvF9_j-:-K4YR29oUL0TwBbt5XLG:-902661439) ([raw file](https://github.com/sillsdev/libpalaso/blob/046235c4abce5e7a600ebd5a4ed2dbf3ee3f3281/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L128)):*
I don't see any risk. The current system is just that if the client were to change the SearchLanguage after showing the control, it wouldn't have any effect.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/327)*

JohnThomson:
LGTM, but possibly the linux integration failure should be investigated? Probably spurious, I don't see how this could have broken anything...

ermshiperete:
That's a general problem that is also on the master branch. The tests time out because of a problem in dbus-sharp. A newer version of that should fix this, but unfortunately the newer version introduces other issues. TomH will fix dbus-sharp ASAP, so that we can update ibusdotnet and then finally get the tests in libpalaso working again.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/327#-:-K4YSttDcxQGFF188SZX:925511377)*

PR #330, opened 2015-12-07
Move classes from FW's SilUtils assembly into libpalaso

ermshiperete:
[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/330)


papeh:
Did a quick check for "FLEx residue" (I'm sure you've changed nothing but the namespace). I agree that it's a good idea for these to live in libpalaso.

---

Reviewed 43 of 46 files at r1, 5 of 5 files at r2.

---

*[SIL.Core/Text/StringUtils.cs, line 112 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/330#-K4xs7guElKGH7ZXBWcY:-K4xs7guElKGH7ZXBWcZ:-1683732534) ([raw file](https://github.com/sillsdev/libpalaso/blob/b3c00e07dbd181194b7b40c7495b9a5f7aca7678/SIL.Core/Text/StringUtils.cs#L112)):*
Now may be a good time to see if it's still necessary to "renamb" this method, but only if it's minimal effort.

---

*[SIL.Core/Unicode/Surrogates.cs, line 8 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/330#-K4xqtUga6A5NNBbwSCj:-K4xqtUga6A5NNBbwSCk:-1747000165) ([raw file](https://github.com/sillsdev/libpalaso/blob/b3c00e07dbd181194b7b40c7495b9a5f7aca7678/SIL.Core/Unicode/Surrogates.cs#L8)):*
no need to mention `FwUtils` here

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/330#-:-K4xuSP_Br93UXdgc8GK:-886262440)*

jasonleenaylor:
Two questions:
What use outside of FLEx have you discovered for these which encouraged you to move them?
Did you spend a little time making sure that none of this functionality exists elsewhere in Palaso?

hatton:
PathwayUtils does not appear to belong in libpalaso.

hatton:
This submission makes me uncomfortable. Some bits, like the code that deals with surrogates, feel like a natural addition to SIL's language/script-focussed needs. Some, like a whole class just to see if one rect is inside another, just feel like they are being dumped in. Others, like PathWayUtils, clearly don't belong. Some, like the FileDialog classes, *might* belong, but they don't come with any coments about why you might want to use them, have no sample code, etc. Almost all deviate from libpalaso's coding standards and instead refect our old c++-inspired hungarian notation, and few following good parameter naming patterns (e.g. ````StringUtils.GetUiString(ResourceManager rm, string kstid)```` instead of ```GetLocalizedString(ResourceManager resourceManager, string localizationId)```. 

I'm not criticizing the original code here. In the context where they came from, it's all normal (hungarian, poor naming, figure out the context from where it is used, etc.).  But for use shared library, little of it would be acceptable if submitted in a PR by itself. It would have to be improved first.

cambell-prince:
I suspect this PR might be motivated by work on the use of Fieldworks code in LanguageForge to get Send/Receive working.  I'm all in favor of separating UI code from non-UI code and gaining reuse into other products.  However, I'm concerned that if its moved into libpalaso as is it would never be tidied up.  Perhaps a better halfway house would be creating a library in the context of fieldworks that could be used by LanguageForge until such time as it gets improved and stagger the submissions to libpalaso at that time.

ermshiperete:
Ditching this PR

PR #331, opened 2015-12-08
Enhance Logger and ExceptionHelper classes

ermshiperete:
Add a few methods that are needed for FW so that the duplicated
code in FW can be removed.

This change adds the following methods: Logger.WriteError,
ExceptionHelper.GetHiearchicalExceptionInfo, .GetExceptionText,
and .LogAndIgnoreErrors. ErrorReport will use these new methods.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/331)

### DIFF
--- a/SIL.Core/Reporting/ErrorReport.cs
+++ b/SIL.Core/Reporting/ErrorReport.cs
@@ -138,6 +138,18 @@ namespace SIL.Reporting
 			ErrorReport.AddStandardProperties();
 		}
 
+		private static void UpdateEmailSubject(Exception error)
+		{
+			var subject = new StringBuilder();
+			subject.AppendFormat("Exception: {0}", error.Message);
+			try
+			{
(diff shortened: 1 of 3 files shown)


SIL.Core/Reporting/ErrorReport.cs
```diff
@@ -138,62 +138,29 @@ public static void Init(string emailAddress)
...
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		///
+		/// </summary>
+		/// <param name="error"></param>
+		/// <returns></returns>
+		/// ------------------------------------------------------------------------------------
+		public static string GetExceptionText(Exception error)
+		{
+			UpdateEmailSubject(error);
```
hatton: This is an unexpected side-effect of a "Get" method. But it was there before this PR... it's just less hidden now :-)

hatton:
:+1:

PR #334, opened 2015-12-15
Allow to set prefix for logging

ermshiperete:
This change adds a new overload for Init() that takes a prefix for
the log file. This is needed for FieldWorks where a new process gets
started when a new project gets opened. Without this change the log
file of the previously started instance would be overwritten when
opening a new project.

Also a little bit of cleanup (moving a method and static variable
around that accidentally got added in the IDisposable region).

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/334)

### DIFF
--- a/SIL.Core/Reporting/Logger.cs
+++ b/SIL.Core/Reporting/Logger.cs
@@ -72,6 +72,9 @@ namespace SIL.Reporting
 	public class Logger: IDisposable, ILogger
 	{
 		private static Logger _singleton;
+		private static string _actualLogPath;
+		private static string _logfilePrefix;
+
 		protected StreamWriter m_out;
 		private StringBuilder m_minorEvents;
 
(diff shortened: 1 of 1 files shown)


marksvc:
Other than my comment, LGTM. Assuming the singleton is shutdown+started only in the new FW process and the old FW process keeps the old singleton.

---

Reviewed 1 of 1 files at r2.

---

*[SIL.Core/Reporting/Logger.cs, line 97 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/334#-K5aUzYmqhYb9y--Yr-o:-K5aUzYmqhYb9y--Yr-p:446181110) ([raw file](https://github.com/sillsdev/libpalaso/blob/97070fbe7af68406ab00c9445e8500f094f07497/SIL.Core/Reporting/Logger.cs#L97)):*
This sentence could use some grammatical attention.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/334#-:-K5aVGGmf9--MBI8I8_a:-557274516)*

marksvc:
LGTM.

---

Reviewed 1 of 1 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/334#-:-K5aYssowDoIuJfg_lLz:-492975211)*

ermshiperete:
The singleton is actually also shutdown and started when the first process of FW starts: for the general initialization and attempt to locate the project we use a different log file. Once we load the database we start a new logger. When the user opens a second project we start a second process which creates its own logger. This change just mirrors existing FW code.

---

---

*[SIL.Core/Reporting/Logger.cs, line 97 \[r2\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/334#-K5aUzYmqhYb9y--Yr-o:-K5aYHJn7BWDlfm-hPhp:-499146009) ([raw file](https://github.com/sillsdev/libpalaso/blob/97070fbe7af68406ab00c9445e8500f094f07497/SIL.Core/Reporting/Logger.cs#L97)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/334#-:-K5aYRkgcg-iLPizFNGy:1500726046)*

PR #336, opened 2015-12-29
Improve performance of Detect.Reentry

rmunn:
We don't need a List here when we're never storing more than a single object. Use a HashSet instead for O(1) lookup.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/336)

### DIFF
--- a/SIL.Core/Code/Detect.cs
+++ b/SIL.Core/Code/Detect.cs
@@ -13,7 +13,7 @@ namespace SIL.Code
 
 	public class ReentryDetecter : IDisposable
 	{
-		private static List<string> _currentlyInScope = new List<string>();
+		private static HashSet<string> _currentlyInScope = new HashSet<string>();
 		private readonly string _name;
 		private bool _didReenter;
 


papeh:
Would not O(n) where n=1 be the same as O(1)? :lgtm: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/336#-:-K7CNeMvvu_Hh55_uOeE:2083954248)*

cambell-prince:
Fine with me.  What motivated the need for this change?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/336#-:-K7ERA9RvcbaH7WKX7JI:387124729)*

rmunn:
There was no specific reason for the change, like profiling code and tracing a slowdown to this class. Rather, I was reading through libpalaso code to get familiar with it, and noticed that this code was less efficient than it could be with a very simple change.

Also, I notice that my commit message isn't quite accurate. We're storing only a single object *per instance*, but it's a static class member, so there might (and likely will, in real usage) be multiple objects stored. If so, then an O(1) lookup is indeed a win over O(N).

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/336#-:-K8ItpS30Ax8R8f7xBhX:308740130)*

ermshiperete:
:lgtm: 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/336#-:-K98IwsqCXFz8czoridt:980077423)*

PR #339, opened 2016-01-07
PG-433: Add function to set permission to 'full control' for everyone

phillip-hopper:
If a directory is created by an administrator in ProgramData, the default
access is read-only for non-admin users

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/339)


cambell-prince:
It would be nice to see this new method accompanied with a test.

PR #340, opened 2016-01-11
BL-2966 Make image library handle image orientation

JohnThomson:
Image metadata can contain an Orientation field which
indicates how it needs to be flipped and/or rotated to
look right. The image library was ignoring this, which
caused images to look wrong in the toolbox, and was not
copying the metadata to new image files, which caused
them to look wrong elsewhere.
This patch fixes both these problems, though fixing
the first made it unnecessary to fix the second at least
to solve THIS issue.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/340)

### DIFF
--- a/SIL.Windows.Forms/ClearShare/MetaData.cs
+++ b/SIL.Windows.Forms/ClearShare/MetaData.cs
@@ -372,7 +372,9 @@ namespace SIL.Windows.Forms.ClearShare
 		/// </summary>
 		public void NormalizeOrientation()
 		{
-			var ifdTag = _originalTablibMetadata?.GetTag(TagTypes.TiffIFD) as IFDTag;
+			if (_originalTablibMetadata == null)
+				return;
+			var ifdTag = _originalTablibMetadata.GetTag(TagTypes.TiffIFD) as IFDTag;
 			if (ifdTag != null)
 				ifdTag.Orientation = ImageOrientation.TopLeft;
 		}
--- a/SIL.Windows.Forms/ClearShare/MetaData.cs
+++ b/SIL.Windows.Forms/ClearShare/MetaData.cs
@@ -7,6 +7,7 @@ using System.Text.RegularExpressions;
 using L10NSharp;
 using SIL.Extensions;
 using TagLib;
+using TagLib.IFD;
 using TagLib.Image;
 using TagLib.Xmp;
 using File = System.IO.File;
@@ -361,7 +362,7 @@ namespace SIL.Windows.Forms.ClearShare
 			return file != null && !file.GetType().FullName.Contains("NoMetadata");
 		}
 
-		public void Write(string path)
+		public void Write(string path, string originalImagePath = null)
 		{
 			// do not attempt to add metadata to a file type that does not support it.
 			if (!FileFormatSupportsMetadata(path))
@@ -374,6 +375,19 @@ namespace SIL.Windows.Forms.ClearShare
 			// If it is, we want this tag to exist, since otherwise tools like exiftool (and hence old versions
 			// of this library and its clients) won't see our copyright notice and creator, at least.
 			file.GetTag(TagTypes.Png, true);
+			// If we know where the image came from, copy as much metadata as we can to the new image.
+			if (originalImagePath != null && File.Exists(originalImagePath))
+			{
+				var originalFile = TagLib.File.Create(originalImagePath) as TagLib.Image.File;
+				if (originalFile != null)
+					file.CopyFrom(originalFile);
+				// But, don't copy any orientation information from the original image. At least for now,
+				// all callers which pass an originalImagePath already corrected for any orientation
+				// recorded in the original image.
+				var ifdTag = file.GetTag(TagTypes.TiffIFD) as IFDTag;
+				if (ifdTag != null)
+					ifdTag.Orientation = ImageOrientation.TopLeft;
+			}
 			SaveInImageTag(file.ImageTag);
 			file.Save();
 			//as of right now, we are clean with respect to what is on disk, no need to save.
--- a/SIL.Windows.Forms/ClearShare/MetaData.cs
+++ b/SIL.Windows.Forms/ClearShare/MetaData.cs
@@ -60,16 +60,21 @@ namespace SIL.Windows.Forms.ClearShare
 		}
 
 		/// <summary>
+		/// If the MetaData was loaded from a file, this stores all the other metadata from that file,
+		/// which will typically be useful to write to a file derived from it.
+		/// </summary>
+		TagLib.Image.File _originalTablibMetadata;
+
+		/// <summary>
 		/// NB: this is used in 2 places; one is loading from the image we are linked to, the other from a sample image we are copying metadata from
 		/// </summary>
 		/// <param name="path"></param>
 		/// <param name="destinationMetadata"></param>
 		private static void LoadProperties(string path, Metadata destinationMetadata)
 		{
-			TagLib.Image.File file;
 			try
 			{
-				file = TagLib.File.Create(path) as TagLib.Image.File;
+				destinationMetadata._originalTablibMetadata = TagLib.File.Create(path) as TagLib.Image.File;
 			}
 			catch (TagLib.UnsupportedFormatException)
 			{
@@ -79,7 +84,7 @@ namespace SIL.Windows.Forms.ClearShare
 				// even in DEBUG mode, because else a lot of simple image tests fail
 				return;
 			}
-			LoadProperties(file.ImageTag, destinationMetadata);
+			LoadProperties(destinationMetadata._originalTablibMetadata.ImageTag, destinationMetadata);
 		}
 
 		/// <summary>
@@ -362,7 +367,26 @@ namespace SIL.Windows.Forms.ClearShare
 			return file != null && !file.GetType().FullName.Contains("NoMetadata");
 		}
 
-		public void Write(string path, string originalImagePath = null)
+		/// <summary>
+		/// Set standard orientation on any metadata saved from the original image (typically because we hard-rotated the image)
+		/// </summary>
+		public void NormalizeOrientation()
+		{
+			var ifdTag = _originalTablibMetadata?.GetTag(TagTypes.TiffIFD) as IFDTag;
+			if (ifdTag != null)
+				ifdTag.Orientation = ImageOrientation.TopLeft;
+		}
+
+		/// <summary>
+		/// Write the metadata to the specified image file. Optionally, if this MetaData was made from a file,
+		/// any other original metadata from that file can be copied to the new one. This can include all kinds
+		/// of details like where and when a photo was taken, image size, and so forth, so it should not be
+		/// done except for copies of the image; for example, not when applying the same ClearShare settings
+		/// to a group of images.
+		/// </summary>
+		/// <param name="path"></param>
+		/// <param name="copyAllMetaDataFromOriginal"></param>
+		public void Write(string path, bool copyAllMetaDataFromOriginal = false)
 		{
 			// do not attempt to add metadata to a file type that does not support it.
 			if (!FileFormatSupportsMetadata(path))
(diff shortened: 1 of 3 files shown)


SIL.Windows.Forms/ClearShare/MetaData.cs
```diff
@@ -361,7 +362,7 @@ public bool FileFormatSupportsMetadata(string path)
 			return file != null && !file.GetType().FullName.Contains("NoMetadata");
 		}
 
-		public void Write(string path)
+		public void Write(string path, string originalImagePath = null)
```
hatton: This seems... awkward. Seems like the metadata should instead pull in all it needs when first exposed to the original file. If for some reason that isn't possible, then it could at least keep the path it was given in FromFile around to use later if it's still available. That kind of state is undesirable, but also par for the course in the .net image api's.

SIL.Windows.Forms/ImageToolbox/PalasoImage.cs
```diff
@@ -288,10 +288,53 @@ public static PalasoImage FromFile(string path)
 				Metadata = Metadata.FromFile(path),
 				_tempFilePath = tempPath
 			};
+			NormalizeImageOrientation(i.Image);
```
hatton: So, do we really need to Hard-Rotate the image? Would just preserving the exif information work?

JohnThomson: I tried that. The resulting image looks right in Explorer. I didn't actually check whether it looks right in Bloom, because it did NOT look right in the image toolbox.
http://www.daveperrett.com/articles/2012/07/28/exif-orientation-handling-is-a-ghetto/ indicates that browser support for this flag is very patchy. So hard-rotating may also make our books more portable across browsers (and eventually e-book readers).

SIL.Windows.Forms/ClearShare/MetaData.cs
JohnThomson: I kind of agree it's awkward. But I think we need more control than you are suggesting, if we do this at all. Not every ClearShare.metadata write should copy all the TagLib.metadata from the image used to create it. For example, if we apply the same ClearShare.metadata to every file in a bloom book, doing that would make them all appear to have been taken in the same place at the same time, etc.
I suppose I can give Write a boolean which says whether to copy the original metadata and default it false...

JohnThomson: OK, pushed a version with this improved.

hatton: > For example, if we apply the same ClearShare.metadata to every file in a bloom book, doing that would make them all appear to have been taken in the same place at the same time, etc.

In that scenario, we either need to filter what gets into the examplar to just Intellectual Property, or when we apply it, limit the application to IP.

hatton:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[SIL.Windows.Forms.Tests/ImageToolbox/ImageToolboxTests.cs, line 26 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7vzhlLEHPVaX5j1N_E:-K7vzhlLEHPVaX5j1N_F:1506992777) ([raw file](https://github.com/sillsdev/libpalaso/blob/9f2d4b1ce532a7e4f4ac4a47107326035a6c8ae0/SIL.Windows.Forms.Tests/ImageToolbox/ImageToolboxTests.cs#L26)):*
Just a check... what happens if the image was originally ".bmp"? The old code assumed that the output would be a PNG.

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 66 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7w2ZWCglqpOoEFVZtc:-K7w2ZWCglqpOoEFVZtd:-273523368) ([raw file](https://github.com/sillsdev/libpalaso/blob/9f2d4b1ce532a7e4f4ac4a47107326035a6c8ae0/SIL.Windows.Forms/ClearShare/MetaData.cs#L66)):*
TabLib --> TagLib

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 391 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7xLXa9VhAzr-ce4OZp:-K7xLXa9VhAzr-ce4OZq:-1126972048) ([raw file](https://github.com/sillsdev/libpalaso/blob/9f2d4b1ce532a7e4f4ac4a47107326035a6c8ae0/SIL.Windows.Forms/ClearShare/MetaData.cs#L391)):*
Sorry for the delay... I just can't get my head to see this problem the way you're approaching it... it's probably me. I keep returning to this, hoping that a light will turn on. Probably we need to have a conversation.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

JohnThomson:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[SIL.Windows.Forms.Tests/ImageToolbox/ImageToolboxTests.cs, line 26 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7vzhlLEHPVaX5j1N_E:-K86Kj43BQ7gLtpX3yRq:1451517338) ([raw file](https://github.com/sillsdev/libpalaso/blob/9f2d4b1ce532a7e4f4ac4a47107326035a6c8ae0/SIL.Windows.Forms.Tests/ImageToolbox/ImageToolboxTests.cs#L26)):*
I fixed it to do png for everything except jpg (at some point I needed to test something making a jpg from a jpg)

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 365 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7lOub-x1cLO6lo_HqK-r1-365:-K86KUKgMAsc5omhB93F:-736098905) ([raw file](https://github.com/sillsdev/libpalaso/blob/81511e533d2649e6292f0d1b3c836137211769c4/SIL.Windows.Forms/ClearShare/MetaData.cs#L365)):*
I think that's what I've done. It will load all the metadata, but write it all only if Write is passed true for copyAllMetaDataFromOriginal

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 66 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7w2ZWCglqpOoEFVZtc:-K86KR7WuM_tUZO2Ehrp:-499146009) ([raw file](https://github.com/sillsdev/libpalaso/blob/9f2d4b1ce532a7e4f4ac4a47107326035a6c8ae0/SIL.Windows.Forms/ClearShare/MetaData.cs#L66)):*
Done.

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 391 \[r3\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7xLXa9VhAzr-ce4OZp:-K86Kex29U7hehHCFUA1:-1917931301) ([raw file](https://github.com/sillsdev/libpalaso/blob/9f2d4b1ce532a7e4f4ac4a47107326035a6c8ae0/SIL.Windows.Forms/ClearShare/MetaData.cs#L391)):*
If things still don't compute let's talk.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

hatton:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 365 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7lOub-x1cLO6lo_HqK-r1-365:-K86XsRba22Y_VquyATA:1558287266) ([raw file](https://github.com/sillsdev/libpalaso/blob/81511e533d2649e6292f0d1b3c836137211769c4/SIL.Windows.Forms/ClearShare/MetaData.cs#L365)):*
This is what I don't get, API design-wise. If I create a metadata from a file, then do something to it, then write it, I would be very surprised if it did not write all the fields.  It seems surprising to have to tell it "by the way, don't forget what I told you".  Instead, I might expect to have a factory method like 

var ipOnlyMetaData = myMetaData.FromIntellectualPropertyFields();

And then use that. But I'm not certain that I'm not missing something... sometimes I have to actually get into the code directly to get a feel for it. Fee free to tell me if that's the case here :-)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

JohnThomson:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 365 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7lOub-x1cLO6lo_HqK-r1-365:-K86bXf48VDA94WR9Byw:1384029194) ([raw file](https://github.com/sillsdev/libpalaso/blob/81511e533d2649e6292f0d1b3c836137211769c4/SIL.Windows.Forms/ClearShare/MetaData.cs#L365)):*
You make a good point.  What we really want is something like an intelligent decision about which metadata to read and/or write, based on whether the place it is being written is essentially the SAME image (even if compressed or reformatted) as the one it was read from. A lot of new stuff, from GPS coordinates to details about the camera, just makes no sense to copy to all images in a collection. For that reason, and also because I don't know how widely this is used and wanted to largely keep the original behavior as the default, it seemed to make sense to require the writer to explicitly indicate that he is writing to a copy of the image. Equivalently, there could be an explicit request to load all metadata, including stuff very specific to a particular image, with a default of just loading clearshare stuff. But you're right that it's surprising for it not to be the default to both read and write everything...except perhaps because that's such a bad default for applying metadata to multiple images.
Another option is to drop the whole idea. I originally added this as a way to get the image to display properly, but while copying the orientation flag helped gecko, it didn't help the image toolbox itself, which led to the solution of truly rotating the image. It's no longer necessary to this PR to copy any extra metadata. It just seemed a shame to discard all the good information some images have in the process of making a Bloom copy of it, once I found out how easy it was to preserve it. But if the main effect is to complicate the UI and raise dangers of inappropriate copying, it may be better to just drop it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

hatton:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 365 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7lOub-x1cLO6lo_HqK-r1-365:-K8UciSCd1rAIZ3SGDyk:-249070838) ([raw file](https://github.com/sillsdev/libpalaso/blob/81511e533d2649e6292f0d1b3c836137211769c4/SIL.Windows.Forms/ClearShare/MetaData.cs#L365)):*
The only place this is used is Bloom and the Image Toolbox, which is used by WeSay. In particular, we should look at the exemplar feature, which saves you from having to enter the same IP information with each image you import/scan/etc. We would need to make sure that that feature isn't copying this other metadata.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

JohnThomson:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms/ClearShare/MetaData.cs, line 365 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/340#-K7lOub-x1cLO6lo_HqK-r1-365:-K8UuHQL7zEYfOnqK1xG:-941947776) ([raw file](https://github.com/sillsdev/libpalaso/blob/81511e533d2649e6292f0d1b3c836137211769c4/SIL.Windows.Forms/ClearShare/MetaData.cs#L365)):*
Turns out the 'copy exemplar' code is creating and applying the metadata using different code paths so it was not affected by my change. I did have to make one change in Bloom (where it copies metadata to all images in book)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

hatton:
Reviewed 1 of 2 files at r4, 2 of 2 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/340)*

PR #345, opened 2016-01-21
App color palette extender

tombogle:
Added an extender to enable applications to use custom color palettes and have the default color scheme apply in Designer.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/345)


andrew-polk:
As far as I understand it, looks okay to me.  I admit I didn't take the time to really understand everything.  But it is a new class which can't affect any existing code negatively.

If I'm nitpicking, I would suggest squashing your commits and rebasing so as to cleanup the history.

PR #349, opened 2016-02-04
PTX-44984 fix problem with limited users trying to change access to S…

jwickberg:
…ldrCache folder that was created by someone else.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/349)

### DIFF
--- a/SIL.WritingSystems/Sldr.cs
+++ b/SIL.WritingSystems/Sldr.cs
@@ -566,6 +566,9 @@ namespace SIL.WritingSystems
 
 		private static void CreateSldrCacheDirectory()
 		{
+			if (Directory.Exists(SldrCachePath))
+				return;
+
 			DirectoryInfo di = Directory.CreateDirectory(SldrCachePath);
 			if (!Platform.IsLinux && !SldrCachePath.StartsWith(Path.GetTempPath()))
 			{


ermshiperete:
Reviewed 1 of 1 files at r1.

---

*[SIL.WritingSystems/Sldr.cs, line 570 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/349#-K9lneZrbyLGDO5IOfRG:-K9lneZrbyLGDO5IOfRH:1537290609) ([raw file](https://github.com/sillsdev/libpalaso/blob/3badc89fad1a57baeac3afd2e2c200349205ea70/SIL.WritingSystems/Sldr.cs#L570)):*
This changes the behaviour for a user with admin privileges on Windows. Previously if the directory had the wrong permissions we'd fix the permissions, now we don't do it any more if the directory exists.

Also I wonder what happens for a limited user if the permissions are wrong? On Linux we create the directory in LocalAppData which is in the user's home directory. On Windows however we create it in CommonApplicationData, so things are slightly different.

But maybe you tested this already and something sensible happens if we simply return here and the directory is still not writeable after calling this method?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/349)*

jwickberg:

---

*[SIL.WritingSystems/Sldr.cs, line 570 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/349#-K9lneZrbyLGDO5IOfRG:-KA0O_UarTl_Z4vgTD7l:-43262711) ([raw file](https://github.com/sillsdev/libpalaso/blob/3badc89fad1a57baeac3afd2e2c200349205ea70/SIL.WritingSystems/Sldr.cs#L570)):*
What I found is that this method was similar to the CreateGlobalWritingSystemRepositoryDirectory method and that this second method is only called when the directory doesn't exist.

So adding the !Exists check here will make the code for these two folders work the same on Windows.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/349)*

PR #352, opened 2016-02-10
On Linux, remove lock file when mutex goes away

rmunn:
Without this change, classes like Sldr will create a lock file in `/var/lock` and never remove it, leading to problems if a different user later runs Fieldworks on the same computer.

Note that I have not yet tested this; I'll be able to test it tomorrow.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/352)

### DIFF
--- a/SIL.Core/Threading/GlobalMutex.cs
+++ b/SIL.Core/Threading/GlobalMutex.cs
@@ -236,6 +236,11 @@ namespace SIL.Threading
 				}
 			}
 
+			protected override void DisposeManagedResources()
+			{
+				Unlink();
+			}
+
 			protected override void DisposeUnmanagedResources()
 			{
 				Syscall.close(_handle);


StephenMcConnel:
Looks reasonable.  If testing succeeds, then looks good.  :-)

---

Reviewed 1 of 1 files at r1.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 243 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KABSHRbP6slyU60jjKc:-KABSHRcq8ipDLVl4xZj:-1906016582) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L243)):*
Looks reasonable to me.  If your testing is positive, feel free to merge.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-:-KABSXDfX0MZAAnR3AUw:708906203)*

rmunn:
We weren't able to test until today... and unfortunately, testing failed. We swapped in the SIL.Core.dll that TeamCity built with the change, and the lock file continues to persist after Fieldworks exits. More work needed before this can be merged.

:-1:

---

---

*[SIL.Core/Threading/GlobalMutex.cs, line 243 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KABSHRbP6slyU60jjKc:-KAcCW-XtVhK8-CLhlme:1234919573) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L243)):*
Sadly, testing failed: after swapping in the SIL.Core.dll that contains this function, Fieldworks is still not deleting the `/var/lock/SldrCache` file it creates. More debugging needed before this can be merged.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-:-KAcBwBEUR8kP1AggySj:274135572)*

ermshiperete:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 145 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KAcvmW4zEa8Tbh5AGMo:-KAcvmW5oFGR4fm1DSDF:-633456844) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L145)):*
This has nothing to do with your change, but I just noticed it: it would be better to follow the [Disposable pattern](https://msdn.microsoft.com/en-us/library/b1yfkh5e%28v=vs.110%29.aspx) here, i.e. implement a `virtual void Dispose(bool)` method and in `Dispose()` also call `GC.SuppressFinalize(this)`.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 243 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KABSHRbP6slyU60jjKc:-KActLe1oJlAgvOW-nyj:-1690101641) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L243)):*
Does this (or rather: `Dispose()`) get called at all? Also I'm wondering if this should better be done in `DisposeUnmanagedResources`. Hmm, but if we do that then we shouldn't lock on `_syncObject` (because that probably no longer exists if called from the finalizer).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/352)*

rmunn:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 243 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KABSHRbP6slyU60jjKc:-KAmNVFO68PkAOfe-w9D:1304287556) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L243)):*
The fact that `Unlock()` locks on `_syncObject` is precisely why I called it from `DisposeManagedResources()`. Also, it calls `Syscall.unlink(_name)`, and `_name` is a C# string -- so by the time you're in `DisposeUnmanagedResources` it may have already been garbage-collected. If we're going to try to unlink the file in `DisposeUnmanagedResources` then we'll probably need to make a copy of the filename as an unmanaged BSTR or something.

Also, to answer your question -- I'm not certain that the MutexAdapter's `Dispose()` method is getting called at all. But it inherits from `DisposableBase`, which will print a "Missing Dispose() call for (type name)" warning message if Dispose() isn't called -- and I didn't see that warning message printed when I tested this change. But without having stepped through the code in a debugger, I can't be 100% certain.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/352)*

cambell-prince:
You might want to reconcile this PR with PR #357 

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-:-KAmgkw4C4HjPynx-4uU:-2115289800)*

ermshiperete:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 145 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KAcvmW4zEa8Tbh5AGMo:-KAnG1u8G5uJZtbHqrUH:-451139532) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L145)):*
Done in #357

---

*[SIL.Core/Threading/GlobalMutex.cs, line 243 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/352#-KABSHRbP6slyU60jjKc:-KAnFLodKeeE4FLx_sNn:-1339613505) ([raw file](https://github.com/sillsdev/libpalaso/blob/899956424b34854311d6a3d8912135d7c2181030/SIL.Core/Threading/GlobalMutex.cs#L243)):*
You can't trust that you always will get a "Missing Dispose" message when `Dispose()` didn't get called - it depends on whether `Console` gets finalized before or after the undisposed object.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/352)*

rmunn:
PR #357 does exactly the same thing as this PR, plus more. Once #357 is merged, there'll be no more need for this one, so I'm withdrawing this PR.

PR #357, opened 2016-02-17
Properly dispose the SLDR cache mutex

ddaspit:
The static Sldr class uses a global mutex to synchronize access to the SLDR cache folder. Unfortunately, this mutex was not being disposed of properly. In order to fix this issue, I added Initialize() and Cleanup() methods to the Sldr class that must be called by any applications that use the writing systems classes.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/357)

### DIFF
--- a/SIL.Core.Tests/Threading/GlobalMutexTests.cs
+++ b/SIL.Core.Tests/Threading/GlobalMutexTests.cs
@@ -13,7 +13,6 @@ namespace SIL.Tests.Threading
 			{
 				mutex.Unlink();
 				Assert.That(mutex.Initialize(), Is.True);
-				mutex.Unlink();
 			}
 		}
 
@@ -25,7 +24,6 @@ namespace SIL.Tests.Threading
 				mutex1.Initialize();
 				using (var mutex2 = new GlobalMutex("test"))
 					Assert.That(mutex2.Initialize(), Is.False);
-				mutex1.Unlink();
 			}
 		}
 
@@ -38,7 +36,6 @@ namespace SIL.Tests.Threading
 				bool createdNew;
 				using (mutex.InitializeAndLock(out createdNew)) {}
 				Assert.That(createdNew, Is.True);
-				mutex.Unlink();
 			}
 		}
 
@@ -54,7 +51,6 @@ namespace SIL.Tests.Threading
 					using (mutex2.InitializeAndLock(out createdNew)) {}
 					Assert.That(createdNew, Is.False);
 				}
-				mutex1.Unlink();
 			}
 		}
 
@@ -67,7 +63,6 @@ namespace SIL.Tests.Threading
 				{
 					using (mutex.Lock()) {}
 				}
-				mutex.Unlink();
 			}
 		}
 
@@ -81,7 +76,6 @@ namespace SIL.Tests.Threading
 				{
 					using (mutex.Lock()) {}
 				}
-				mutex.Unlink();
 			}
 		}
 	}
--- a/SIL.Core/Threading/GlobalMutex.cs
+++ b/SIL.Core/Threading/GlobalMutex.cs
@@ -238,6 +238,7 @@ namespace SIL.Threading
 
 			protected override void DisposeUnmanagedResources()
 			{
+				Unlink();
 				Syscall.close(_handle);
 			}
 		}
--- a/SIL.Core/Threading/GlobalMutex.cs
+++ b/SIL.Core/Threading/GlobalMutex.cs
@@ -131,7 +131,7 @@ namespace SIL.Threading
 
 		[SuppressMessage("Gendarme.Rules.Correctness", "DisposableFieldsShouldBeDisposedRule",
 			Justification="m_adapter is a reference.")]
-		private sealed class ReleaseDisposable : IDisposable
+		private sealed class ReleaseDisposable : DisposableBase
 		{
 			private readonly IGlobalMutexAdapter _adapter;
 
@@ -140,7 +140,7 @@ namespace SIL.Threading
 				_adapter = adapter;
 			}
 
-			public void Dispose()
+			protected override void DisposeManagedResources()
 			{
 				_adapter.Release();
 			}
@@ -236,9 +236,13 @@ namespace SIL.Threading
 				}
 			}
 
-			protected override void DisposeUnmanagedResources()
+			protected override void DisposeManagedResources()
 			{
 				Unlink();
+			}
+
+			protected override void DisposeUnmanagedResources()
+			{
 				Syscall.close(_handle);
 			}
 		}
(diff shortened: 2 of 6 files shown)


ermshiperete:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 56 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAirWMMVNXsMKqh7GBp:-KAirWMMVNXsMKqh7GBq:1833897463) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L56)):*
Nothing to do with this change, just noticed it now - `IsInitialized` is already defined in the baseclass, so we shouldn't redefine it here.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 143 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAis5WpHM3PeO4TsXQ4:-KAis5WpHM3PeO4TsXQ5:-633456844) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L143)):*
This has nothing to do with your change, but I just noticed it: it would be better to follow the [Disposable pattern](https://msdn.microsoft.com/en-us/library/b1yfkh5e%28v=vs.110%29.aspx) here, i.e. implement a `virtual void Dispose(bool)` method and in `Dispose()` also call `GC.SuppressFinalize(this)`.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 241 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAiqF11oFg58bFVo9xV:-KAiqF11oFg58bFVo9xW:597049417) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L241)):*
Calling `Unlink` locks on `_syncObject` which is a managed object. I'd think that this could cause problems when we forget to call `Dispose()` and this method gets called from the Finalizer which might already have released the memory held by `_syncObject`.

If you think it's safe to do it this way a comment might be good.

---

*[SIL.WritingSystems/Sldr.cs, line 122 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAisnuQ2bDnv85tTjkN:-KAisnuQ2bDnv85tTjkO:282068117) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.WritingSystems/Sldr.cs#L122)):*
Since `Initialize` can be called only once this begs for a `IsInitialized` property - that might make it easier when used in different libraries.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/357)*

ddaspit:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 56 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAirWMMVNXsMKqh7GBp:-KAmCh1YcVzHtsakl3zX:-1158170017) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L56)):*
I'm not sure I understand this comment. The base class for GlobalMutex is DiposableBase, which doesn't define IsInitialized.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/357)*

ddaspit:
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 143 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAis5WpHM3PeO4TsXQ4:-KAmFn2O13PYJjLQT5oT:-499146009) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L143)):*
Done.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 241 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAiqF11oFg58bFVo9xV:-KAmFpkVJPXuTTKhEDff:-499146009) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L241)):*
Done.

---

*[SIL.WritingSystems/Sldr.cs, line 122 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAisnuQ2bDnv85tTjkN:-KAmFv3c4yh3463s-c_A:-499146009) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.WritingSystems/Sldr.cs#L122)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/357)*

ermshiperete:
Reviewed 2 of 2 files at r2.

---

*[SIL.Core/Threading/GlobalMutex.cs, line 56 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/357#-KAirWMMVNXsMKqh7GBp:-KAnCIgLbuR-j4S1T_on:-1371774132) ([raw file](https://github.com/sillsdev/libpalaso/blob/b873ca4db3ef28a18f958cd6b294639c92b516fd/SIL.Core/Threading/GlobalMutex.cs#L56)):*
Oops - I mixed up `IsInitialized` and `IsDisposed`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/357)*

hatton:
@ddaspit First indications are that this change broke Bloom. We now get 
image

tombogle:
It broke Glyssen, too. For now, we’ve added an explicit call to Initialize and that seems to have fixed things, but we wrote to Damien to see if this was an intentional breaking change.

-tom

hatton:
@ddaspit It is unfortunate that this is requiring its own explicit setup/teardown. Should be possible for the SLDR system to manage its own setup, when first used. That would leave teardown... rather unusual, for a library to require that too. Is there a different way this could be done?

PR #368, opened 2016-03-10
Merge feature/LinuxSyslog-2.6 into lfmerge

rmunn:
SyslogLogger will only run on Linux (and Linux-like systems such as OS X, as long as they have a libc library). It uses the syslog() API defined in libc to log messages to the local syslog program.

SyslogExceptionHandler likewise will only run on Linux, since it relies on SyslogLogger to log exceptions to syslog.

This branch is the libpalaso-2.6 version of feature/LinuxSyslog, using the older Palaso.\* namespaces instead of the SIL.\* namespaces that the feature/LinuxSyslog branch uses.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/368)


rmunn:
Now superseded by #373, which is this PR plus the changes that Damien suggested during his review of #369, all squashed into a single commit for better Git history. Closing this PR unmerged.

PR #370, opened 2016-03-11
Fix BL-3222 TempFileForSafeWriting  fails on JAARS network drives

hatton:
Underlying problem is that File.Replace(), fails on JAARS network drives, so this has a fall-back to do the same thing.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/370)

### DIFF
--- a/SIL.Core.Tests/IO/TempFileForSafeWritingTests.cs
+++ b/SIL.Core.Tests/IO/TempFileForSafeWritingTests.cs
@@ -11,8 +11,21 @@ namespace SIL.Tests.IO
 		[Test]
 		public void WriteWasSuccessful_TargetDidNotExist_TargetHasContents()
 		{
+			_WriteWasSuccessful_TargetDidNotExist_TargetHasContents(false);
+		}
+
+		// Simulate situation on the JAARS network in 2016, .net's built-in File.Replace fails on the network drives.
+		[Test]
+		public void ReplaceFails_WriteWasSuccessful_TargetDidNotExist_TargetHasContents()
+		{
+			_WriteWasSuccessful_TargetDidNotExist_TargetHasContents(true);
+		}
+
+		private static void _WriteWasSuccessful_TargetDidNotExist_TargetHasContents(bool simulateVolumeThatCannotHandleFileReplace)
+		{
 			var targetPath = Path.GetTempFileName();
 			var f = new TempFileForSafeWriting(targetPath);
+			f.SimulateVolumeThatCannotHandleFileReplace = simulateVolumeThatCannotHandleFileReplace;
 			File.WriteAllText(f.TempFilePath, "hello");
 			f.WriteWasSuccessful();
 			Assert.That(File.ReadAllText(targetPath), Is.EqualTo("hello"));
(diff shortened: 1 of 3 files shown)


JohnThomson:
Nice work. A few suggestions, but it would be a good improvement as it stands.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[SIL.Core.Tests/IO/TempFileForSafeWritingTests.cs, line 67 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/370#-KCaY8sSw3HrG9MkuAv2:-KCaY8sSw3HrG9MkuAv3:167369434) ([raw file](https://github.com/sillsdev/libpalaso/blob/62c14d939e53bc6932eb2b4d7c10a7fa61b54838/SIL.Core.Tests/IO/TempFileForSafeWritingTests.cs#L67)):*
This is very surprising. It was supposed to be left alone, according to the test name, but we verify that it changed from "old bak" to empty.

---

*[SIL.Core.Tests/IO/TempFileForSafeWritingTests.cs, line 114 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/370#-KCaYiHrq3s5NaJH5dbu:-KCaYiHrq3s5NaJH5dbv:-527886084) ([raw file](https://github.com/sillsdev/libpalaso/blob/62c14d939e53bc6932eb2b4d7c10a7fa61b54838/SIL.Core.Tests/IO/TempFileForSafeWritingTests.cs#L114)):*
Seems like every test is (or should be) duplicated with this flag set.
How about just making a subclass of TempFileForSafeWritingTests, like TempFileForSafeWriting_ReplaceFailsTests? If you encapsulated setting targetPath and making the TempFileForSafeWriting into a virtual setup method, the subclass would differ only by overriding setup() to call base.setup and then f.SimulateVolumeThatCannotHandleFileReplace = true;

---

*[SIL.Core/IO/TempFileForSafeWriting.cs, line 52 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/370#-KCa_OG_r3C87pTmqQ2E:-KCa_OG_r3C87pTmqQ2F:829310936) ([raw file](https://github.com/sillsdev/libpalaso/blob/62c14d939e53bc6932eb2b4d7c10a7fa61b54838/SIL.Core/IO/TempFileForSafeWriting.cs#L52)):*
A slightly more elegant way to handle this would be to encapsulate File.Replace in a virtual method, and have the test code override it to throw. Not a big deal, but I prefer to avoid test-only code in production as far as reasonably possible.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/370#-:-KCaa0I4NXxCnEkCjHwP:-1738890669)*

JohnThomson:
Reviewed 1 of 3 files at r1, 1 of 1 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/libpalaso/370)*

PR #369, opened 2016-03-10
Merge feature/LinuxSyslog into master

rmunn:
SyslogLogger will only run on Linux (and Linux-like systems such as OS X, as long as they have a libc library). It uses the syslog() API defined in libc to log messages to the local syslog program.

SyslogExceptionHandler likewise will only run on Linux, since it relies on SyslogLogger to log exceptions to syslog.

Since all the Unix-only code is contained in SIL.LinuxLogging.dll, there should be no problems on a Windows system, where we simply won't use that DLL.

This code has been used extensively for the past two months in the LfMerge project, and no bugs have been found so far. I believe it's ready to merge into master.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/369)

PR #372, opened 2016-03-24
Add SyslogLogger, exception handler, and tests

rmunn:
SyslogLogger will only run on Linux (and Linux-like systems such as OS X, as long as they have a libc library). It uses the syslog() API defined in libc to log messages to the local syslog program.

SyslogExceptionHandler likewise will only run on Linux, since it relies on SyslogLogger to log exceptions to syslog.

Both classes are contained in the SIL.Linux.Logging DLL, which will only build on Linux (i.e., in the Mono build configurations).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/372)


rmunn:
Note that this is a rework of https://github.com/sillsdev/libpalaso/pull/369, which had grown into several commits as I fixed issues raised by Damien in his code review. This is the same code as that PR, but squashed into a single commit for cleaner merge history. See https://github.com/sillsdev/libpalaso/pull/369 for all the code-review discussions.

rmunn:
Everything builds, and it's been code reviewed already. Merging.

PR #374, opened 2016-03-24
Merge feature/LinuxSyslog-2.6 into libpalaso-2.6

rmunn:
SyslogLogger will only run on Linux (and Linux-like systems such as OS X,
as long as they have a libc library). It uses the syslog() API defined in
libc to log messages to the local syslog program.

SyslogExceptionHandler likewise will only run on Linux, since it relies on
SyslogLogger to log exceptions to syslog.

Both classes are contained in the SIL.Linux.Logging DLL, which will only
build on Linux (i.e., in the Mono build configurations).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/374)


rmunn:
This is a rework of #366 with changes suggested by Damien Daspit during his code review of #369, with all changes squashed into a single commit for cleaner Git history. See #369 for all the changes and comments.

rmunn:
Looks good, and already code reviewed. Merging.

PR #378, opened 2016-05-18
BL-3352 Improve error reporting on missing directory

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/378)

### DIFF
--- a/SIL.Core/IO/FileLocator.cs
+++ b/SIL.Core/IO/FileLocator.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using Microsoft.Win32;
 using SIL.Code;
 using SIL.PlatformUtilities;
@@ -309,7 +310,8 @@ namespace SIL.IO
 			if (Directory.Exists(path))
 				return path;
 
-			foreach (var directoryHoldingFiles in new[] {"", "DistFiles", "common" /*for wesay*/, "src" /*for Bloom*/})
+			var directoriesHoldingFiles = new[] {"DistFiles", "common" /*for wesay*/, "src" /*for Bloom*/};
+			foreach (var directoryHoldingFiles in directoriesHoldingFiles)
 			{
 				path = Path.Combine(FileLocator.DirectoryOfApplicationOrSolution, directoryHoldingFiles);
 				foreach (var part in partsOfTheSubPath)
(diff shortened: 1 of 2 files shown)


SIL.Core/IO/FileLocator.cs
```diff
@@ -309,7 +310,8 @@ public static string GetDirectoryDistributedWithApplication(bool optional, param
 			if (Directory.Exists(path))
 				return path;
 
-			foreach (var directoryHoldingFiles in new[] {"", "DistFiles", "common" /*for wesay*/, "src" /*for Bloom*/})
+			var directoriesHoldingFiles = new[] {"DistFiles", "common" /*for wesay*/, "src" /*for Bloom*/};
+			foreach (var directoryHoldingFiles in directoriesHoldingFiles)
```
JohnThomson: Note, this change removes the empty string from the list of subdirectories to try. Unless I'm misreading the code, having it caused a redundant repeat check of the root directory we just tried. Not having it also makes it easier to produce a nice message.

hatton:

---

*[SIL.Core/IO/FileLocator.cs, line 314 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/378#-KI3MicBgV-0nOCmVYc_-r1-314:-KIE4SfXvTZlFW9Fvixw:936029668) ([raw file](https://github.com/sillsdev/libpalaso/blob/eb8b51836c77537d274b05cba5a1dd35e53aee60/SIL.Core/IO/FileLocator.cs#L314)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary>
> Note, this change removes the empty string from the list of subdirectories to try. Unless I'm misreading the code, having it caused a redundant repeat check of the root directory we just tried. Not having it also makes it easier to produce a nice message.
</details>
Probably *I'm* reading this wrong, but I would have thought that it is the previous code (which just checks this one page) that is redundant, and that by removing the "" from the list, we then don't list it as a place that was checked? I do see that the message would either need to list each directory as starting with the application one, or else have a special case for the "" so that the full directory is listed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/378)*

JohnThomson:

---

*[SIL.Core/IO/FileLocator.cs, line 314 \[r1\]](https://reviewable.io:443/reviews/sillsdev/libpalaso/378#-KI3MicBgV-0nOCmVYc_-r1-314:-KIEVO4hQWk0dQ5hTWgy:689273824) ([raw file](https://github.com/sillsdev/libpalaso/blob/eb8b51836c77537d274b05cba5a1dd35e53aee60/SIL.Core/IO/FileLocator.cs#L314)):*
<details><summary><i>Previously, hatton (John Hatton) wrote…</i></summary>
> Probably *I'm* reading this wrong, but I would have thought that it is the previous code (which just checks this one page) that is redundant, and that by removing the "" from the list, we then don't list it as a place that was checked? I do see that the message would either need to list each directory as starting with the application one, or else have a special case for the "" so that the full directory is listed.
</details>
Yes, it would equally fix the redundancy to remove the preliminary check of the root directory. Code would be shorter but I'm not sure it would be clearer.

Since the message ends up saying "We looked in X and its subdirectories DistFiles, common and src" we don't actually want to list empty-string among the subdirectories. I'm not currently making a list of the full paths.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/378)*

hatton:
OK, clearer trumps shorter :-)

PR #382, opened 2016-08-03
Update language data files

rdb:
This PR updates the following: 
- I noticed the IANA subtag registry was out of date; I updated it to the latest version. 
- Updated LanguageIndex.txt and LanguageCodes.txt to latest from the Ethnologue site, but fixing some incompatibilities with ISO 639-3:
- `[aka] Akan` and `[nor] Norwegian` removed from LanguageCodes.txt, which are macrolanguages according to ISO 639-3.
- Added the following ISO 639-3 sub-languages of the above to both LanguageCodes.txt and LanguageIndex.txt:
  - `[nob] Norwegian Bokmål`
  - `[nno] Norwegian Nynorsk`
  - `[fat] Fanti`
  - `[twi] Twi`
- Removes the unit test related to Akan, which broke (since Akan is no longer listed in LanguageCodes.txt)

There are many languages from ISO 639-3 missing in these files.  It would be good if the other missing codes were added as well.  But these changes should at least be a stopgap measure until that happens.

---

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/382)

### DIFF
--- a/SIL.WritingSystems.Tests/LanguageLookupTests.cs
+++ b/SIL.WritingSystems.Tests/LanguageLookupTests.cs
@@ -142,25 +142,5 @@ namespace SIL.WritingSystems.Tests
 			Assert.True(languages.Any(l => l.Names.Contains("Andaki")));
 			Assert.True(languages.Any(l => l.Names.Contains("Churuba")));
 		}
-
-		[Test]
-		public void SuggestLanguages_Akan_DoesnotCrash()
-		{
-			var lookup = new LanguageLookup();
-			LanguageInfo[] languages = lookup.SuggestLanguages("a").ToArray();
-			Assert.True(languages.Any(l => l.LanguageTag == "ak"));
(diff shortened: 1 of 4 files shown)


ddaspit:
We apologize for taking so long to get to this pull request. Unfortunately, we won't be able to merge this PR as is. We don't want to make any changes to the LanguageIndex.txt and LanguageCodes.txt files, because they are retrieved from the Ethnologue website. The next time these files are updated from the Ethnologue site, it will overwrite the changes that were made in this PR. It looks like you are trying to augment the language code lookup with ISO 639-3 codes that aren't in the LanguageIndex.txt and LanguageCodes.txt files. This is a known issue. We hope to fix this by augmenting the LanguageIndex.txt and LanguageCodes.txt files with the complete list of ISO 639-3 codes so that we have complete coverage of all codes in the language code lookup.

PR #407, opened 2016-10-25
Add new PathUtilities.PathContainsDirectory() method

ermshiperete:
This method checks that the path contains a subdirectory named
'subdir' instead of checking that path contains the string
'subdir'.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/407)

### DIFF
--- a/SIL.Core.Tests/IO/PathUtilitiesTests.cs
+++ b/SIL.Core.Tests/IO/PathUtilitiesTests.cs
@@ -323,6 +323,18 @@ namespace SIL.Tests.IO
 			}
 			PathUtilities.OpenDirectoryInExplorer(path);
 		}
+
+		[TestCase(null, Result = false)]
+		[TestCase("", Result = false)]
+		[TestCase("rect", Result = false)]
+		[TestCase("none", Result = false)]
+		[TestCase("directory", Result = true)]
+		[TestCase("subdir", Result = true)]
+		public bool PathContainsDirectory(string directory)
+		{
+			var path = Path.Combine(Path.GetTempPath(), "some", "directory", "and", "other", "subdir");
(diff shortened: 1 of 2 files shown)


SIL.Core.Tests/IO/PathUtilitiesTests.cs
```diff
@@ -323,6 +323,17 @@ public void OpenDirectoryInExplorer_PathIsADirectoryHasCombiningCharacters_Still
 			}
 			PathUtilities.OpenDirectoryInExplorer(path);
 		}
+
+		[TestCase(null, Result = false)]
+		[TestCase("", Result = false)]
+		[TestCase("directory", Result = true)]
```
hatton: The only "true" test here, "directory", is ambiguous because it appears non-final and final, so it doesn't prove either works, just that one of them works.

ermshiperete:

---

*[SIL.Core.Tests/IO/PathUtilitiesTests.cs, line 329 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/407#-KUwpUcViQLGOvzw1MxD-r1-329:-KUzlFTZcYovXY6Bj_YI:b-qu6n94) ([raw file](https://github.com/sillsdev/libpalaso/blob/eebb5c236635c70ef516024d6a035e75879e4fec/SIL.Core.Tests/IO/PathUtilitiesTests.cs#L329)):*
<details><summary><i>Previously, hatton (John Hatton) wrote…</i></summary>
> The only "true" test here, "directory", is ambiguous because it appears non-final and final, so it doesn't prove either works, just that one of them works.
</details>
Good point. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/407)*

PR #409, opened 2016-11-03
Allow GeckoBrowserAdapter to work when ilrepacked/ilmerged

ermshiperete:
When we ilrepack/ilmerge assemblies the GeckoFxWebBrowserAdapter
class is in the executing assembly, so we try to get it from
there first. (It is important to ilrepack both SIL.Windows.Forms
and GeckoBrowserAdapter assemblies, otherwise the interface
definition will be different)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/409)


tombogle:
Looks okay to me, but I don't completely follow it all. Since we're going to try to use ILRepack in Paratext instead of Glyssen, maybe this won't be needed. On the other hand, maybe it's a good enhancement anyway to future-proof this in case something *does* need to use ILRepack to embed the GeckoBrowser stuff.

PR #418, opened 2016-11-29
Made WriteMinorEvent more robust

tombogle:
Do not attempt to treat the message as a format string if no args are supplied. Report the problem if the format is attempted and fails.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/418)

### DIFF
--- a/SIL.Core/Reporting/Logger.cs
+++ b/SIL.Core/Reporting/Logger.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Text;
 using SIL.IO;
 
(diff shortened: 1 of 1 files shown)


SIL.Core/Reporting/Logger.cs
```diff
@@ -481,7 +482,19 @@ private void WriteMinorEventCore(string message, params object[] args)
 						m_minorEvents.Remove(0, cutoff);
 					}
 					m_minorEvents.Append(DateTime.Now.ToLongTimeString() + "\t");
-					m_minorEvents.AppendFormat(message, args);
+					if (args.Any())
```
hatton: In case anybody was wondering if args could be null, Section 7.4.1 of the C# Language Specification says
> In particular, note that an empty array is created when there are zero arguments given for the parameter array.

PR #422, opened 2016-12-13
PTX-49759 prevent corrupted .ldml file in global repository from stopping initialization

jwickberg:
A bad .ldml file was stopping the global repository from initializing. The bad file was most likely caused by a system crash since it was all nulls.

Changed code to catch XmlException's that occur when reading the .ldml file and to rename the file to be a .ldml.bad file (thought renaming was betting than deleting in case the file was needed for investigation).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/422)

### DIFF
--- a/SIL.WritingSystems.Tests/GlobalWritingSystemRepositoryTests.cs
+++ b/SIL.WritingSystems.Tests/GlobalWritingSystemRepositoryTests.cs
@@ -21,6 +21,24 @@ namespace SIL.WritingSystems.Tests
 		}
 
 		[Test]
+		public void Initialize_SkipsBadFile()
+		{
+			using (var e = new TemporaryFolder("GlobalWritingSystemRepositoryTests"))
+			{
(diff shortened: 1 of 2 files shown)


tombogle:
Might there be some way to inform the caller (or the caller of the caller...) of the problem. Just backing up the file doesn't alert anyone to the possibility of a real problem that needs to be addressed.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/422#-:-KYt3Z51N6_jsWd20_gj:brbpcwg)*

jwickberg:
I thought about it, but didn't see any good way for the alert the caller besides an exception.

This solution seemed better than just causing everything to break because of one bad file.

andrew-polk:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/422#-:-KYu-qu3cbup-UEmohaT:bnfp4nl)*

PR #424, opened 2016-12-13
Add option to unit test build task to fail if a test fails (BL-4087)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/424)

### DIFF
--- a/Palaso.MSBuildTasks/UnitTestTasks/NUnit.cs
+++ b/Palaso.MSBuildTasks/UnitTestTasks/NUnit.cs
@@ -14,6 +14,12 @@ namespace Palaso.BuildTasks.UnitTestTasks
 	/// </summary>
 	public class NUnit : TestTask
 	{
+		public NUnit()
+		{
+			// For NUnit, a negative exit code indicates a problem running the tests
+			_failTaskIfNegativeExitCode = true;
+		}
+
 		/// <summary>
 		/// Gets or sets the full path to the NUnit assemblies (test DLLs).
 		/// </summary>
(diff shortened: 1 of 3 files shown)


Palaso.MSBuildTasks/UnitTestTasks/NUnit3.cs
```diff
@@ -11,6 +11,14 @@ namespace Palaso.BuildTasks.UnitTestTasks
 	/// </summary>
 	public class NUnit3 : NUnit
 	{
+		public NUnit3()
+		{
+			// REVIEW: This should probably be true for NUnit also, but changing
+			// the logic there could potentially cause unexpected results for existing
+			// callers whereas the NUnit3 task is new enough, I think we are okay.
+			FailTaskIfAnyTestsFail = true;
```
ermshiperete: I'm not convinced that we want to always fail on test failures. At least on Jenkins this eliminates the distinction between failing tests (unstable) and failing build (failed). If TC needs this then we should make it conditional on the `TeamCity` property.

Palaso.MSBuildTasks/UnitTestTasks/NUnit.cs
```diff
@@ -76,8 +82,29 @@ protected override string GetWorkingDirectory()
...
+				_failTaskIfAnyTestsFail = value;
+
+				// For NUnit, a positive exit code indicates the number of failed tests
+				_failTaskIfPositiveExitCode = value;
+			}
+		}
+
 		// make this nullable so we have a third state, not set
 		private bool? _testInNewThread;
+		private bool _failTaskIfAnyTestsFail;
```
ermshiperete: Why do we need this property? It doesn't seem to be used anywhere.

PR #435, opened 2017-01-16
Make DeleteDirectoryRobust mow down readonly files, by default

hatton:
Also make the TestUtilities equivalent do the same, and make the RobustIO version notice if DeleteDirectoryRobust fails.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/435)

### DIFF
--- a/SIL.Core.Tests/IO/DirectoryUtilitiesTests.cs
+++ b/SIL.Core.Tests/IO/DirectoryUtilitiesTests.cs
@@ -589,5 +589,31 @@ namespace SIL.Tests.IO
 				File.Delete(fileName);
 			}
 		}
+
+		[Test]
+		public void DeleteDirectoryRobust_ContainsReadOnlyFile_StillRemoves()
+		{
+			using (var tempDir = new TemporaryFolder("DeleteDirectoryRobust_ContainsReadOnlyFile_StillRemoves"))
+			{
+				var fileName = tempDir.Combine("tempFile.txt");
+				File.WriteAllText(fileName, @"Some test text");
+				new System.IO.FileInfo(fileName).IsReadOnly = true;
+				DirectoryUtilities.DeleteDirectoryRobust(tempDir.Path);
+				Assert.IsFalse(Directory.Exists(tempDir.Path), "Did not delete directory");
+			}
+		}
+
+		[Test]
+		public void DeleteDirectoryRobust_NoOverrideContainsReadOnlyFile_ReturnsFalse()
+		{
+			using (var tempDir = new TemporaryFolder("DeleteDirectoryRobust_NoOverrideContainsReadOnlyFile_ReturnsFalse"))
+			{
+				var fileName = tempDir.Combine("tempFile.txt");
+				File.WriteAllText(fileName, @"Some test text");
+				new System.IO.FileInfo(fileName).IsReadOnly = true;
+				Assert.IsFalse(DirectoryUtilities.DeleteDirectoryRobust(tempDir.Path, overrideReadOnly:false));
+				Assert.IsTrue(Directory.Exists(tempDir.Path), "Did not expect it to delete directory because of the readonly file");
+			}
+		}
 	}
 }
--- a/SIL.Core/IO/DirectoryUtilities.cs
+++ b/SIL.Core/IO/DirectoryUtilities.cs
@@ -210,7 +210,7 @@ namespace SIL.IO
 		/// This method uses all the tricks to do its best.
 		/// </summary>
 		/// <returns>returns true if the directory is fully deleted</returns>
-		public static bool DeleteDirectoryRobust(string path)
+		public static bool DeleteDirectoryRobust(string path, bool overrideReadOnly=true)
 		{
 			// ReSharper disable EmptyGeneralCatchClause
 
@@ -239,9 +239,10 @@ namespace SIL.IO
 					{
 						try
 						{
-							/* we could do this too, but it's dangerous
-							 *  File.SetAttributes(filePath, FileAttributes.Normal);
-							 */
+							if(overrideReadOnly)
+							{
+								File.SetAttributes(filePath, FileAttributes.Normal);
+							}
 							File.Delete(filePath);
 						}
 						catch (Exception)
@@ -254,7 +255,7 @@ namespace SIL.IO
 					}
 
 				}
-				catch (Exception)//yes, even these simple queries can throw exceptions, as stuff suddenly is deleted base on our prior request
+				catch (Exception)//yes, even these simple queries can throw exceptions, as stuff suddenly is deleted based on our prior request
 				{
 				}
 				//sleep and let some OS things catch up
--- a/SIL.Core/IO/DirectoryUtilities.cs
+++ b/SIL.Core/IO/DirectoryUtilities.cs
@@ -213,7 +213,12 @@ namespace SIL.IO
 		public static bool DeleteDirectoryRobust(string path, bool overrideReadOnly=true)
 		{
 			// ReSharper disable EmptyGeneralCatchClause
-
+#if __MonoCS__
+			// The Mono runtime deletes readonly files and directories that contain readonly files.
+			// This violates the MSDN specification of Directory.Delete and File.Delete.
+			if (!overrideReadOnly && DirectoryContainsReadOnly(path))
+					return false;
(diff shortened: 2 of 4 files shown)


andrew-polk:
Reviewed 4 of 4 files at r1.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 213 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KagYRt3YdwFLE-bsn1s:-KagYRt3YdwFLE-bsn1t:b-ynm2w4) ([raw file](https://github.com/sillsdev/libpalaso/blob/15b48084d50f937bad2c238d41e0cc4c15982517/SIL.Core/IO/DirectoryUtilities.cs#L213)):*
> overrideReadOnly=true

Seems like it would be better to default to false. It is possible an existing caller is relying on this method *not* to delete when there are readonly files. This seems especially possible given the comment you removed (i.e. someone considered implementing it the way it now works and specifically decided not to.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

hatton:

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 213 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KagYRt3YdwFLE-bsn1s:-KagxqWEW8Vat9flRV_7:bten50t) ([raw file](https://github.com/sillsdev/libpalaso/blob/15b48084d50f937bad2c238d41e0cc4c15982517/SIL.Core/IO/DirectoryUtilities.cs#L213)):*

> overrideReadOnly=true

Seems like it would be better to default to false. It is possible an existing caller is relying on this method *not* to delete when there are readonly files. This seems especially possible given the comment you removed (i.e. someone considered implementing it the way it now works and specifically decided not to.)
</blockquote></details>

Yeah... I judge that is far more likely to be bug than intentional. As further evidence, in Windows explorer, if you delete a read only file, or delete a folder that has a read only file, it is just deleted, without discussion with the user.  So that tells us that Microsoft means for "read only" to mean something like "don't edit, but feel free to delete". I'll send out a query to the email list and see if anyone might be using the old behavior intentionally.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

andrew-polk:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 295 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KaheUAfNchwWPVJoEE6:-KaheUAg1SZo6FBqUsgp:b-o8wsmj) ([raw file](https://github.com/sillsdev/libpalaso/blob/68d3b14f1607f93c983f47adf2e878a0d1dffd01/SIL.Core/IO/DirectoryUtilities.cs#L295)):*
> #if __MonoCS__

Is there a reason to have this method be for Mono only?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

hatton:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 295 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KaheUAfNchwWPVJoEE6:-KahgtbOkvyX41Tco9Dz:biw32vr) ([raw file](https://github.com/sillsdev/libpalaso/blob/68d3b14f1607f93c983f47adf2e878a0d1dffd01/SIL.Core/IO/DirectoryUtilities.cs#L295)):*

> #if __MonoCS__

Is there a reason to have this method be for Mono only?
</blockquote></details>

There is... if we use it for both, then we're *really* changing the behavior. SIgh. I hate it when little fixes turn into huge operations. If we want the same behavior on both, then, because mono FIle.Delete ignores readonly, we'll have to have the method rewritten.

Now I'm leaning towards removing this override option and all the complications it brings out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

andrew-polk:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 295 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KaheUAfNchwWPVJoEE6:-KahjV8wM2y-JG4W4vMT:btl349m) ([raw file](https://github.com/sillsdev/libpalaso/blob/68d3b14f1607f93c983f47adf2e878a0d1dffd01/SIL.Core/IO/DirectoryUtilities.cs#L295)):*

There is... if we use it for both, then we're *really* changing the behavior. SIgh. I hate it when little fixes turn into huge operations. If we want the same behavior on both, then, because mono FIle.Delete ignores readonly, we'll have to have the method rewritten.

Now I'm leaning towards removing this override option and all the complications it brings out.
</blockquote></details>

I'm not questioning whether the call to the method should be Mono only, only the implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

ermshiperete:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 295 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KaheUAfNchwWPVJoEE6:-Kai-gHkdU509nhUIhM1:bui9o5v) ([raw file](https://github.com/sillsdev/libpalaso/blob/68d3b14f1607f93c983f47adf2e878a0d1dffd01/SIL.Core/IO/DirectoryUtilities.cs#L295)):*

I'm not questioning whether the call to the method should be Mono only, only the implementation.
</blockquote></details>

Can we use a runtime check instead (something like Platform.IsLinux)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

jasonleenaylor:
FWIW, I am not aware of any code that expected this to fail when read only files were present. LGTM

hatton:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 295 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KaheUAfNchwWPVJoEE6:-KasjA9vi36kCkdWqglq:b-legupu) ([raw file](https://github.com/sillsdev/libpalaso/blob/68d3b14f1607f93c983f47adf2e878a0d1dffd01/SIL.Core/IO/DirectoryUtilities.cs#L295)):*

Can we use a runtime check instead (something like Platform.IsLinux)?
</blockquote></details>

Steve marked it mono only because it is only used by mono, it's not public, etc.

Eberhard, I started to make that change for you, but didn't because it's not clear to me that it is Linux behavior he is working around, but rather a difference in behavior of the mono implementation of File.Delete.  Maybe a good thing to do I'll leave it to you mono guys.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

ermshiperete:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 295 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/435#-KaheUAfNchwWPVJoEE6:-KauoVSRR2bFwPnpQSNv:br425rw) ([raw file](https://github.com/sillsdev/libpalaso/blob/68d3b14f1607f93c983f47adf2e878a0d1dffd01/SIL.Core/IO/DirectoryUtilities.cs#L295)):*

Steve marked it mono only because it is only used by mono, it's not public, etc.

Eberhard, I started to make that change for you, but didn't because it's not clear to me that it is Linux behavior he is working around, but rather a difference in behavior of the mono implementation of File.Delete.  Maybe a good thing to do I'll leave it to you mono guys.
</blockquote></details>

OK, thanks. I think it's a Linux behaviour because of the difference in the underlying file system where the Windows ReadOnly attribute doesn't exactly translate to something on Linux. I'm pretty sure if we'd run the same code in Mono on Windows it'd work. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/435)*

PR #437, opened 2017-01-19
PTX-50439 character sets for crossrefs and versesegments need to be to reated same as footnotes since order needs to be preserved

jwickberg:
The cross reference callers and verse segments need to be ordered sequences, so I changed the CharacterSetDefinition code to treat these the same way as it was already treating footnote callers.

Main change is to replace the hard coded check for "footnotes" to a new boolean IsSequenceType property.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/437)


andrew-polk:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/437#-:-KarXv5FGrgq141kGEOF:bnfp4nl)*

PR #440, opened 2017-01-23
Normalize line endings when writing XML files

ermshiperete:
The previous code assumed that all line endings are written as \r\n. However that was not the case when the data that we wrote contained line endings. This change forces line ending normalization so that line endings always end up as \r\n even when in the data we write. That helps with FW and LfMerge S/R where the ParserParameters contain line endings which will otherwise introduce a difference between Windows and Linux.

This change also allows to pass a `NewLineHandling` parameter when calling `CreateXmlWriterSettings`. Passing `NewLineHandling.None` will keep the previous behavior and is needed for Lift where one test expects that the line endings stay as they are.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/440)

### DIFF
--- a/Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs
+++ b/Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs
@@ -1744,6 +1744,8 @@ namespace Palaso.DictionaryServices.Tests.Lift
 		[Test]
 		public void Add_TextWithSpanAndMeaningfulWhiteSpace_FormattingAndWhitespaceIsUntouched()
 		{
+			// REVIEW (EberhardB): does it really make sense to preserve the line endings? It seems
+			// that for a text node line endings should be standardized.
 			const string formattedText = "\rThis's <span href=\"reference\">\n is a\t\t\n\r\t span</span> with annoying whitespace!\r\n";
 			const string expected = "<form\r\n\tlang=\"de\">\r\n\t<text>" + formattedText + "</text>\r\n</form>";
 			using (var session = new LiftExportAsFragmentTestSession())
--- a/Palaso.DictionaryServices/Lift/LiftWriter.cs
+++ b/Palaso.DictionaryServices/Lift/LiftWriter.cs
@@ -46,7 +46,7 @@ namespace Palaso.DictionaryServices.Lift
 		///
 		/// </summary>
 		/// <param name="path"></param>
-		/// <param name="includeByteOrderMark">PrinceXML (at least v7 chokes if given a BOM, Lexique Pro chokes without it) </param>
+		/// <param name="byteOrderStyle">PrinceXML (at least v7 chokes if given a BOM, Lexique Pro chokes without it) </param>
 		public LiftWriter(string path, ByteOrderStyle byteOrderStyle)
 			: this()
 		{
@@ -61,7 +61,7 @@ namespace Palaso.DictionaryServices.Lift
 		public LiftWriter(StringBuilder builder, bool produceFragmentOnly): this()
 		{
 			_writer = XmlWriter.Create(builder, CanonicalXmlSettings.CreateXmlWriterSettings(
-				produceFragmentOnly ? ConformanceLevel.Fragment : ConformanceLevel.Document)
+				produceFragmentOnly ? ConformanceLevel.Fragment : ConformanceLevel.Document, NewLineHandling.None)
 			);
 			if (!produceFragmentOnly)
 			{
--- a/Palaso.Tests/Xml/CanonicalXmlSettingsTests.cs
+++ b/Palaso.Tests/Xml/CanonicalXmlSettingsTests.cs
@@ -32,11 +32,7 @@ namespace Palaso.Tests.Xml
 		private static void CheckReaderSettings(XmlReaderSettings settings, ConformanceLevel expectedConformanceLevel)
 		{
 			Assert.IsFalse(settings.CheckCharacters);
-#if NET_4_0 && !__MonoCS__
 			Assert.AreEqual(DtdProcessing.Parse, settings.DtdProcessing);
-#else
-			Assert.IsTrue(settings.ProhibitDtd);
-#endif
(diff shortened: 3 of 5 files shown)


cambell-prince:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs, line 1749 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-KbJkNWicIVdE5JqCSc7:-KbJkNWicIVdE5JqCSc8:bzhh3um) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9bed8b0c133b645c1ef45ae899f9f5570cda059/Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs#L1749)):*
> ```Smalltalk
> 			// REVIEW (EberhardB): does it really make sense to preserve the line endings? It seems
> 			// that for a text node line endings should be standardized.
> 			const string formattedText = "\rThis's <span href=\"reference\">\n is a\t\t\n\r\t span</span> with annoying whitespace!\r\n";
> ```

Line endings that are persisted to disk need to have the same form regardless of platform.  We chose to do this rather than relying on mercurial and plugins / config to do the right thing.

---

*[Palaso.DictionaryServices/Lift/LiftWriter.cs, line 65 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-KbJkpsMeWu2e9Gs8ggl:-KbJkpsMeWu2e9Gs8ggm:b-5ohv89) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9bed8b0c133b645c1ef45ae899f9f5570cda059/Palaso.DictionaryServices/Lift/LiftWriter.cs#L65)):*
> ```Smalltalk
> 			_writer = XmlWriter.Create(builder, CanonicalXmlSettings.CreateXmlWriterSettings(
> 				produceFragmentOnly ? ConformanceLevel.Fragment : ConformanceLevel.Document, false)
> 			);
> ```

I think this would be better expressed as a type (enum) to know what setting is being set.    Or do we rely on intellisense in the editor for that now? :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/440)*

cambell-prince:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/440)*

ermshiperete:
Review status: 2 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs, line 1749 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-KbJkNWicIVdE5JqCSc7:-KbK3jhrGUcfZNcWqIT_:b-a7m9p) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9bed8b0c133b645c1ef45ae899f9f5570cda059/Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs#L1749)):*
<details><summary><i>Previously, cambell-prince (Cambell) wrote…</i></summary><blockquote>

Line endings that are persisted to disk need to have the same form regardless of platform.  We chose to do this rather than relying on mercurial and plugins / config to do the right thing.
</blockquote></details>

Yes, but wouldn't it make more sense to normalize to \r\n then instead of persisting whatever we get in the data?

---

*[Palaso.DictionaryServices/Lift/LiftWriter.cs, line 65 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-KbJkpsMeWu2e9Gs8ggl:-KbK6QxgOWTehIfcWhzR:bcvjgzp) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9bed8b0c133b645c1ef45ae899f9f5570cda059/Palaso.DictionaryServices/Lift/LiftWriter.cs#L65)):*
<details><summary><i>Previously, cambell-prince (Cambell) wrote…</i></summary><blockquote>

I think this would be better expressed as a type (enum) to know what setting is being set.    Or do we rely on intellisense in the editor for that now? :-)
</blockquote></details>

Good idea. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/440)*

cambell-prince:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs, line 1749 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-KbJkNWicIVdE5JqCSc7:-KbKRZkW-HPD1TpSceSc:b66utvl) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9bed8b0c133b645c1ef45ae899f9f5570cda059/Palaso.DictionaryServices.Tests/Lift/LiftWriterTests.cs#L1749)):*

Yes, but wouldn't it make more sense to normalize to \r\n then instead of persisting whatever we get in the data?
</blockquote></details>

The line endings are standardized "\r\n", but the internal white space is meaningful and persisted in this fashion by FLEx.  So, this was done in order to merge well with FLEx from an independent implementation of LIFt.  Not saying its self consistent, but it is consistent with what it has to merge with.  All in the name of reducing white space differences.  So, normalize to \r\n on non meaningful white space.  Meaningful white space is left as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/440)*

cambell-prince:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-:-KbKUGF4oDvJQ4eURNp8:bnfp4nl)*

megahirt:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/440#-:-KbNcY2e5InlCpjvSDXk:bnfp4nl)*

PR #471, opened 2017-03-29
Fix two issues with copying directories

ermshiperete:
- pass `true` to File.Copy so that existant files get overwritten
- don't hide exceptions for subdirectories but do as the method
  name promises and throw the exception

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/471)

### DIFF
--- a/SIL.Core/IO/DirectoryUtilities.cs
+++ b/SIL.Core/IO/DirectoryUtilities.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) 2017 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+using System;
 using System.IO;
 using System.Linq;
 using System.Security.AccessControl;
@@ -8,7 +10,7 @@ using SIL.Reporting;
 
 namespace SIL.IO
 {
-	public class DirectoryUtilities
+	public static class DirectoryUtilities
 	{
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


andrew-polk:
Reviewed 1 of 1 files at r1.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 110 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/471#-KgPGhqwIJSs-xU9yhmx:-KgPGhqwIJSs-xU9yhmy:b-4izlai) ([raw file](https://github.com/sillsdev/libpalaso/blob/c0188cce438c12ffd741e2b5832294e9be635cda/SIL.Core/IO/DirectoryUtilities.cs#L110)):*
> , true

It seems at least possible that the exception the client is expecting is the case when the file exists.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/471)*

ermshiperete:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/DirectoryUtilities.cs, line 110 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/471#-KgPGhqwIJSs-xU9yhmx:-KgPJcgD8QsZBPSDwN9x:bd4f6fy) ([raw file](https://github.com/sillsdev/libpalaso/blob/c0188cce438c12ffd741e2b5832294e9be635cda/SIL.Core/IO/DirectoryUtilities.cs#L110)):*

> , true

It seems at least possible that the exception the client is expecting is the case when the file exists.
</blockquote></details>

Good point. Now defaulting to `false` to keep the previous behaviour.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/471)*

andrew-polk:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/471#-:-KgPLuyv7y7E6-vMGK7K:bnfp4nl)*

PR #474, opened 2017-04-17
Add info about a language's primary country

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/474)

### DIFF
--- a/SIL.WritingSystems.Tests/LanguageLookupTests.cs
+++ b/SIL.WritingSystems.Tests/LanguageLookupTests.cs
@@ -1,5 +1,6 @@
 ﻿using System.Linq;
 using NUnit.Framework;
+using SIL.Extensions;
 
 namespace SIL.WritingSystems.Tests
 {
@@ -95,6 +96,37 @@ namespace SIL.WritingSystems.Tests
 			Assert.That(lookup.SuggestLanguages("english").First().Names.Count(s => s == "English"), Is.EqualTo(1));
 		}
 
+		[TestCase("en", "United Kingdom")] // a typical result
+		[TestCase("ro", "Romania")] // even more typical (and different from langInfo.Countries.First()).
+		[TestCase("zrp", "France")] // a three-letter code that has a region
+		[TestCase("xak", "Venezuala")] // two special cases, the countries without regions and with >1 country
+		[TestCase("itd", "Indonesia")]
+		[TestCase("qaa", "?")] // a special case
+		[TestCase("bua", "Russian Federation")] // no region, but does have a unique country
+		public void FindsCorrectPrimaryCountry(string code, string primaryCountry)
+		{
+			var lookup = new LanguageLookup();
+			var lang = lookup.GetLanguageFromCode(code);
+			Assert.That(lang.PrimaryCountry, Is.EqualTo(primaryCountry));
+		}
+
+		/// <summary>
+		/// At the time I wrote these tests, only the indicated two languages had more than one
+		/// country and lacked a region specification to disambigute the primary country.
+		/// This test is designed to catch a change in that situation when the language data
+		/// tables are updated.
+		/// </summary>
+		[Test]
+		public void AllExpectedLanguagesHaveUniquePrimaryCountries()
+		{
+			var languagesWithoutRegions = new LanguageLookup().LanguagesWithoutRegions();
+			var languagesWithAmbiguousPrimaryCountry =
+				languagesWithoutRegions.Where(l => l.Countries.Count() > 1);
+			foreach (var lang in languagesWithAmbiguousPrimaryCountry)
+			{
+				Assert.That(lang.LanguageTag, Is.EqualTo("xak").Or.EqualTo("itd"));
+			}
+		}
 
 		[Test]
 		public void SuggestLanguages_GivenUnambigous3LetterCode_ReturnsLanguage()
--- a/SIL.WritingSystems/LanguageInfo.cs
+++ b/SIL.WritingSystems/LanguageInfo.cs
@@ -13,6 +13,13 @@ namespace SIL.WritingSystems
 		public string ThreeLetterTag { get; set; }
 		public bool IsMacroLanguage { get; set; }
 
+		/// <summary>
+		/// In a LanguageInfo retrieved from LanguageLookup, this will be the region (if any)
+		/// indicated for that language in the LanguageCodes resource.
+		/// </summary>
+		/// <remarks> Over 1000 languages return null.</remarks>
+		public RegionSubtag PrimaryRegion { get; internal set; }
+
 		public IList<string> Names
 		{
 			get { return _names; }
@@ -21,6 +28,33 @@ namespace SIL.WritingSystems
 		public ISet<string> Countries
 		{
 			get { return _countries; }
+		}
+
+		/// <summary>
+		/// The primary country where the language is spoken. In a LanguageInfo retrieved from
+		/// LanguageLookup, this should match the main title in the
+		/// Ethnologue website, where it says things like Romanian//A language of Romania
+		/// or English//A language of United Kingdom
+		/// </summary>
+		public string PrimaryCountry
+		{
+			get
+			{
+				if (PrimaryRegion != null)
+					return PrimaryRegion.Name;
+				// Two special cases for the only languages in our database that have two countries
+				// but no specified region. Following what the actual ethnologue.com site does.
+				if (LanguageTag == "itd")
+					return "Indonesia";
+				if (LanguageTag == "xak")
+					return "Venezuala";
+				// All remaining languages have a single country, so just return it.
+				// It might be an empty string.
+				if (Countries.Any())
+					return Countries.First();
+				// For completeness, but this currently never happens.
+				return "";
+			}
 		}
 
 		/// <summary>
--- a/SIL.WritingSystems/LanguageLookup.cs
+++ b/SIL.WritingSystems/LanguageLookup.cs
@@ -58,6 +58,39 @@ namespace SIL.WritingSystems
 				foreach (string langname in language.Names)
 					GetOrCreateListFromName(langname).Add(language);
 			}
+
+			var langData = LanguageRegistryResources.LanguageCodes.Replace("\r\n", "\n")
+				.Split(new[] {"\n"}, StringSplitOptions.RemoveEmptyEntries);
+			foreach (var languageCode in langData)
+			{
+				var data = languageCode.Split(new[] {'\t'}, StringSplitOptions.RemoveEmptyEntries);
+				if (data.Length < 2)
+					continue;
+				var langCode = data[0];
+				var countryCode = data[1];
+				var lang = this.GetLanguageFromCode(langCode);
+				if (lang == null)
+					continue;
+				RegionSubtag region;
+				if (!StandardSubtags.RegisteredRegions.TryGet(countryCode, out region))
+					continue;
+				lang.PrimaryRegion = region;
+			}
+		}
+
+		/// <summary>
+		///  For testing; used to detect if we need more special cases in LanguageInfo.PrimaryCountry.
+		/// </summary>
+		/// <returns></returns>
+		internal List<LanguageInfo> LanguagesWithoutRegions()
+		{
+			var result = new List<LanguageInfo>();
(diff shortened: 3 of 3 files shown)


glasseyes:
Thanks for working on this John. As ChrisHi says in PR 475 "We'd like to see this logic of adding country to the search be a part of the index itself, as opposed to being bolted on after the index is already created. The search facility has been simplified to just search an index, and not generate the index (this is now done separately)."

I've indicated below where it can be added to LanguageData/NewLanguageIndex.txt. Also, it needs to be written into the index in NewLanguageIndex.WriteIndex and NewLanguageIndex.WriteJson.

Then it needs to be read back again into the LanguageInfo objects in the LanguageLookup() constructor

---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 21 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz4Yks_Zd_SdQW8FU7:-Khz4YktXzkEVoiWVnG1:blxdrov) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L21)):*
> ```Smalltalk
> 		/// </summary>
> 		/// <remarks> Over 1000 languages return null.</remarks>
> 		public RegionSubtag PrimaryRegion { get; internal set; }
> ```

Do we need PrimaryRegion as well as PrimaryCountry? cf When the list of countries is created it converts the region codes to country names

---

*[SIL.WritingSystems/LanguageInfo.cs, line 50 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz8_Hm5iaJYxhQxg8e:-Khz8_Hm5iaJYxhQxg8f:b-u6rot1) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L50)):*
> ```Smalltalk
> 					return "Indonesia";
> 				if (LanguageTag == "xak")
> 					return "Venezuala";
> ```

Venezuela?

---

*[SIL.WritingSystems/LanguageInfo.cs, line 56 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz8VbKaTy8OqnRU9sC:-Khz8VbLTcONvXfVEEi8:bqtp22t) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L56)):*
> ```Smalltalk
> 					return Countries.First();
> 				// For completeness, but this currently never happens.
> 				return "";
> ```

These special cases should be in the place that creates the LanguageDataIndex, like assigned at the same time as the region

---

*[SIL.WritingSystems/LanguageLookup.cs, line 80 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz75iRyWoIFFa246Wx:-Khz75iRyWoIFFa246Wy:b-sp075s) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageLookup.cs#L80)):*
> ```Smalltalk
> 			}
> 		}
> ```

The above code could be adapted for LanguageData/NewLanguageIndex.cs after the localisation loop

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-:-Khz5KNc9l921q8z0hvu:b9trw01)*

JohnThomson:
I haven't read those comments yet, but I don't see the connection. Are you saying that we should completely change the LanguageLookup constructor to somehow use the data written out by NewLanguageIndex instead of the resources it now reads? Are those resources somehow created by NewLanguageIndex?

---

Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 21 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz4Yks_Zd_SdQW8FU7:-Ki0LJLDYoDG3tnMUJCK:bmfo3to) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L21)):*
<details><summary><i>Previously, glasseyes wrote…</i></summary><blockquote>

Do we need PrimaryRegion as well as PrimaryCountry? cf When the list of countries is created it converts the region codes to country names
</blockquote></details>

We could do this immediate function without it. It seemed a natural implementation to me since the file contains  region codes and the code base contains region objects to keep track of the complete region object that is specified by the code.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 50 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz8_Hm5iaJYxhQxg8e:-Ki0I4KzGGFgKltEfj74:b-1ilzry) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L50)):*
<details><summary><i>Previously, glasseyes wrote…</i></summary><blockquote>

Venezuela?
</blockquote></details>

As the comment says, ethnologue.com describes Maku (xak) as  "a language of Venezuela" so I went with that.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 56 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz8VbKaTy8OqnRU9sC:-Ki0IXOfwKuAEjfuZRDm:b-l74oy4) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L56)):*
<details><summary><i>Previously, glasseyes wrote…</i></summary><blockquote>

These special cases should be in the place that creates the LanguageDataIndex, like assigned at the same time as the region
</blockquote></details>

Why? These languages, like many others, don't have a region specified (that's why I later added PrimaryCountry as an additional new property) so there's nothing to store in the original data.

---

*[SIL.WritingSystems/LanguageLookup.cs, line 80 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz75iRyWoIFFa246Wx:-Ki0JZr9lhO8l2eqBpjr:b-p90r8p) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageLookup.cs#L80)):*
<details><summary><i>Previously, glasseyes wrote…</i></summary><blockquote>

The above code could be adapted for LanguageData/NewLanguageIndex.cs after the localisation loop
</blockquote></details>

I don't know why it should be (partly because whoever wrote NewLanguageIndex completely failed to write a class comment to describe what the class is about and what responsibilities it has). It seems its only purpose is to write a list of language data to a file or JSON string. Unless we want PrimaryRegion/Country data in the file or string, I don't see why this is needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-:-Ki0MLqMZAPu0f0O44g-:bthu91k)*

hatton:
A guess at why they probably want this, John... LanguageForge, as a web app, probably doesn't make use of libpalaso, but may well consume the json index. So by pushing this function over into the index creator, they can get at the country from the web apps.

---

Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-:-Ki0a1SgMnpX6EXnMem5:b-8i7hql)*

JohnThomson:
That feels to me like an additional change (which I would be fine with) rather than an alternative one, since as far as I can see LanguageLookup doesn't consume NewLanguageIndex in any way. (Unless the idea is to deprecate LanguageLookup? I gather from one comment that NewLanguageIndex is supposed to be renamed and replace something eventually? But there isn't a LanguageIndex class...)

---

Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-:-Ki0w9d66KhRY9L4H7Ad:bejmg25)*

glasseyes:
NewLanguageIndex creates the LanguageDataIndex.txt file that LanguageLookup consumes (which is why I've renamed it to LanguageDataIndex, the NewLanguageIndex name is historical from the development so irrelevant now).
LanguageLookup has already been changed to use that data.

Thanks JohnH, that is indeed the reason I understand from Chris.

I've done the changes to integrate it with the LanguageDataIndex. I'll make a separate PR for it then it hopefully shouldn't be much to make the separate PRs for the other parts.

---

Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 21 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz4Yks_Zd_SdQW8FU7:-Ki4cGQlfEOfK3lJJ8OG:bizrkuw) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L21)):*

We could do this immediate function without it. It seemed a natural implementation to me since the file contains  region codes and the code base contains region objects to keep track of the complete region object that is specified by the code.
</blockquote></details>

Other things consuming it may not have the region to country name conversion so I've gone ahead and removed it.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 50 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz8_Hm5iaJYxhQxg8e:-Ki4cQW4o5_wurfmyGDv:bkvf5vt) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L50)):*

As the comment says, ethnologue.com describes Maku (xak) as  "a language of Venezuela" so I went with that.
</blockquote></details>

Sorry, I badly miscommunicated there. It was intended to be a gentle pointer to your typo.

---

*[SIL.WritingSystems/LanguageInfo.cs, line 56 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz8VbKaTy8OqnRU9sC:-Ki4cbAAtkmX5AV_1HG1:b-7bavi4) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageInfo.cs#L56)):*

Why? These languages, like many others, don't have a region specified (that's why I later added PrimaryCountry as an additional new property) so there's nothing to store in the original data.
</blockquote></details>

Because that is where the data is created

---

*[SIL.WritingSystems/LanguageLookup.cs, line 80 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-Khz75iRyWoIFFa246Wx:-Ki4cpBOEn6CT2CwJXmM:b-suj158) ([raw file](https://github.com/sillsdev/libpalaso/blob/1b0b4ac785544e4787b4d584f4c8ac3a780dc395/SIL.WritingSystems/LanguageLookup.cs#L80)):*

I don't know why it should be (partly because whoever wrote NewLanguageIndex completely failed to write a class comment to describe what the class is about and what responsibilities it has). It seems its only purpose is to write a list of language data to a file or JSON string. Unless we want PrimaryRegion/Country data in the file or string, I don't see why this is needed.
</blockquote></details>

Sorry, you are right, I missed writing that class comment. The purpose of NewLanguageIndex (now called LanguageDataIndex) is to create the LanguageDataIndex.txt/json for LanguageLookup to consume.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/474#-:-Ki4dGy16-j51KRybiPY:b-45n8kb)*

JohnThomson:
Abandoned in favor of https://github.com/sillsdev/libpalaso/pull/478

PR #480, opened 2017-04-19
Added SafeInvoke method to ControlExtensions

tombogle:
This method was based on implementations in LogBox, Glyssen, and BloomDesktop, plus insights from MSDN and SO. Note that it represents a definite change to the implementation in LogBox (which I removed). TimS and I studied it and feel like it is a slight improvement in that it is actually checking IsDisposed on the correct control, rather than the LogBox itself. But it does get away from using the SynchronizationContext, which Microsoft argues is actually the best way to handle synchronization. I left in place the apparently redundant try-catch blocks in the calling code, somewhat against my better judgment (but scared off by the confusing comments there). I did test it in HearThis, but not in WeSay (or any other clients).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/480)

### DIFF
--- a/SIL.Windows.Forms/Extensions/ControlExtensions.cs
+++ b/SIL.Windows.Forms/Extensions/ControlExtensions.cs
@@ -1,7 +1,10 @@
 ﻿using System;
 using System.ComponentModel;
+using System.Diagnostics;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Windows.Forms;
+using SIL.Code;
 
 namespace SIL.Windows.Forms.Extensions
 {
@@ -81,6 +84,83 @@ namespace SIL.Windows.Forms.Extensions
 		public static void BeginInvoke(this Control control, Action action)
 		{
 			control.BeginInvoke(action);
+		}
+
+		public enum ErrorHandlingAction
+		{
+			Throw,
+			IgnoreIfDisposed,
+			IgnoreAll,
+		}
+
+		/// <summary>
+		/// Invoke an action safely even if called from the background thread.
+		/// </summary>
+		/// <remarks>
+		/// Invoking on the ui thread from background threads works *most* of the time, with occasional crash.
+		/// Stackoverflow has a good collection of people trying to deal with these corner cases, where
+		/// InvokeRequired(), for example, is unreliable (it doesn't tell you if the control hasn't even
+		/// got a handle yet).
+		/// The exact behavior of some of these methods have changed between different versions of .net.
+		/// Here's the relevant page in msdn that explains the current situation:
+		/// https://msdn.microsoft.com/en-us/library/system.windows.forms.control.invokerequired(v=vs.110).aspx
+		/// So now I'm trying something more mainstream here, from a highly voted SO answer.
+		/// </remarks>
+		public static void SafeInvoke(this Control control, Action action, string nameForErrorReporting = "context not supplied",
+			ErrorHandlingAction errorHandling = ErrorHandlingAction.Throw, bool forceSynchronous = false)
+		{
+			Guard.AgainstNull(control, nameForErrorReporting); // throw this one regardless of the throwIfAnythingGoesWrong
+			Guard.AgainstNull(action, nameForErrorReporting); // throw this one regardless of the throwIfAnythingGoesWrong
+
+			if (control.IsDisposed)
+			{
+				if (errorHandling == ErrorHandlingAction.Throw)
+					throw new ObjectDisposedException("Control is already disposed. (" + nameForErrorReporting + ")");
+				return; // Caller asked to ignore this.
+			}
+
+			if (!control.InvokeRequired)
+			{
+				// InvokeRequired will return false if the control isn't set up yet
+				if (!control.IsHandleCreated)
+				{
+					// This situation happened in BL-2918, prompting the introduction of this SafeInvoke method
+
+					if (errorHandling == ErrorHandlingAction.IgnoreAll)
+						return;
+					throw new ApplicationException("SafeInvoke.Invoke apparently called before control created (" + nameForErrorReporting + ")");
+
+					// Resist the temptation to work around this by just making the handle be created with something like
+					// var unused = control.Handle
+					// This can create the handle on the wrong thread and make the application unstable. (I believe it would crash instantly on Linux.)
+				}
+			}
+
+			// This implementation began with http://stackoverflow.com/a/809186/723299, but allows the caller to specify the
+			// desired handling of various types of errors and also deals specially with a control that is an IProgress and
+			// therefore has a SyncContext (which is better, according to MSDN).
+			try
+			{
+				if (control.InvokeRequired)
+				{
+					var delgate = (Action)delegate { SafeInvoke(control, action, nameForErrorReporting, errorHandling, forceSynchronous); };
+					if (forceSynchronous)
+						control.Invoke(delgate);
+					else
+						control.BeginInvoke(delgate);
+				}
+				else
+					action();
+			}
+			catch (Exception error)
+			{
+				SIL.Reporting.Logger.WriteEvent("**** " + error.Message);
+
+				if (errorHandling != ErrorHandlingAction.IgnoreAll)
+					throw new TargetInvocationException(nameForErrorReporting + ":" + error.Message, error);
+
+				Debug.Fail("This error would be swallowed in release version: " + error.Message);
+			}
 		}
 
 		/// ------------------------------------------------------------------------------------
--- a/SIL.Windows.Forms/Progress/LogBox.cs
+++ b/SIL.Windows.Forms/Progress/LogBox.cs
@@ -7,6 +7,7 @@ using System.Windows.Forms.VisualStyles;
 using SIL.Extensions;
 using SIL.Progress;
 using SIL.Reporting;
+using SIL.Windows.Forms.Extensions;
 
 namespace SIL.Windows.Forms.Progress
 {
@@ -223,45 +224,9 @@ namespace SIL.Windows.Forms.Progress
 		/// This is an attempt to avoid a mysterious crash (B. Waters) where the invoke was
 		/// happening before the window's handle had been created
 		/// </summary>
-		public void SafeInvoke(Control box, Action action)
+		private void SafeInvoke(Control box, Action action)
 		{
-			try
-			{
-				if (!box.IsHandleCreated)
-				{
-					//Debug.Fail("In release build, would have given up writing this message, because the destination control isn't built yet.");
-					return;
-				}
-
-				if(!IsDisposed) //note, we'll check it again on the UI thread, since it could be disposed while waiting to run
-				{
-					if (SynchronizationContext.Current == SyncContext) //we're on the UI thread
-					{
-						action();
-					}
-					else
-					{
-						//NB: if you're getting eratic behavior here, make sure there isn't one of those "access to modified closure" situations in your calling method
-						SyncContext.Post(state =>
-							{
-								//assumption is that since this is run on the UI thread, we can't be disposed in between the check of IsDiposed and the action itself
-								if(!IsDisposed)
-									action();
-							}, null);
-					}
-				}
-			}
-			catch (Exception)
-			{
-#if DEBUG
-				throw;
-#else
-				//WS-34006
-				// better to swallow the message than raise a stink
-#endif
-			}
-
-
+			box.SafeInvoke(action, errorHandling:ControlExtensions.ErrorHandlingAction.IgnoreAll);
 		}
 
 		private void Write(Color color, FontStyle style, string msg, params object[] args)


hatton:
If I understand correctly, the effect on Bloom is through Bloom's use of logbox? We can bear that risk. Elsewhere in Bloom, I'd be wary of switching to this immediately, because this is basically impossible to test.  The only way to have confidence is through deployment and use by at least hundreds of people.  However after a few years of someone else using it in an app that gets lots of use, then it will be great to use :-) If instead, the hope was that everyone would use it, then I would have favored the Bloom implementation, which is proven with thousands of users over many years.

tombogle:
Indeed, my hope is/was to have Bloom switch at some point in the relatively near future. Otherwise, the benefit of having a single well-conceived implementation is somewhat thwarted. I am sympathetic to your concerns. But the Bloom implementation was a little shy of what Glyssen wanted and seemed to be slightly less intuitive in the way it throws, catches and re-throws exceptions. The implementation we settled on also attempts to do a better job of reporting the source of the problem when throwing unexpected exceptions that happen in the actual code being invoked. I did consider trying to write unit tests for it. Maybe I should. But that still may not allay your fears.

JohnThomson:
A few thoughts and questions...even though I know this has been merged, JohnH asked me to take a look.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 109 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGInRnf2xWEZXHVPis:-KiGInRnf2xWEZXHVPit:bc6t42y) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L109)):*
> ```Smalltalk
> 		/// So now I'm trying something more mainstream here, from a highly voted SO answer.
> 		/// </remarks>
> 		public static void SafeInvoke(this Control control, Action action, string nameForErrorReporting = "context not supplied",
> ```

Seems to me highly likely that if we need this function, we will also need one where a result is needed from the thing that can be safely invoked. Of course we can only get that if not using BeginInvoke (i.e., forceSynchronous true). But you're not adding a new function, just improving an existing one, so I suppose it's fair to leave that YAGNI, even in a library function.

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 131 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGFyuqcRPv1LkTFOdH:-KiGFyurgJ3HJ2f-UM7q:b-d56etd) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L131)):*
> ```Smalltalk
> 					if (errorHandling == ErrorHandlingAction.IgnoreAll)
> 						return;
> 					throw new ApplicationException("SafeInvoke.Invoke apparently called before control created (" + nameForErrorReporting + ")");
> ```

Would be more precise to say, "...before control Handle created..."

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 141 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGI7AJ-ghEAczTXyuR:-KiGI7AJ-ghEAczTXyuS:b-khrcmp) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L141)):*
> ```Smalltalk
> 			// This implementation began with http://stackoverflow.com/a/809186/723299, but allows the caller to specify the
> 			// desired handling of various types of errors and also deals specially with a control that is an IProgress and
> 			// therefore has a SyncContext (which is better, according to MSDN).
> ```

I don't see any special handling of IProgress or any use of SyncContext. If it's better to use the SyncContext approach, why was that changed?

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGGOF-FFrNrC95uQlP:-KiGGOF-FFrNrC95uQlQ:b-shwk86) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L148)):*
> ```Smalltalk
> 					var delgate = (Action)delegate { SafeInvoke(control, action, nameForErrorReporting, errorHandling, forceSynchronous); };
> 					if (forceSynchronous)
> 						control.Invoke(delgate);
> ```

Why do we want to do this rather than just control.Invoke(action)? Another Invoke won't be required on the thread that Invoke selects, and we wouldn't have gotten here unless all the other checks passed,  and all this does on the proper thread is run action()...is it just so any exceptions in action() can be caught in the invoke thread?
It feels to me as if SafeInvoke is about safely doing the Invoke process if necessary...it shouldn't be about suppressing (or converting to inner exceptions) things that go wrong in the action itself.  But maybe there's good reason for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-:-KiGJwFHvNf1ubrJj2h0:bupf752)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 109 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGInRnf2xWEZXHVPis:-KiGaAwkBsxfqsgdKOS0:bzaro26) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L109)):*

Seems to me highly likely that if we need this function, we will also need one where a result is needed from the thing that can be safely invoked. Of course we can only get that if not using BeginInvoke (i.e., forceSynchronous true). But you're not adding a new function, just improving an existing one, so I suppose it's fair to leave that YAGNI, even in a library function.
</blockquote></details>

True. In practice I think forceSynchronous == true is not very commonly used (if ever?). The original implementation (as TimS pointed out) was not actually implemented consistently. It was always invoking synchronously. In practice, I think that was maybe okay, since in that case the UI thread is already tied up and therefore won't be responsive.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 131 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGFyuqcRPv1LkTFOdH:-KiGbhpOoTfzYeOGfe49:b-chf9gv) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L131)):*

Would be more precise to say, "...before control Handle created..."
</blockquote></details>

True, depending on what youi mean by "control" and "created." That wording came over from the Bloom implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 141 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGI7AJ-ghEAczTXyuR:-KiGc1pACgNzheXDz3KN:b26cvqq) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L141)):*

I don't see any special handling of IProgress or any use of SyncContext. If it's better to use the SyncContext approach, why was that changed?
</blockquote></details>

Oops. I wrote that comment when the implementation WAS checking to see if the control was an IProgress and, if so, using its SyncContext. That was to try to accommodate the way the SynchContext was being used in the LogBox implementation. But after we studied it thoroughly we realized that the implementation in that class was using the passed in control (assumed to be a child control of the logbox) and *this* in inconsistent and confusing ways. (For example, it was used the passed in control to see if the handle had been created but checking to see whether *this* -- the logbox -- had been disposed.) In any case, there didn't seem to be any possibility that the two controls could have different synchronization contexts, so we decided it wasn't worth the added complexity to search up the controls ownership stack to to see if there was an IProgress lurking in there, just for the sake of using the SynchronizationContext. I'm not sure why MS thinks it's better. My guess is that there is some theoretical (non-Windows, non-Linux) world where the synch rules could be more complex.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGGOF-FFrNrC95uQlP:-KiGeVmpIaC3Sxlo2ew5:b-2cj4ru) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L148)):*

Why do we want to do this rather than just control.Invoke(action)? Another Invoke won't be required on the thread that Invoke selects, and we wouldn't have gotten here unless all the other checks passed,  and all this does on the proper thread is run action()...is it just so any exceptions in action() can be caught in the invoke thread?
It feels to me as if SafeInvoke is about safely doing the Invoke process if necessary...it shouldn't be about suppressing (or converting to inner exceptions) things that go wrong in the action itself.  But maybe there's good reason for that.
</blockquote></details>

We almost changed that, but in the interests of trying not to change the actual behavior of the Bloom version, we left it. As you point out, the only possible advantage would be to allow the catching exceptions thrown by the invoked code. It seemed like a dubious benefit. We tried to (very slightly) improve it by converting it to an inner exception with the outer one providing the context information passed in. If Bloom isn't relying on this implementation, I'd be very happy to just invoke the action.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

JohnThomson:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 141 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGI7AJ-ghEAczTXyuR:-KiHLiq5xHrrhr2FRHb2:bkskl9r) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L141)):*

Oops. I wrote that comment when the implementation WAS checking to see if the control was an IProgress and, if so, using its SyncContext. That was to try to accommodate the way the SynchContext was being used in the LogBox implementation. But after we studied it thoroughly we realized that the implementation in that class was using the passed in control (assumed to be a child control of the logbox) and *this* in inconsistent and confusing ways. (For example, it was used the passed in control to see if the handle had been created but checking to see whether *this* -- the logbox -- had been disposed.) In any case, there didn't seem to be any possibility that the two controls could have different synchronization contexts, so we decided it wasn't worth the added complexity to search up the controls ownership stack to to see if there was an IProgress lurking in there, just for the sake of using the SynchronizationContext. I'm not sure why MS thinks it's better. My guess is that there is some theoretical (non-Windows, non-Linux) world where the synch rules could be more complex.
</blockquote></details>

The big advantage might be that you can get the proper sync context for all windows.Forms controls from WindowsFormsSynchronizationContext.Current, and then posting the action to that thread doesn't require considering whether some control has had its handle created. But that seems to be roughly what the old Palaso code was doing and amounts  to "Invoking on the ui thread from background threads" which the comment implies the Bloom method was made to AVOID doing because it only "works *most* of the time, with occasional crash." So if the commenter was right and I'm understanding him, we do it this way because the SyncContext approach does NOT work better, or even as well, for our situation. So probably the code you have should stand, but it would be nice to fix the comment.

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGGOF-FFrNrC95uQlP:-KiHKN5aQdNreEjDWIUs:blks29d) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L148)):*

We almost changed that, but in the interests of trying not to change the actual behavior of the Bloom version, we left it. As you point out, the only possible advantage would be to allow the catching exceptions thrown by the invoked code. It seemed like a dubious benefit. We tried to (very slightly) improve it by converting it to an inner exception with the outer one providing the context information passed in. If Bloom isn't relying on this implementation, I'd be very happy to just invoke the action.
</blockquote></details>

Bloom uses its version of this function for exactly one thing, which is to update its book collection control when a watcher notices new or modified files in the disk collection. That's a big complicated bit of code and I hate the idea of suppressing all exceptions that might happen in it, but if we really need to, it's only one thing to change when we switch from Bloom's version to the new Palaso version. I don't think that's a good enough reason to make this new shared method encourage such (usually) bad behavior.
To be consistent the non-invoked call to action() should somehow be outside the try/catch, too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 131 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGFyuqcRPv1LkTFOdH:-KiW-Vrq9kb3_yH65xTE:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L131)):*

True, depending on what youi mean by "control" and "created." That wording came over from the Bloom implementation.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 141 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGI7AJ-ghEAczTXyuR:-KiW-qodR2j23e4gbkvW:b-jod0x8) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L141)):*

The big advantage might be that you can get the proper sync context for all windows.Forms controls from WindowsFormsSynchronizationContext.Current, and then posting the action to that thread doesn't require considering whether some control has had its handle created. But that seems to be roughly what the old Palaso code was doing and amounts  to "Invoking on the ui thread from background threads" which the comment implies the Bloom method was made to AVOID doing because it only "works *most* of the time, with occasional crash." So if the commenter was right and I'm understanding him, we do it this way because the SyncContext approach does NOT work better, or even as well, for our situation. So probably the code you have should stand, but it would be nice to fix the comment.
</blockquote></details>

I interpreted the "works *most* of the time, with occasional crash" to mean that *blindly* invoking on the the UI thread without checking to ensure that the handle was created and that the control has not subsequently been disposed works most of the time. In any cash, getting the current (global) synch context and invoking on it will guarantee that you're on the correct thread, but if you try to interact with any of the methods or properties of the control that require it to be "live" (i.e., have a handle and not be disposed), you'll still end up getting into trouble. At least a fair percentage of the time (and perhaps even *most* of the times when a simple unsafe invoke fails), it is probably a situation where the caller wants to interact with the control. Maybe the best way would be to use the current synch context and then if necessary, once we get on the UI thread, we could safely create the handle if it had not yet been created. But my guess is that in most situations where this would be needed, it would probably be masking a design deficiency.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGGOF-FFrNrC95uQlP:-KiW26X6SiT6cXNLEnUy:bgh4ib) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L148)):*

Bloom uses its version of this function for exactly one thing, which is to update its book collection control when a watcher notices new or modified files in the disk collection. That's a big complicated bit of code and I hate the idea of suppressing all exceptions that might happen in it, but if we really need to, it's only one thing to change when we switch from Bloom's version to the new Palaso version. I don't think that's a good enough reason to make this new shared method encourage such (usually) bad behavior.
To be consistent the non-invoked call to action() should somehow be outside the try/catch, too.
</blockquote></details>

I agree. A far better way (at least in terms of code readability) would be to require callers to include a top-level try-catch block inside their action if they needed that kind of error suppression. If we really felt it was necessary, I guess we could later write a SagfeInvokeWithInternalErrorSuppression method that would wrap the passed-in action with a try-catch block. I'll make your suggested improvement.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/480#-KiGGOF-FFrNrC95uQlP:-KiW2xbRwfz3hLw-NbKi:b3ifkg8) ([raw file](https://github.com/sillsdev/libpalaso/blob/13ef298569fa42d866890bce5300248fb819587f/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L148)):*

I agree. A far better way (at least in terms of code readability) would be to require callers to include a top-level try-catch block inside their action if they needed that kind of error suppression. If we really felt it was necessary, I guess we could later write a SagfeInvokeWithInternalErrorSuppression method that would wrap the passed-in action with a try-catch block. I'll make your suggested improvement.
</blockquote></details>

Edit: SafeInvokeWithInternalErrorSuppression

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/480)*

PR #482, opened 2017-04-21
Allow image chooser to draw from additional collections

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/482)


SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs
```diff
@@ -91,7 +162,7 @@ public int LoadMultilingualIndex(string pathToIndexFile)
 					Debug.Assert(parts.Length > defaultColumn);
 					if (parts.Length <= defaultColumn)
 						continue;
-					var partialPath = String.Format("{0}:AOR_{1}.png", parts[3], parts[1]);
+					var partialPath = GetPartialPath(rootImagePath, parts[3], parts[1], ref filenamePrefix, pathPrefix);
```
jasonleenaylor: Maybe it is too late on Friday, but it seems that partialPath is assigned to but never used.

SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs
```diff
@@ -160,21 +250,33 @@ private string GetCaption(string path)
 
 		public IEnumerable<string> GetPathsFromResults(IEnumerable<object> results, bool limitToThoseActuallyAvailable)
 		{
-			foreach (var macPath in results)
+			foreach (string macPath1 in results)
```
jasonleenaylor: macPath1 isn't helping me

SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs
```diff
@@ -160,21 +250,33 @@ private string GetCaption(string path)
 
 		public IEnumerable<string> GetPathsFromResults(IEnumerable<object> results, bool limitToThoseActuallyAvailable)
```
jasonleenaylor: Perhaps a method comment here would be a nice addition, while you're at it.

JohnThomson:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 165 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHH3VECXv_Hi4ayRZ7-r1-165:-KiHOcCMHOvdN-AC2HDr:b-1nx0om) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L165)):*

Maybe it is too late on Friday, but it seems that partialPath is assigned to but never used.
</blockquote></details>

Look about 10 and 15 lines down.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHH3VECXv_Hi4ayRZ7-r1-251:-KiHS5hHhIzLP_kC2OK6:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L251)):*

Perhaps a method comment here would be a nice addition, while you're at it.
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 253 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHH3VECXv_Hi4ayRZ7-r1-253:-KiHOlf6zi9vtwT8c2FD:b-yof539) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L253)):*

macPath1 isn't helping me
</blockquote></details>

macPath is original...I think because it has colons in it instead of platform-specific separators. I added macPath1 because I needed to modify macPath and don't like code that modifies the loop control variable. Can you think of anything better? It's basically a sort of internal compact path which LoadMultilingualIndex knows how to create and this method knows how to interpret. I'll try adding the method comment and see if it helps.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

hatton:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 17 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSDH9DA9iej-kRIZP:-KiHSDH9DA9iej-kRIZQ:b-petvwh) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L17)):*
> ```Smalltalk
> 
> 	/// <summary>
> 	/// Originally a wrapper for the Art of Reading image collection, it may now include others.
> ```

Name wants to change the change of role. AOR is just one of the collections now.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 22 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSXbhR96_N_6lho8Z:-KiHSXbhR96_N_6lho8_:b1nv4u6) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L22)):*
> ```Smalltalk
> 	/// The root folder contains the master index file. It should be the only file with a name ending in
> 	/// MultilingualIndex.txt. Typically the previous part of the name corresponds to the collection name
> 	/// without spaces, e.g., ArtOfReadingMultilingualIndex.txt.
> ```

I think this is now awkward. It probably started because we were (and probably still do?) shipping the aor index with our software that was more current than the index on people's machines.

I know you are working under the constraint of still working with already-installed AOR collections... but if feasible it would make more sense to just have "index.txt", rather than repeat the name of the index inside of a folder that probably has the same name.

Perhaps it would come out clean to have the main code know *nothing* of AOR, and just have standards and names that make sense on their own, as if we were starting from scratch with a clean architecture. Then, have a subclass that is only instantiated in the case of AOR 3.0 that handles all the exceptions. The next release of AOR, we can make it conformant to standard and eventually drop the special-case class.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHUL4K7KDeoLOHSiTC:-KiHUL4K7KDeoLOHSiTD:b68sx08) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L251)):*
> ```Smalltalk
> 
> 
> 		public IEnumerable<string> GetPathsFromResults(IEnumerable<object> results, bool limitToThoseActuallyAvailable)
> ```

results really has to be <object>?

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 253 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHH3VECXv_Hi4ayRZ7-r1-253:-KiHUQljYgnFG_9x7-Zk:b-1c8pqj) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L253)):*

macPath is original...I think because it has colons in it instead of platform-specific separators. I added macPath1 because I needed to modify macPath and don't like code that modifies the loop control variable. Can you think of anything better? It's basically a sort of internal compact path which LoadMultilingualIndex knows how to create and this method knows how to interpret. I'll try adding the method comment and see if it helps.
</blockquote></details>

How about "rawMacOSPath".

This whole macOS path thing... I can see how the original index must have had them... but the current one  (shipping for many years now) doesn't. Can we just retire this?

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 396 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHVSQJfuSLWzJkeUjX:-KiHVSQK4QRE0UoJLaMa:b-wap1pz) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L396)):*
> ```Smalltalk
> 				// By analogy with Linux case in TryToGetRootImageCatalogPath.
> 				// I'm not sure Linux actually needs a special case here...would the Windows code
> 				// give a reasonable result?
> ```

Sure hope so... seems to be going against the spirit of the system to hard code it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

JohnThomson:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 17 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSDH9DA9iej-kRIZP:-KiKpSp8pqMfju0iMLPX:b9e48gy) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L17)):*

Name wants to change the change of role. AOR is just one of the collections now.
</blockquote></details>

Yes, I thought about that. But then we're making much more of a breaking change to libpalaso...changing a class name of a public API. I suppose it's not actually very likely that anything but the image gallery and toolbox use it. Also (currently at least) it does still treat AOR specially...always includes it, knows various places to look and about older versions.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 22 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSXbhR96_N_6lho8Z:-KiKq6t9qCkQH1B68XXE:b-2pbmft) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L22)):*

I think this is now awkward. It probably started because we were (and probably still do?) shipping the aor index with our software that was more current than the index on people's machines.

I know you are working under the constraint of still working with already-installed AOR collections... but if feasible it would make more sense to just have "index.txt", rather than repeat the name of the index inside of a folder that probably has the same name.

Perhaps it would come out clean to have the main code know *nothing* of AOR, and just have standards and names that make sense on their own, as if we were starting from scratch with a clean architecture. Then, have a subclass that is only instantiated in the case of AOR 3.0 that handles all the exceptions. The next release of AOR, we can make it conformant to standard and eventually drop the special-case class.
</blockquote></details>

There's no requirement to include the collection name. Plain "MultilingualIndex.txt" will work fine. It would be trivial to shorten it so it will accept any file ending in Index.txt.

I actually started out attempting to refactor into a base class that knew nothing of AOR and an AOR-specific one, and it was a mess. Thing is, the client code isn't really looking for an object that represents one folder of images, it's looking for one that represents all of them. So we'd really need three classes, ImageCollection, ArtOfReadingImageCollection, and CollectionOfImageCollections. And clients would need to switch to using the CollectionOfImageCollections to get the new functionality. No doubt something could be done, but it would involve either breaking changes or a good bit of redundant code, and considerably more work. I can have another go if you think it's worth it.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHUL4K7KDeoLOHSiTC:-KiKqY_5Aj_-0HXUnBHF:bh7hxwb) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L251)):*

results really has to be <object>?
</blockquote></details>

I added a long comment to this method with my speculations on the reasons. It needs to match the return type of GetMatchingPictures. As the remarks indicate, I don't know why both aren't IEnumerable<string>. Again, a question is whether we want to break a shipping API.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 253 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHH3VECXv_Hi4ayRZ7-r1-253:-KiKr7m8GHeomeFGEeXD:bpwfjbr) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L253)):*

How about "rawMacOSPath".

This whole macOS path thing... I can see how the original index must have had them... but the current one  (shipping for many years now) doesn't. Can we just retire this?
</blockquote></details>

I'm thinking now of changing to 'internalPath' or maybe 'compactPath'. As the new comment explains, it's not really any sort of OS path (not even relative...it omits the 'Images' bit). I think the main point is to save space by storing the minimum required to find the file in our internal dictionaries.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 396 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHVSQJfuSLWzJkeUjX:-KiKrg_mr8TljyALzw7d:baivkks) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L396)):*

Sure hope so... seems to be going against the spirit of the system to hard code it.
</blockquote></details>

I should check with Steve.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

hatton:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 17 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSDH9DA9iej-kRIZP:-KiLBVTXBOZzW_QsC1HY:b-ljpoop) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L17)):*

Yes, I thought about that. But then we're making much more of a breaking change to libpalaso...changing a class name of a public API. I suppose it's not actually very likely that anything but the image gallery and toolbox use it. Also (currently at least) it does still treat AOR specially...always includes it, knows various places to look and about older versions.
</blockquote></details>

I think breaking changes are fine. Class names that no longer reflect what they are are not fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

JohnThomson:
Review status: 0 of 11 files reviewed at latest revision, 7 unresolved discussions.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 17 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSDH9DA9iej-kRIZP:-KiVquH2NIPgH0W-cXrE:b291rfk) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L17)):*

I think breaking changes are fine. Class names that no longer reflect what they are are not fine.
</blockquote></details>

Changed name to simply ImageCollection

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 22 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHSXbhR96_N_6lho8Z:-KiVk1pTr_vD_G0_mGBi:bivvrvk) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L22)):*

There's no requirement to include the collection name. Plain "MultilingualIndex.txt" will work fine. It would be trivial to shorten it so it will accept any file ending in Index.txt.

I actually started out attempting to refactor into a base class that knew nothing of AOR and an AOR-specific one, and it was a mess. Thing is, the client code isn't really looking for an object that represents one folder of images, it's looking for one that represents all of them. So we'd really need three classes, ImageCollection, ArtOfReadingImageCollection, and CollectionOfImageCollections. And clients would need to switch to using the CollectionOfImageCollections to get the new functionality. No doubt something could be done, but it would involve either breaking changes or a good bit of redundant code, and considerably more work. I can have another go if you think it's worth it.
</blockquote></details>

Changed to accept simply "Index.txt" (or any file name ending in that)

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHUL4K7KDeoLOHSiTC:-KiVmUTD8QIxhZLRNkEl:bmciqml) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L251)):*

I added a long comment to this method with my speculations on the reasons. It needs to match the return type of GetMatchingPictures. As the remarks indicate, I don't know why both aren't IEnumerable<string>. Again, a question is whether we want to break a shipping API.
</blockquote></details>

OK, I changed the two methods that return and take this collection to IEnumerable<string>

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 253 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHH3VECXv_Hi4ayRZ7-r1-253:-KiVmFLqS8WW7S5uZRgs:bjwx3nn) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L253)):*

I'm thinking now of changing to 'internalPath' or maybe 'compactPath'. As the new comment explains, it's not really any sort of OS path (not even relative...it omits the 'Images' bit). I think the main point is to save space by storing the minimum required to find the file in our internal dictionaries.
</blockquote></details>

Renamed to internalPath/rawInternalPath

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 396 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHVSQJfuSLWzJkeUjX:-KiVpyIsMdF3CqZOQF9y:b-fz6wqa) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L396)):*

I should check with Steve.
</blockquote></details>

I took out the special Linux case, which would typically have given the same result as what I had, anyway. The result is not exactly a standard way of doing things on Linux, but I think we can reasonably wait and see if anyone uses this and how they actually want to distribute their image collection before making special cases for Linux.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

jasonleenaylor:
Found one odd niggle, probably not worth addressing unless another change is required
:lgtm:

---

Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Windows.Forms/ImageGallery/IImageCollection.cs, line 13 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiWI4WwuUfsnVeOvZBs:-KiWI4WwuUfsnVeOvZBt:bs8uria) ([raw file](https://github.com/sillsdev/libpalaso/blob/adba33818ddc32593f459b80e8ae05fbf899ae52/SIL.Windows.Forms/ImageGallery/IImageCollection.cs#L13)):*
> ```Smalltalk
> 
> 		/// <summary>
> 		/// The imageTOken here could be a path or whatever, the client doesn't need to know or care
> ```

Not your fault, but strange case here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-:-KiWJ59b5lBbrM9Efrtn:b-y3v9wu)*

JohnThomson:
Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

*[SIL.Windows.Forms/ImageGallery/IImageCollection.cs, line 13 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiWI4WwuUfsnVeOvZBs:-KiWMiR_iOVQ58n1fvx1:b-7lwkp7) ([raw file](https://github.com/sillsdev/libpalaso/blob/adba33818ddc32593f459b80e8ae05fbf899ae52/SIL.Windows.Forms/ImageGallery/IImageCollection.cs#L13)):*

Not your fault, but strange case here
</blockquote></details>

Interesting. I hadn't seen this message. This might be the one place where the author explains the mysterious (previous) use of 'object' in the return type of GetMatchingPictures and input to GetPathsFromResults. Quite possibly it's mean to be one of the objects that GetMatchingPicures (used to) return. For consistency, if so, it should be changed to string too.
It's not surprising that there are no usages of this method since the one implementation of the interface throws a NotImplementedException. So there's no real clue I can find what this 'token' is supposed to be.
Leaning towards a new commit that removes the method entirely from both interface and implementation...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

hatton:
I haven't been able to get this to work yet, so I'll leave off reviewing for now in anticipation of code changes.

---

Review status: 0 of 11 files reviewed at latest revision, 7 unresolved discussions.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHUL4K7KDeoLOHSiTC:-KiVt76d10hwHJCb3sA2:b-6224f) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L251)):*
> Again, a question is whether we want to break a shipping API.

Sure no problem. Especially as we are now about to make a libpalaso 4. Perhaps we need a branch for that?

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 67 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Ki_jvor8Lwyy30dqo4h:-Ki_jvor8Lwyy30dqo4i:b10edpz) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L67)):*
> ```Smalltalk
> 		/// the parent of this directory, unless both are passed explicitly (a case not tested
> 		/// much if at all).
> 		/// Would be nice to rename to something like DefaultAorRootImagePath, but I don't want
> ```

Since we have agreed to change the API, shall we do this now?

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 482 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Ki_lHP_0wtmctrlAyI2:-Ki_lHP_0wtmctrlAyI3:b-8itxka) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L482)):*
> ```Smalltalk
> 
> 			c.RootImagePath = path;
> 			c.GetIndexLanguages();
> ```

Maybe just part of my confusion about the architecture here, but I read this be getting languages out of the single root collection, not all the collections.

---

*[SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs, line 490 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Ki_jVbrLwN8sYCdCpPU:-Ki_jVbrLwN8sYCdCpPV:b-2g1x79) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ArtOfReadingImageCollection.cs#L490)):*
> ```Smalltalk
> 			// be done before the user finishes typing a search string.
> 			var thr = new Thread(c.LoadImageIndex);
> 			thr.Name = "LoadArtOfReadingIndex";
> ```

Should this be renamed something generic?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-:-Ki_lZ372ja6DFD5ZjZb:b-6k377k)*

JohnThomson:
Hopefully working now, and I think I've addressed all the comments to date.

---

Review status: 0 of 16 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 67 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Ki_jvor8Lwyy30dqo4h:-KifcLPfXz7PhjxRIcgU:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L67)):*

Since we have agreed to change the API, shall we do this now?
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiHUL4K7KDeoLOHSiTC:-KifcQQJ6vAU-kwkSjxC:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L251)):*

> Again, a question is whether we want to break a shipping API.

Sure no problem. Especially as we are now about to make a libpalaso 4. Perhaps we need a branch for that?
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 482 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Ki_lHP_0wtmctrlAyI2:-KifgEskVr-dTY43VS3p:b-k2rwtz) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L482)):*

Maybe just part of my confusion about the architecture here, but I read this be getting languages out of the single root collection, not all the collections.
</blockquote></details>

Good catch. Fixed.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 490 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Ki_jVbrLwN8sYCdCpPU:-KifgPlvZ213K4Uq7Swz:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/8ae7ddca1e7df296f3d02af822ec2d9b867570f3/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L490)):*

Should this be renamed something generic?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-:-KifgSyoMDjcUAGCYC1b:b-oks4wl)*

hatton:
Review status: 0 of 16 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 502 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kig-ReJKXxvyNwNkHb7:-Kig-ReJKXxvyNwNkHb8:b6x0l0s) ([raw file](https://github.com/sillsdev/libpalaso/blob/e8f20a3844da914db781d3007cc68e3fce3f1846/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L502)):*
> ```Smalltalk
> 			else
> 			{
> 				//look for the folder created by the ArtOfReadingFree installer
> ```

At the end of this sentence, please add " 3.0. Later versions will put in the ImageCollections folder."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

hatton:
Review status: 0 of 16 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 502 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kig-ReJKXxvyNwNkHb7:-Kig823SsIGrAgWeQdJJ:bcj88xq) ([raw file](https://github.com/sillsdev/libpalaso/blob/e8f20a3844da914db781d3007cc68e3fce3f1846/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L502)):*

At the end of this sentence, please add " 3.0. Later versions will put in the ImageCollections folder."
</blockquote></details>

Correction: the above should be " before 3.3, the first version to put it in the ImageCollections folder"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

jasonleenaylor:
Reviewed 1 of 12 files at r2, 2 of 6 files at r4.
Review status: 3 of 17 files reviewed at latest revision, 10 unresolved discussions.

---

*[DistFiles/Palaso.en.tmx, line 246 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiksnQtb2nq8Ze9gYXU:-KiksnQtb2nq8Ze9gYXV:bsnkch7) ([raw file](https://github.com/sillsdev/libpalaso/blob/cdc0bf5447b25999fde1409904de2dfb30f27bf3/DistFiles/Palaso.en.tmx#L246)):*
> ```
>       <tu tuid="ImageToolbox.SearchArtOfReading">
>          <tuv xml:lang="en">
>             <seg>Search the Art of Reading Gallery</seg>
> ```

This doesn't seem to have been replaced by anything equivalent. Just checking.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 350 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kil-9kVV4tqBb1HBZZX:-Kil-9kVV4tqBb1HBZZY:blei1wf) ([raw file](https://github.com/sillsdev/libpalaso/blob/647b26e1c9d94eb75f177f9b5cf7d2495f2a06d8/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L350)):*
> ```Smalltalk
> 		}
> 
> 		private string ImageFolderForInternalPath(string rawInternalPath, out string relativePath)
> ```

In one use of this you append the relative path to the root path. In 2 others the relative path is ignored.
This is functional, but I wonder if it would be cleaner after a refactor to avoid the out parameter, say taking a bool to append the relative path when desired?

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 703 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kil15C8_ieR3enzyGV_:-Kil15C8_ieR3enzyGVa:bswzp9f) ([raw file](https://github.com/sillsdev/libpalaso/blob/647b26e1c9d94eb75f177f9b5cf7d2495f2a06d8/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L703)):*
> ```Smalltalk
> 		}
> 
> 		void RestoreEditabilityOfCollection(string path)
> ```

Is Editability the right word here? EnabledState seems more accurate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

jasonleenaylor:
Reviewed 1 of 12 files at r2, 1 of 6 files at r4.
Review status: 4 of 17 files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

jasonleenaylor:
Reviewed 4 of 11 files at r3, 2 of 6 files at r4.
Review status: 10 of 17 files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

JohnThomson:
Review status: 9 of 17 files reviewed at latest revision, 10 unresolved discussions.

---

*[DistFiles/Palaso.en.tmx, line 246 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-KiksnQtb2nq8Ze9gYXU:-Kil2XiOqFsGsSD8g9L9:b-cl69i) ([raw file](https://github.com/sillsdev/libpalaso/blob/cdc0bf5447b25999fde1409904de2dfb30f27bf3/DistFiles/Palaso.en.tmx#L246)):*

This doesn't seem to have been replaced by anything equivalent. Just checking.
</blockquote></details>

The heading at the top of the control is now "Image Galleries" which has duly been added.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 502 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kig-ReJKXxvyNwNkHb7:-KilBBfzl6dqQ3Zhncwz:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/e8f20a3844da914db781d3007cc68e3fce3f1846/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L502)):*

Correction: the above should be " before 3.3, the first version to put it in the ImageCollections folder"
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 350 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kil-9kVV4tqBb1HBZZX:-KilAz2TDPlxH6n5h891:b2tjsya) ([raw file](https://github.com/sillsdev/libpalaso/blob/647b26e1c9d94eb75f177f9b5cf7d2495f2a06d8/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L350)):*

In one use of this you append the relative path to the root path. In 2 others the relative path is ignored.
This is functional, but I wonder if it would be cleaner after a refactor to avoid the out parameter, say taking a bool to append the relative path when desired?
</blockquote></details>

I can't think of a reasonable name for such a method. Instead I've made an overload without the out param, and a new method GetFileSystemPathFromInternalPath for when you want the actual path.

---

*[SIL.Windows.Forms/ImageGallery/ImageCollection.cs, line 703 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-Kil15C8_ieR3enzyGV_:-KilBTfA2rdfrUBt1JdR:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/647b26e1c9d94eb75f177f9b5cf7d2495f2a06d8/SIL.Windows.Forms/ImageGallery/ImageCollection.cs#L703)):*

Is Editability the right word here? EnabledState seems more accurate.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482)*

jasonleenaylor:
:lgtm:

---

Reviewed 2 of 9 files at r1, 4 of 12 files at r2, 3 of 11 files at r3, 1 of 6 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/482#-:-KipJHP5eU7IwpJMWrt6:bnfp4nl)*

PR #486, opened 2017-04-25
Improved the SafeInvoke method in response to JohnT's feedback

tombogle:
on PR #480 and added unit tests, which showed need for additional improvements to improve robustness and handle edge cases.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/486)

### DIFF
--- a/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs
+++ b/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs
@@ -12,7 +12,12 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 	public class ControlExtensionsTests
 	{
 		private Control _control;
+		private Exception _threadException;
 
+		// Theoretically, setting up the handler to call ApplicationOnThreadException is not needed on Linux because, for some
+		// reason, Mono works differently than the Windows implementation of .Net, which *does* fire the ThreadException event.
+		// But since it doesn't hurt anything, rather than making this conditionally compiled, we'll leave it here for the Mono
+		// builds, too, just in case they ever make Mono work the same way.
 		[TestFixtureSetUp]
 		public void TestFixtureSetup()
 		{
@@ -28,6 +33,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		[SetUp]
 		public void Setup()
 		{
+			_threadException = null;
 			_control = new Control();
 		}
 
@@ -61,6 +67,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		public void SafeInvoke_OnUiThread_HandleNotCreated_IgnoreAll_ReturnsWithoutInvokingOrThrowing()
 		{
 			int i = 0;
+			// TODO: Check the return result for this call and all other tests.
 			_control.SafeInvoke(() => { i++; }, "HandleNotCreatedTest", ControlExtensions.ErrorHandlingAction.IgnoreAll);
 			Assert.AreEqual(0, i);
 		}
@@ -135,8 +142,10 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 			while (worker.IsBusy)
 				Application.DoEvents();
 			Assert.AreEqual("First action got invoked", confirmationMessage);
-			Assert.IsNotNull(resultOfSafeInvokeCall1);
-			Assert.IsNotNull(resultOfSafeInvokeCall2);
+			Assert.IsNotNull(resultOfSafeInvokeCall1, "First call to SafeInvoke should have returned a non-null IAsyncResult.");
+			Assert.IsFalse(resultOfSafeInvokeCall1.CompletedSynchronously, "Expected asynchronous invocation of action 1.");
+			Assert.IsNull(_threadException, "If this is not null, then the second call to SafeInvoke must have thrown an exception.");
+			Assert.IsNotNull(resultOfSafeInvokeCall2, "Second call to SafeInvoke should have returned a non-null IAsyncResult (even though it is later dequeued).");
 			_control.EndInvoke(resultOfSafeInvokeCall1);
 			Assert.Throws<ObjectDisposedException>(() => _control.EndInvoke(resultOfSafeInvokeCall2));
 		}
@@ -254,11 +263,13 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 			Exception exceptionThrownBySafeInvoke = null;
 			_control.CreateControl();
 			IAsyncResult resultOfSafeInvoke = null;
+			bool invokeWasRequired = false;
 			bool actionWasInvoked = false;
 			var worker = new BackgroundWorker();
 			worker.DoWork += (sender, args) =>
 			{
 				var ctrl = (Control)args.Argument;
+				invokeWasRequired = ctrl.InvokeRequired;
 				try
 				{
 					resultOfSafeInvoke = ctrl.SafeInvoke(() =>
@@ -278,6 +289,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 				Application.DoEvents();
 			Application.DoEvents(); // Just for good measure. We're desperate!
 
+			Assert.IsTrue(invokeWasRequired);
 			Assert.IsTrue(actionWasInvoked);
 			Assert.IsNull(exceptionThrownBySafeInvoke);
 			Assert.IsNotNull(resultOfSafeInvoke);
@@ -290,6 +302,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		private void ApplicationOnThreadException(object sender, ThreadExceptionEventArgs threadExceptionEventArgs)
 		{
 			// This will be handled and returned when EndInvoke is called.
+			_threadException = threadExceptionEventArgs.Exception;
 		}
 
 		[Test]
--- a/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs
+++ b/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs
@@ -11,7 +11,6 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 	[TestFixture]
 	public class ControlExtensionsTests
 	{
-		private Exception _threadException;
 		private Control _control;
 
 		[TestFixtureSetUp]
@@ -29,7 +28,6 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		[SetUp]
 		public void Setup()
 		{
-			_threadException = null;
 			_control = new Control();
 		}
 
@@ -111,12 +109,14 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		{
 			string confirmationMessage = null;
 			_control.CreateControl();
+			IAsyncResult resultOfSafeInvokeCall1 = null;
+			IAsyncResult resultOfSafeInvokeCall2 = null;
 			var worker = new BackgroundWorker();
 			worker.DoWork += (sender, args) =>
 			{
 				var ctrl = (Control)args.Argument;
 				Trace.WriteLine("SafeInvoke 1");
-				ctrl.SafeInvoke(() =>
+				resultOfSafeInvokeCall1 = ctrl.SafeInvoke(() =>
 				{
 					Trace.WriteLine("invoking 1");
 					confirmationMessage = "First action got invoked";
@@ -124,7 +124,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 					ctrl.Dispose();
 				}, "Getting initial value, resetting control text, and disposing control.");
 				Trace.WriteLine("SafeInvoke 2");
-				ctrl.SafeInvoke(() =>
+				resultOfSafeInvokeCall2 = ctrl.SafeInvoke(() =>
 				{
 					Assert.Fail("This should have been de-queued when the control was disposed.");
 				}, "DisposedAfterInvokingOnUiThread", errorHandling);
@@ -135,6 +135,10 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 			while (worker.IsBusy)
 				Application.DoEvents();
 			Assert.AreEqual("First action got invoked", confirmationMessage);
+			Assert.IsNotNull(resultOfSafeInvokeCall1);
+			Assert.IsNotNull(resultOfSafeInvokeCall2);
+			_control.EndInvoke(resultOfSafeInvokeCall1);
+			Assert.Throws<ObjectDisposedException>(() => _control.EndInvoke(resultOfSafeInvokeCall2));
 		}
 
 		[Test]
@@ -249,6 +253,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		{
 			Exception exceptionThrownBySafeInvoke = null;
 			_control.CreateControl();
+			IAsyncResult resultOfSafeInvoke = null;
 			bool actionWasInvoked = false;
 			var worker = new BackgroundWorker();
 			worker.DoWork += (sender, args) =>
@@ -256,7 +261,7 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 				var ctrl = (Control)args.Argument;
 				try
 				{
-					ctrl.SafeInvoke(() =>
+					resultOfSafeInvoke = ctrl.SafeInvoke(() =>
 					{
 						actionWasInvoked = true;
 						throw new InvalidOperationException("Blah");
@@ -275,13 +280,16 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 
 			Assert.IsTrue(actionWasInvoked);
 			Assert.IsNull(exceptionThrownBySafeInvoke);
-			Assert.IsNotNull(_threadException);
-			Assert.AreEqual(typeof(InvalidOperationException), _threadException.GetType());
+			Assert.IsNotNull(resultOfSafeInvoke);
+			var ex = Assert.Throws<InvalidOperationException>(() => _control.EndInvoke(resultOfSafeInvoke));
+			Assert.AreEqual("Blah", ex.Message);
+			//Assert.IsNotNull(_threadException);
+			//Assert.AreEqual(typeof(InvalidOperationException), _threadException.GetType());
 		}
 
 		private void ApplicationOnThreadException(object sender, ThreadExceptionEventArgs threadExceptionEventArgs)
 		{
-			_threadException = threadExceptionEventArgs.Exception;
+			// This will be handled and returned when EndInvoke is called.
 		}
 
 		[Test]
@@ -310,7 +318,6 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 			Assert.IsFalse(finished);
 			while (worker.IsBusy)
 				Application.DoEvents();
-			Assert.IsNull(_threadException);
 			Assert.IsNotNull(exceptionThrownBySafeInvoke);
 			string nestedExceptions = "";
 			Exception ex = exceptionThrownBySafeInvoke;
--- a/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs
+++ b/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.IO;
 using System.Threading;
 using System.Windows.Forms;
 using NUnit.Framework;
@@ -11,7 +12,8 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 	[TestFixture]
 	public class ControlExtensionsTests
 	{
-		Exception _threadException;
+		private Exception _threadException;
+		private Control _control;
 
 		[TestFixtureSetUp]
 		public void TestFixtureSetup()
@@ -29,6 +31,14 @@ namespace SIL.Windows.Forms.Tests.ControlExtensionsTests
 		public void Setup()
 		{
 			_threadException = null;
+			_control = new Control();
+		}
+
+		[TearDown]
+		public void Teardown()
+		{
+			if (!_control.IsDisposed)
+				_control.Dispose();
 		}
 
 		[TestCase(ControlExtensions.ErrorHandlingAction.IgnoreAll)]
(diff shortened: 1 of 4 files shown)


JohnThomson:
Nice piece of work. A few details that I think are worth improving...see what you think.

---

Reviewed 3 of 3 files at r1.

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 107 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_n4sB22gYQpvlJ1NT:-Ki_n4sB22gYQpvlJ1NU:bnicvpc) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L107)):*
> ```Smalltalk
> 		/// destroyed, the action is dequeued, and an ObjectDisposedException is registered in the entry. That exception is available via
> 		/// EndInvoke, but since SafeInvoke does not return the IAsyncResult, this test cannot show that.
> 		/// Not that the first SafeInvoke in this test also tests the normal case of an action invoked asynchronously on a non-UI thread.
> ```

Not => Note?

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_o1N8PA9ih06bfQ-f:-Ki_o1N9RgEvu5_RRvEm:bvmjsiw) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L148)):*
> ```Smalltalk
> 				Trace.WriteLine("SafeInvoke 1");
> 				ctrl.SafeInvoke(() => { textFromControl = ctrl.Text; ctrl.Text = "Final Text"; }, "Force synchronous.", forceSynchronous: true);
> 				Assert.AreEqual("Initial Text", textFromControl);
> ```

This surprises me. Isn't the meaning of forceSynchronous: true that the calling thread should block until the action has been done on the main thread? I would have expected this result...perhaps unreliably since a race condition is involved...with forceSynchronous:false. With it true, the action should actually take place before SafeInvoke returns on the background thread.
I don't see a test involving forceSynchronous false on a background thread...is it possible to write one that reveals the difference in behavior?

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 138 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_uRohNIP_eQgY2hbn:-Ki_uRohNIP_eQgY2hbo:b-wfe0xd) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L138)):*
> ```Smalltalk
> 				// Technically, if forceSynchronous is false, we could call control.BeginInvoke, but then we'd need to wrap it in a
> 				// delegate (as above) to re-check to see if the control has been disposed. Probably not worth it. If we're already
> 				// on the UI thread, even if synchronous execution wasn't required, it should be okay.
> ```

It seems counter-intuitive to me that we would WANT the action delayed until some unpredictable later time when we're already on the thread that can perform it immediately. The purpose of forceSynchronous: false is so that a background thread doesn't have to wait for the UI thread if it doesn't matter what order things happen. If it's false, by definition the caller doesn't care whether action happens before or after SafeInvoke returns. So (unless the client has a bug) it is SURELY OK to do action() first.
So I'd make the comment something like, "Technically, if forceSynchronous is false, we could call control.BeginInvoke, but that just complicates our code (another chance for the control to get disposed before action() happens), and by definition the order in which things happen doesn't matter if we aren't forcing things to be synchronous"

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 160 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_rQX2a8rr1tYetA7D:-Ki_rQX2a8rr1tYetA7E:b-sv2yzj) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L160)):*
> ```Smalltalk
> 				catch (InvalidOperationException e)
> 				{
> 					if (treatInvalidOperationExceptionAsObjectDisposedException)
> ```

This has me puzzled, so probably at least a comment is needed.
This flag is false while executing the original action, so the only way I can see for an exception to happen with it true is that control.Invoke() fails before it begins to execute the synchronousAction.
Presumably the only likely (possible?) scenario for this is that the control gets disposed between the time when we call Invoke and the time something on the UI thread attempts the action? And for some reason this produces an InvalidOperationException instead of ObjectDisposedException?

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 163 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_sS-KphkMmsQzzJEy:-Ki_sS-KphkMmsQzzJEz:b-92pwe) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L163)):*
> ```Smalltalk
> 					{
> 						if (errorHandling == ErrorHandlingAction.Throw)
> 							throw new ObjectDisposedException("SafeInvoke called after the control was disposed. (" + nameForErrorReporting + ")", e);
> ```

If my understanding above is correct, technically, the problem isn't that SafeInvoke was called after the control was disposed...that would be caught above...but that the control was disposed (in a race condition) between the time SafeInvoke was called and the time the UI thread actually got around to the action.
Seems to me it is WELL worth calling out this difference in the error message...any clue to the nature of a race condition is to be treasured.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-:-Ki_tkER9ZPmq9l7I2LH:b-l2ivu0)*

tombogle:
*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 107 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_n4sB22gYQpvlJ1NT:-KiaivK_YTIyMhR_fvyK:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L107)):*

Not => Note?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

tombogle:
*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_o1N8PA9ih06bfQ-f:-Kiaj-_rPmYVlQakfNbV:b-slsyyw) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L148)):*

This surprises me. Isn't the meaning of forceSynchronous: true that the calling thread should block until the action has been done on the main thread? I would have expected this result...perhaps unreliably since a race condition is involved...with forceSynchronous:false. With it true, the action should actually take place before SafeInvoke returns on the background thread.
I don't see a test involving forceSynchronous false on a background thread...is it possible to write one that reveals the difference in behavior?
</blockquote></details>

If I understand your comment correctly, I think you mis-read the code. I've broken out the lines of the delegate to make it more readable. Hope that clears up the confusion. In the synchronous case, the action should be executed immediately before SafeInvoke returns, so the textFromControl variable will be set (to the *initial* value). The only "normal" test for asynchronous SafeInvoke was buried inside a non-normal test. I've broken it out into its  own unit test now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 138 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_uRohNIP_eQgY2hbn:-KiakzH5HeqtkTzBg_Tz:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L138)):*

It seems counter-intuitive to me that we would WANT the action delayed until some unpredictable later time when we're already on the thread that can perform it immediately. The purpose of forceSynchronous: false is so that a background thread doesn't have to wait for the UI thread if it doesn't matter what order things happen. If it's false, by definition the caller doesn't care whether action happens before or after SafeInvoke returns. So (unless the client has a bug) it is SURELY OK to do action() first.
So I'd make the comment something like, "Technically, if forceSynchronous is false, we could call control.BeginInvoke, but that just complicates our code (another chance for the control to get disposed before action() happens), and by definition the order in which things happen doesn't matter if we aren't forcing things to be synchronous"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 160 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_rQX2a8rr1tYetA7D:-Kial4HdcuVzEeFRreEH:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L160)):*

This has me puzzled, so probably at least a comment is needed.
This flag is false while executing the original action, so the only way I can see for an exception to happen with it true is that control.Invoke() fails before it begins to execute the synchronousAction.
Presumably the only likely (possible?) scenario for this is that the control gets disposed between the time when we call Invoke and the time something on the UI thread attempts the action? And for some reason this produces an InvalidOperationException instead of ObjectDisposedException?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 163 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_sS-KphkMmsQzzJEy:-KialAF1wURRDOyHbclh:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L163)):*

If my understanding above is correct, technically, the problem isn't that SafeInvoke was called after the control was disposed...that would be caught above...but that the control was disposed (in a race condition) between the time SafeInvoke was called and the time the UI thread actually got around to the action.
Seems to me it is WELL worth calling out this difference in the error message...any clue to the nature of a race condition is to be treasured.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

JohnThomson:
Much improved...couple of suggestions if you patience isn't exhausted. If it is, I wouldn't strongly object to merging as-is.

---

Reviewed 2 of 2 files at r2.

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 148 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_o1N8PA9ih06bfQ-f:-Kiern0eKJpdsiVvkN8m:bkhar5y) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L148)):*

If I understand your comment correctly, I think you mis-read the code. I've broken out the lines of the delegate to make it more readable. Hope that clears up the confusion. In the synchronous case, the action should be executed immediately before SafeInvoke returns, so the textFromControl variable will be set (to the *initial* value). The only "normal" test for asynchronous SafeInvoke was buried inside a non-normal test. I've broken it out into its  own unit test now.
</blockquote></details>

Yes, I mis-read. See new comment for what I think threw me off.

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 158 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Kier-DtQuBsn4AZAr6Z:-Kier-DtQuBsn4AZAr6_:bterpvm) ([raw file](https://github.com/sillsdev/libpalaso/blob/673c0871536ec5948b9c381b1f4fd8cea966930d/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L158)):*
> ```Smalltalk
> 				Trace.WriteLine("SafeInvoke 1");
> 				ctrl.SafeInvoke(() => { Trace.WriteLine("invoking 1"); textFromControl = ctrl.Text; ctrl.Text = "Final Text"; }, "Getting initial value and resetting control text.");
> 				Assert.IsNull(textFromControl);
> ```

This one is actually a race condition, isn't it? Nothing guarantees that the system won't choose to run the async block we are invoking before it returns control to our thread. Or does the way Invoke works prevents it from running until DoEvents()? If it's not a race condition a comment explaining how you know that would be nice. (Sorry to keep complaining after you already went out of your way to add this test after my mis-reading.)

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 184 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Kiepcl6LZsqS0IQr8Mj:-Kiepcl6LZsqS0IQr8Mk:b-7wvle7) ([raw file](https://github.com/sillsdev/libpalaso/blob/673c0871536ec5948b9c381b1f4fd8cea966930d/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L184)):*
> ```Smalltalk
> 					ctrl.Text = "Final Text";
> 				}, "Force synchronous.", forceSynchronous: true);
> 				Assert.AreEqual("Initial Text", textFromControl, "Since the invoke was done synchronously, this should have been set before SafeInvoke returned.");
> ```

Hmm, I think the thing that threw me off is that the most obvious thing the SafeInvoke action does is set something to "Final Text" and here right afterwards you are testing that the value of something is "Initial Text". A test that something is "Initial" very strongly suggests testing that a change has NOT happened yet. The comment on the Assert helps. But you could equally (or as well) test that ctrl.Text Is.EqualTo("Final Text") which would be much more obviously looking for a change done by the safeInvoke.
Is there any way to be sure this passes because the synchronous call was properly forced rather then just because a race condition usually turns out the way we want? (I hate threading!)

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 160 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_rQX2a8rr1tYetA7D:-KiesZMPqBKGCztLNllM:b-m2gq6) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L160)):*

Done.
</blockquote></details>

Nice job, thanks.

---

*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 173 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-KiesseX5wZMSZwYOYWg:-KiesseX5wZMSZwYOYWh:b-z3doj7) ([raw file](https://github.com/sillsdev/libpalaso/blob/673c0871536ec5948b9c381b1f4fd8cea966930d/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L173)):*
> ```Smalltalk
> 						if (errorHandling == ErrorHandlingAction.Throw)
> 							throw new ObjectDisposedException("Control was disposed before " + (forceSynchronous ? "Invoke" : "BeginInvoke") +
> 								" could be called. To suppress this kind of race-condition error, call SafeInvoke with IgnoreIfDisposed. (" + nameForErrorReporting + ")", e);
> ```

Alternatively, you could avoid invoking code on controls you are about to dispose :-). (No, I don't think that needs to be said.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-:-Kiet57qlkS37oBPS_dJ:b-lm41bg)*

tombogle:
*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 158 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Kier-DtQuBsn4AZAr6Z:-Kijkuk88AquVgJqS2Ps:b-1918dc) ([raw file](https://github.com/sillsdev/libpalaso/blob/673c0871536ec5948b9c381b1f4fd8cea966930d/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L158)):*

This one is actually a race condition, isn't it? Nothing guarantees that the system won't choose to run the async block we are invoking before it returns control to our thread. Or does the way Invoke works prevents it from running until DoEvents()? If it's not a race condition a comment explaining how you know that would be nice. (Sorry to keep complaining after you already went out of your way to add this test after my mis-reading.)
</blockquote></details>

Never fear. My patience is inexhaustible! :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

tombogle:
*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 184 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Kiepcl6LZsqS0IQr8Mj:-KijrA2eOjK1uoP_zPX7:b-x9jufe) ([raw file](https://github.com/sillsdev/libpalaso/blob/673c0871536ec5948b9c381b1f4fd8cea966930d/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L184)):*

Hmm, I think the thing that threw me off is that the most obvious thing the SafeInvoke action does is set something to "Final Text" and here right afterwards you are testing that the value of something is "Initial Text". A test that something is "Initial" very strongly suggests testing that a change has NOT happened yet. The comment on the Assert helps. But you could equally (or as well) test that ctrl.Text Is.EqualTo("Final Text") which would be much more obviously looking for a change done by the safeInvoke.
Is there any way to be sure this passes because the synchronous call was properly forced rather then just because a race condition usually turns out the way we want? (I hate threading!)
</blockquote></details>

Done. I added an Assert check for both changes to avoid confusion, and put in a few Thread.Sleep calls in the invoked code to demonstrate that the synchronous call blocks both threads so that neither one can do anything.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

JohnThomson:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-:-KikL4QbWkUqQKW0uSc7:bnfp4nl)*

JohnThomson:
Wonder whether we should seek a review by someone like Eberhard or Damien that's smart about threading. Otherwise, I think you're good to merge.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-:-KikLBDjsMtkOm7RAWjN:b1e1pp2)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 173 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-KiesseX5wZMSZwYOYWg:-Kj5e74iJP-tTXkPSC4y:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/673c0871536ec5948b9c381b1f4fd8cea966930d/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L173)):*

Alternatively, you could avoid invoking code on controls you are about to dispose :-). (No, I don't think that needs to be said.)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

tombogle:
*[SIL.Windows.Forms/Extensions/ControlExtensions.cs, line 160 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Ki_rQX2a8rr1tYetA7D:-Kj5eERe9M_EDGIcrv16:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/374a8489ff8b9cae7d800f277b52bd1e14e3d7a6/SIL.Windows.Forms/Extensions/ControlExtensions.cs#L160)):*

Nice job, thanks.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486)*

JohnThomson:
:lgtm: with two very minor comment suggestions...Feel free to merge after (or even without) fixing them.

---

Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 187 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Kj8nfymOUy9_ARrJMjH:-Kj8nfynV8_JZ6wjQBql:b-swvql4) ([raw file](https://github.com/sillsdev/libpalaso/blob/b3b682b34ed533f9291887d50631af5a15760214/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L187)):*
> ```Smalltalk
> 			Assert.IsFalse(resultOfSafeInvokeCall.CompletedSynchronously, "Expected asynchronous invocation of action.");
> 			Assert.IsTrue(resultOfSafeInvokeCall.IsCompleted, "Asynchronous action should have completed.");
> 			Assert.IsNull(_threadException, "If this is not null, then the second call to SafeInvoke must have thrown an exception.");
> ```

Don't think we have a second call in this test?

---

*[SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs, line 220 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-Kj8npp1ULdxxKjUjYqJ:-Kj8npp2F_yGyVQIzRfh:b1vbhw1) ([raw file](https://github.com/sillsdev/libpalaso/blob/b3b682b34ed533f9291887d50631af5a15760214/SIL.Windows.Forms.Tests/Extensions/ControlExtensionsTests.cs#L220)):*
> ```Smalltalk
> 				Application.DoEvents(); // Even though the SafeInvoke call is synchronous, the worker thread obviously isn't, so we need to wait for it to finish.
> 			Assert.AreEqual("Final Text", _control.Text, "This just proves that we don't get here without having executed the worker thread's code.");
> 			Assert.IsNull(resultOfSafeInvokeCall, "When SafeInvoke results in a synchronous call, not IAsyncResult should be returned.");
> ```

not => no?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/486#-:-Kj8pLPeeCM3IAeKTjkd:bf0hd3i)*

PR #490, opened 2017-04-26
Added VerseControl and dependencies from Paratext to be used in other applications.

FoolRunning:
This changeset is fallout from splitting out ParatextShared.dll into an actually-shareable library. Requests from @hindlemail to make the new DLL not reference any UI components made it necessary to find a new home for **VerseControl** (which is used by Glyssen).

Dependencies mainly include classes for versification information (**ScrVers**, **Versification**), reference-handling (**VerseRef**, **Cannon**, **BookSet**), and a few small custom controls (**EnterTextBox**, **HorizontalSplitter**, **SafeComboBox**).

Note that there were some breaking changes to the existing Palaso API (e.g. removing all of **ScrReference** and massive changes to the existing **ScrVers** class), but @tombogle searched through projects and only projects that he is a part of are the users of the breaking changes - and he's fine with the breakage. I'm fine with having this changeset pushed to Palaso 4.0 instead, if needed.

Also note that I did replacements of spaces with tabs and some files (mainly **BCVRef** and **ReflectionHelper**) had some unintended changes as a result. If someone feels strongly, I can attempt to fix these so spurious changes are removed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/490)


FoolRunning:
Working on fixing the build errors and failing tests... will update the PR.
(forgot to try in Release mode... everything works in Debug mode)

JohnThomson:
Guidelines we have been following call for breaking API changes to involve a major version number change, so this ought to go into 4.0 unless there is a very good reason it must be in 3.1.

FoolRunning:
Okay, I will redo this PR on the master branch.

PR #494, opened 2017-05-03
Added a FlexibleMessageBox

tombogle:
I got this from http://www.codeproject.com/Articles/601900/FlexibleMessageBox (and made a few changes to support better localization and overriding of the LinkClicked handler).
Some of the benefits of this over the standard .Net MessageBox are:
* It can be resized and the content is correctly word-wrapped
* It tries to auto-size the width to show the longest text row
* It never exceeds the current desktop working area
* It displays a vertical scrollbar when needed
* It supports (auto-detected) hyperlinks in text
* It allows copying of messages to clipboard

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/494)

### DIFF
--- a/DistFiles/Palaso.en.tmx
+++ b/DistFiles/Palaso.en.tmx
@@ -97,7 +97,7 @@
       </tu>
       <tu tuid="Common.NoButton">
          <tuv xml:lang="en">
-            <seg>&No</seg>
+            <seg>&amp;No</seg>
          </tuv>
       </tu>
       <tu tuid="Common.OKButton">
--- a/DistFiles/Palaso.en.tmx
+++ b/DistFiles/Palaso.en.tmx
@@ -75,9 +75,19 @@
             <seg>Waiting for other {0} to finish...</seg>
          </tuv>
       </tu>
+      <tu tuid="Common.AbortButton">
+         <tuv xml:lang="en">
+            <seg>&amp;Abort</seg>
+         </tuv>
+      </tu>
       <tu tuid="Common.CancelButton">
          <tuv xml:lang="en">
             <seg>&amp;Cancel</seg>
+         </tuv>
+      </tu>
+      <tu tuid="Common.IgnoreButton">
+         <tuv xml:lang="en">
+            <seg>&amp;Ignore</seg>
          </tuv>
       </tu>
       <tu tuid="Common.No">
@@ -85,9 +95,19 @@
             <seg>No</seg>
          </tuv>
       </tu>
+      <tu tuid="Common.NoButton">
+         <tuv xml:lang="en">
+            <seg>&No</seg>
+         </tuv>
+      </tu>
       <tu tuid="Common.OKButton">
          <tuv xml:lang="en">
             <seg>&amp;OK</seg>
+         </tuv>
+      </tu>
+      <tu tuid="Common.RetryButton">
+         <tuv xml:lang="en">
+            <seg>&amp;Retry</seg>
          </tuv>
       </tu>
       <tu tuid="Common.Yes">
@@ -95,19 +115,9 @@
             <seg>Yes</seg>
          </tuv>
       </tu>
-      <tu tuid="Common.Abort">
+      <tu tuid="Common.YesButton">
          <tuv xml:lang="en">
-            <seg>&amp;Abort</seg>
-         </tuv>
-      </tu>
-      <tu tuid="Common.Ignore">
-         <tuv xml:lang="en">
-            <seg>&amp;Ignore</seg>
-         </tuv>
-      </tu>
-      <tu tuid="Common.Retry">
-         <tuv xml:lang="en">
-            <seg>&amp;Retry</seg>
+            <seg>&amp;Yes</seg>
          </tuv>
       </tu>
       <tu tuid="DialogBoxes.ConfirmRecycleDialog.MessageForMultipleItems">
--- a/DistFiles/Palaso.en.tmx
+++ b/DistFiles/Palaso.en.tmx
@@ -95,6 +95,21 @@
             <seg>Yes</seg>
          </tuv>
       </tu>
+      <tu tuid="Common.Abort">
+         <tuv xml:lang="en">
+            <seg>&amp;Abort</seg>
+         </tuv>
+      </tu>
+      <tu tuid="Common.Ignore">
+         <tuv xml:lang="en">
+            <seg>&amp;Ignore</seg>
+         </tuv>
+      </tu>
+      <tu tuid="Common.Retry">
+         <tuv xml:lang="en">
+            <seg>&amp;Retry</seg>
+         </tuv>
+      </tu>
       <tu tuid="DialogBoxes.ConfirmRecycleDialog.MessageForMultipleItems">
          <note>Parameter {0} is a description of the things being deleted (e.g., "The selected files")</note>
          <tuv xml:lang="en">
--- a/DistFiles/Palaso.es.tmx
+++ b/DistFiles/Palaso.es.tmx
@@ -39,7 +39,7 @@
     </tu>
 	<tu tuid="Common.No">
 	  <tuv xml:lang="en">
-		<seg>&amp;No</seg>
+		<seg>No</seg>
 	  </tuv>
 	  <tuv xml:lang="es">
 		<seg>No</seg>
--- a/DistFiles/Palaso.es.tmx
+++ b/DistFiles/Palaso.es.tmx
@@ -13,6 +13,14 @@
 		<seg>Acerca de</seg>
 	  </tuv>
 	</tu>
+    <tu tuid="Common.AbortButton">
+      <tuv xml:lang="en">
+        <seg>&amp;Abort</seg>
+      </tuv>
+	  <tuv xml:lang="es">
+	    <seg>&amp;Abortar</seg>
+	  </tuv>
+    </tu>
 	<tu tuid="Common.CancelButton">
 	  <tuv xml:lang="en">
 		<seg>&amp;Cancel</seg>
@@ -21,31 +29,7 @@
 		<seg>&amp;Cancelar</seg>
 	  </tuv>
 	</tu>
-	<tu tuid="Common.No">
-	  <tuv xml:lang="en">
-		<seg>No</seg>
-	  </tuv>
-	  <tuv xml:lang="es">
-		<seg>No</seg>
-	  </tuv>
-	</tu>
-	<tu tuid="Common.OKButton">
-	  <tuv xml:lang="en">
-		<seg>&amp;OK</seg>
-	  </tuv>
-	  <tuv xml:lang="es">
-		<seg>Aceptar</seg>
-	  </tuv>
-	</tu>
-    <tu tuid="Common.Abort">
-      <tuv xml:lang="en">
-        <seg>&amp;Abort</seg>
-      </tuv>
-	  <tuv xml:lang="es">
-	    <seg>&amp;Abortar</seg>
-	  </tuv>
-    </tu>
-    <tu tuid="Common.Ignore">
+    <tu tuid="Common.IgnoreButton">
       <tuv xml:lang="en">
         <seg>&amp;Ignore</seg>
       </tuv>
@@ -53,7 +37,31 @@
 	    <seg>&amp;Ignorar</seg>
 	  </tuv>
     </tu>
-    <tu tuid="Common.Retry">
+	<tu tuid="Common.No">
+	  <tuv xml:lang="en">
+		<seg>&amp;No</seg>
+	  </tuv>
+	  <tuv xml:lang="es">
+		<seg>No</seg>
+	  </tuv>
+	</tu>
+	<tu tuid="Common.NoButton">
+	  <tuv xml:lang="en">
+		<seg>&amp;No</seg>
+	  </tuv>
+	  <tuv xml:lang="es">
+		<seg>&amp;No</seg>
+	  </tuv>
+	</tu>
+	<tu tuid="Common.OKButton">
+	  <tuv xml:lang="en">
+		<seg>&amp;OK</seg>
+	  </tuv>
+	  <tuv xml:lang="es">
+		<seg>&amp;Aceptar</seg>
+	  </tuv>
+	</tu>
+    <tu tuid="Common.RetryButton">
       <tuv xml:lang="en">
         <seg>&amp;Retry</seg>
       </tuv>
@@ -61,6 +69,22 @@
 	    <seg>&amp;Reintentar</seg>
 	  </tuv>
     </tu>
+	<tu tuid="Common.Yes">
+	  <tuv xml:lang="en">
+		<seg>Yes</seg>
+	  </tuv>
+	  <tuv xml:lang="es">
+		<seg>Si</seg>
+	  </tuv>
+	</tu>
+	<tu tuid="Common.YesButton">
+	  <tuv xml:lang="en">
+		<seg>&amp;Yes</seg>
+	  </tuv>
+	  <tuv xml:lang="es">
+		<seg>&amp;Si</seg>
+	  </tuv>
+	</tu>
 	<tu tuid="DialogBoxes.FileDlg.AllFilesLabel">
       <tuv xml:lang="en">
         <seg>All Files</seg>
@@ -86,14 +110,6 @@
         <seg>Paquetes de descarga de texto</seg>
       </tuv>
     </tu>
-	<tu tuid="Common.Yes">
-	  <tuv xml:lang="en">
-		<seg>Yes</seg>
-	  </tuv>
-	  <tuv xml:lang="es">
-		<seg>Si</seg>
-	  </tuv>
-	</tu>
 	<tu tuid="ImageToolbox.ArtOfReading">
 	  <tuv xml:lang="en">
 		<seg>Art Of Reading</seg>
--- a/DistFiles/Palaso.es.tmx
+++ b/DistFiles/Palaso.es.tmx
@@ -37,6 +37,30 @@
 		<seg>Aceptar</seg>
 	  </tuv>
 	</tu>
+    <tu tuid="Common.Abort">
+      <tuv xml:lang="en">
+        <seg>&amp;Abort</seg>
+      </tuv>
+	  <tuv xml:lang="es">
+	    <seg>&amp;Abortar</seg>
+	  </tuv>
+    </tu>
+    <tu tuid="Common.Ignore">
+      <tuv xml:lang="en">
+        <seg>&amp;Ignore</seg>
+      </tuv>
+	  <tuv xml:lang="es">
+	    <seg>&amp;Ignorar</seg>
+	  </tuv>
+    </tu>
+    <tu tuid="Common.Retry">
+      <tuv xml:lang="en">
+        <seg>&amp;Retry</seg>
+      </tuv>
+	  <tuv xml:lang="es">
+	    <seg>&amp;Reintentar</seg>
+	  </tuv>
+    </tu>
 	<tu tuid="DialogBoxes.FileDlg.AllFilesLabel">
       <tuv xml:lang="en">
         <seg>All Files</seg>
--- a/DistFiles/Palaso.fr.tmx
+++ b/DistFiles/Palaso.fr.tmx
@@ -13,6 +13,14 @@
         <seg>À propos de</seg>
       </tuv>
     </tu>
+    <tu tuid="Common.AbortButton">
+      <tuv xml:lang="en">
+        <seg>&amp;Abort</seg>
+      </tuv>
+	  <tuv xml:lang="fr">
+	    <seg>&amp;Annuler</seg>
+	  </tuv>
+    </tu>
     <tu tuid="Common.CancelButton">
       <tuv xml:lang="en">
         <seg>&amp;Cancel</seg>
@@ -21,12 +29,28 @@
         <seg>Annuler</seg>
       </tuv>
     </tu>
+    <tu tuid="Common.IgnoreButton">
+      <tuv xml:lang="en">
+        <seg>&amp;Ignore</seg>
+      </tuv>
+	  <tuv xml:lang="fr">
+	    <seg>&amp;Ignorer</seg>
+	  </tuv>
+    </tu>
     <tu tuid="Common.No">
       <tuv xml:lang="en">
         <seg>No</seg>
       </tuv>
       <tuv xml:lang="fr">
         <seg>Non</seg>
+      </tuv>
+    </tu>
+    <tu tuid="Common.NoButton">
+      <tuv xml:lang="en">
+        <seg>&amp;No</seg>
+      </tuv>
+      <tuv xml:lang="fr">
+        <seg>&amp;Non</seg>
       </tuv>
     </tu>
     <tu tuid="Common.OKButton">
@@ -37,6 +61,14 @@
         <seg>&amp;OK</seg>
       </tuv>
     </tu>
+    <tu tuid="Common.RetryButton">
+      <tuv xml:lang="en">
+        <seg>&amp;Retry</seg>
+      </tuv>
+	  <tuv xml:lang="fr">
+	    <seg>&amp;Réessayer</seg>
+	  </tuv>
+    </tu>
     <tu tuid="Common.Yes">
       <tuv xml:lang="en">
         <seg>Yes</seg>
@@ -45,29 +77,13 @@
         <seg>Oui</seg>
       </tuv>
     </tu>
-    <tu tuid="Common.Abort">
+    <tu tuid="Common.YesButton">
       <tuv xml:lang="en">
-        <seg>&amp;Abort</seg>
+        <seg>&amp;Yes</seg>
       </tuv>
-	  <tuv xml:lang="fr">
-	    <seg>&amp;Annuler</seg>
-	  </tuv>
-    </tu>
-    <tu tuid="Common.Ignore">
-      <tuv xml:lang="en">
-        <seg>&amp;Ignore</seg>
+      <tuv xml:lang="fr">
+        <seg>&amp;Oui</seg>
       </tuv>
-	  <tuv xml:lang="fr">
-	    <seg>&amp;Ignorer</seg>
-	  </tuv>
-    </tu>
-    <tu tuid="Common.Retry">
-      <tuv xml:lang="en">
-        <seg>&amp;Retry</seg>
-      </tuv>
-	  <tuv xml:lang="fr">
-	    <seg>&amp;Réessayer</seg>
-	  </tuv>
     </tu>
     <tu tuid="ImageToolbox.ArtOfReading">
       <tuv xml:lang="en">
--- a/DistFiles/Palaso.fr.tmx
+++ b/DistFiles/Palaso.fr.tmx
@@ -26,7 +26,7 @@
         <seg>&amp;Cancel</seg>
       </tuv>
       <tuv xml:lang="fr">
-        <seg>Annuler</seg>
+        <seg>&amp;Annuler</seg>
       </tuv>
     </tu>
     <tu tuid="Common.IgnoreButton">
--- a/DistFiles/Palaso.fr.tmx
+++ b/DistFiles/Palaso.fr.tmx
@@ -45,6 +45,30 @@
         <seg>Oui</seg>
       </tuv>
     </tu>
+    <tu tuid="Common.Abort">
+      <tuv xml:lang="en">
+        <seg>&amp;Abort</seg>
+      </tuv>
+	  <tuv xml:lang="fr">
+	    <seg>&amp;Annuler</seg>
+	  </tuv>
+    </tu>
+    <tu tuid="Common.Ignore">
+      <tuv xml:lang="en">
+        <seg>&amp;Ignore</seg>
+      </tuv>
+	  <tuv xml:lang="fr">
+	    <seg>&amp;Ignorer</seg>
+	  </tuv>
+    </tu>
+    <tu tuid="Common.Retry">
+      <tuv xml:lang="en">
+        <seg>&amp;Retry</seg>
+      </tuv>
+	  <tuv xml:lang="fr">
+	    <seg>&amp;Réessayer</seg>
+	  </tuv>
+    </tu>
     <tu tuid="ImageToolbox.ArtOfReading">
       <tuv xml:lang="en">
         <seg>Art Of Reading</seg>
--- a/SIL.Windows.Forms/FlexibleMessageBox.cs
+++ b/SIL.Windows.Forms/FlexibleMessageBox.cs
@@ -3,60 +3,58 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
-using System.Globalization;
 using System.Linq;
 using System.Windows.Forms;
 using L10NSharp;
-
 namespace SIL.Windows.Forms
 {
 	/*  FlexibleMessageBox – A flexible replacement for the .NET MessageBox
-	 * 
+	 *
 	 *  Author:		Jörg Reichert (public@jreichert.de)
 	 *  Contributors:   Thanks to: David Hall, Roink
 	 *  Version:		1.3
 	 *  Published at:   http://www.codeproject.com/Articles/601900/FlexibleMessageBox
-	 *  
+	 *
 	 ************************************************************************************************************
 	 * Features:
 	 *  - It can be simply used instead of MessageBox since all important static "Show"-Functions are supported
-	 *  - It is small, only one source file, which could be added easily to each solution 
+	 *  - It is small, only one source file, which could be added easily to each solution
 	 *  - It can be resized and the content is correctly word-wrapped
 	 *  - It tries to auto-size the width to show the longest text row
 	 *  - It never exceeds the current desktop working area
 	 *  - It displays a vertical scrollbar when needed
 	 *  - It does support hyperlinks in text
 	 *  - It allows copying of messages to clipboard
-	 * 
-	 *  Because the interface is identical to MessageBox, you can add this single source file to your project 
-	 *  and use the FlexibleMessageBox almost everywhere you use a standard MessageBox. 
-	 *  The goal was NOT to produce as many features as possible but to provide a simple replacement to fit my 
+	 *
+	 *  Because the interface is identical to MessageBox, you can add this single source file to your project
+	 *  and use the FlexibleMessageBox almost everywhere you use a standard MessageBox.
+	 *  The goal was NOT to produce as many features as possible but to provide a simple replacement to fit my
 	 *  own needs. Feel free to add additional features on your own, but please left my credits in this class.
-	 * 
+	 *
 	 ************************************************************************************************************
 	 * Usage examples:
-	 * 
+	 *
 	 *  FlexibleMessageBox.Show("Just a text");
-	 * 
-	 *  FlexibleMessageBox.Show("A text", "A caption"); 
-	 *  
-	 *  FlexibleMessageBox.Show("Some text with a link: www.google.com", 
+	 *
+	 *  FlexibleMessageBox.Show("A text", "A caption");
+	 *
+	 *  FlexibleMessageBox.Show("Some text with a link: www.google.com",
 	 *							"Some caption",
-	 *							MessageBoxButtons.AbortRetryIgnore, 
+	 *							MessageBoxButtons.AbortRetryIgnore,
 	 *							MessageBoxIcon.Information,
 	 *							MessageBoxDefaultButton.Button2);
-	 *  
-	 *  var dialogResult = FlexibleMessageBox.Show("Do you know the answer to life the universe and everything?", 
+	 *
+	 *  var dialogResult = FlexibleMessageBox.Show("Do you know the answer to life the universe and everything?",
 	 *											   "One short question",
-	 *											   MessageBoxButtons.YesNo);	 
-	 * 
+	 *											   MessageBoxButtons.YesNo);
+	 *
 	 ************************************************************************************************************
 	 *  THE SOFTWARE IS PROVIDED BY THE AUTHOR "AS IS", WITHOUT WARRANTY
 	 *  OF ANY KIND, EXPRESS OR IMPLIED. IN NO EVENT SHALL THE AUTHOR BE
 	 *  LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY ARISING FROM,
 	 *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OF THIS
 	 *  SOFTWARE.
-	 *  
+	 *
 	 ************************************************************************************************************
 	 * History:
 	 *	Adapted for use in libpalaso
@@ -70,17 +68,17 @@ namespace SIL.Windows.Forms
 	 *  - Tab handling is now corrected (only tabbing over the visible buttons)
 	 *  - Added standard MessageBox handling for ALT-Keyboard shortcuts
 	 *  - SetDialogSizes: Refactored completely: Corrected sizing and added caption driven sizing
-	 * 
+	 *
 	 *	Version 1.2 - 10.August 2013
 	 *   - Do not ShowInTaskbar anymore (original MessageBox is also hidden in taskbar)
 	 *   - Added handling for Escape-Button
 	 *   - Adapted top right close button (red X) to behave like MessageBox (but hidden instead of deactivated)
-	 * 
+	 *
 	 *	Version 1.1 - 14.June 2013
 	 *   - Some Refactoring
 	 *   - Added internal form class
 	 *   - Added missing code comments, etc.
-	 *  
+	 *
 	 *	Version 1.0 - 15.April 2013
 	 *   - Initial Version
 	 */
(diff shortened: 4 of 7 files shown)


andrew-polk:
JohnH suggests this be tested on mono before being merged.
JohnT suggests this be tested on a Windows machine not set to 100% font size (you may have already done that).

---

Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[DistFiles/Palaso.en.tmx, line 98 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDevn-Oi0L8QZWGCcw:-KjDevn-Oi0L8QZWGCcx:bmbdwmk) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/DistFiles/Palaso.en.tmx#L98)):*
> ```
>          </tuv>
>       </tu>
>       <tu tuid="Common.Abort">
> ```

I wonder if these should be AbortButton, etc. That would be more consistent with OKButton, etc. where the ampersand is part of the string.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 11 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDe_tWGtuN-NBDW-Ke:-KjDe_tWGtuN-NBDW-Kf:b-usg2io) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L11)):*
> ```Smalltalk
> using L10NSharp;
> 
> namespace SIL.Windows.Forms
> ```

Should this be in Widgets? And named BetterMessageBox to follow the pattern of similar classes already in libpalaso?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDg9mWdrUZS6Z1pj9E:-KjDg9mWdrUZS6Z1pj9F:bi246ph) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L14)):*
> ```Smalltalk
> {
>     /*  FlexibleMessageBox – A flexible replacement for the .NET MessageBox
>      * 
> ```

There is a fair amount of trailing whitespace in this file. Let's go ahead and eliminate it since we are starting fresh.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 96 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDfw4FQEXwDX7QEElN:-KjDfw4GUtluq8gCX5CO:btehc5q) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L96)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>
>         /// 0.2 means:  The FlexibleMessageBox can be at most half as wide as the working area.
>         /// 1.0 means:  The FlexibleMessageBox can be as wide as the working area.
>         ///
>         /// Default is: 70% of the working area width.
</details>

so if .7 is 70%, why is .2 50%? I'm guessing a typo? Same with other comments below.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 100 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDgdVbGiW2-lb6yYjD:-KjDgdVbGiW2-lb6yYjE:b-pajg7b) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L100)):*
> ```Smalltalk
>         /// Default is: 70% of the working area width.
>         /// </summary>
>         public static double MAX_WIDTH_FACTOR = 0.7;
> ```

Is all caps the standard for public static in libpalaso? Doesn't seem right to me...

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 124 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDh-eVpR7mtY3o4jvE:-KjDh-eVpR7mtY3o4jvF:b9c3dvq) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L124)):*
> "FlexibleMessageBoxBtn.Ok", "OK"

Isn't this supposed to be Common.OKButton?
Shouldn't there be an ampersand on OK?

Same for others.

FWIW, this property isn't used anywhere.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 345 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDqFnuDuwnBDj-5QQL:-KjDqFnuDuwnBDj-5QQM:b-u40t6t) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L345)):*
> Form-Designer generated code

Will this actually work in designer? Doesn't look like it. If not, is there any way to enable it?
If there is no way to enable it, maybe we should add a comment which explains that while this was originally generated, it must be modified manually in the future.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 348 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDjt8oRdupHxfMSsVs:-KjDjt8oRdupHxfMSsVt:b-evsbyw) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L348)):*
> Erforderliche Designervariable.

Lots of German throughout here...

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 589 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDr9kruYFFFlKHZoj5:-KjDr9ksa0y9xWwC0WgS:b-3qropz) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L589)):*
> The corrected given working area factor.

This is confusing.
How about "The given working area factor, corrected" or simply "The corrected working area factor"?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 609 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDrbhCvPzjh6Q_45fP:-KjDrbhCvPzjh6Q_45fQ:b-byc5ao) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L609)):*
> If no owner given: Center on current screen

I guess this is supposed to indicate centering on current screen is the default behavior? Could be clarified.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 881 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjENu9F0bf9v79WjFX5:-KjENu9F0bf9v79WjFX6:b-thogmg) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L881)):*
> Public show function

It is internal, actually

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-:-KjDlO3Av1N60kPOF0Y3:b-gojk5n)*

JohnThomson:
I think I recall our already having a custom messagebox class...maybe it's only in FLEx? One of the main additions I recall in that one was provision for a Help button...would that be worth doing here, too?

---

Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 134 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEVcnUKtxhMykA6jYE:-KjEVcnUKtxhMykA6jYF:b-tq6bp4) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L134)):*
> ```Smalltalk
> 		}
> 
> 		public static Func<string, string> GetButtonText { get; set; }
> ```

Wants either to go away, or have a comment saying normal L10NSharp localization will be used if you do NOT set it.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 616 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEWGe36izyvFzqZMQQ:-KjEWGe36izyvFzqZMQR:b9qhnwb) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L616)):*
> ```Smalltalk
> 					flexibleMessageBoxForm.Left = screen.Bounds.Left + screen.Bounds.Width / 2 - flexibleMessageBoxForm.Width / 2;
> 					flexibleMessageBoxForm.Top = screen.Bounds.Top + screen.Bounds.Height / 2 - flexibleMessageBoxForm.Height / 2;
> 				}
> ```

And if owner isn't null? Surely we should center on the owner screen? Not leave the position unspecified?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 629 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEWjCF2ogR8jngGJvg:-KjEWjCF2ogR8jngGJvh:bj01rb5) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L629)):*
> ```Smalltalk
> 			{
> 				//First set the bounds for the maximum dialog size
> 				flexibleMessageBoxForm.MaximumSize = new Size(Convert.ToInt32(SystemInformation.WorkingArea.Width * FlexibleMessageBoxForm.GetCorrectedWorkingAreaFactor(MAX_WIDTH_FACTOR)),
> ```

Feels like it should be based on the width of the screen we decide to show it on. 80% of my primary monitor's width would not fit at all on either of the others.
Note that if you do this, you'll need to account for the fact that SetDialogStartPosition figures out the right screen but then assumes the width is already set.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 629 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEfB-RK0yyA3XDWcgW:-KjEfB-RK0yyA3XDWcgX:bgvj857) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L629)):*
> ```Smalltalk
> 			{
> 				//First set the bounds for the maximum dialog size
> 				flexibleMessageBoxForm.MaximumSize = new Size(Convert.ToInt32(SystemInformation.WorkingArea.Width * FlexibleMessageBoxForm.GetCorrectedWorkingAreaFactor(MAX_WIDTH_FACTOR)),
> ```

Feels odd to set the system maximum size of the dialog here. I can understand that this class has a maximum size that IT will ever make the dialog, but why shouldn't the USER be allowed to make it larger and see the whole message if he wants to?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 640 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEXatp3M3LyR2aolV3:-KjEXatp3M3LyR2aolV4:b-jxlklp) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L640)):*
> ```Smalltalk
> 					
> 				//Calculate width for longest text line
> 				const int SCROLLBAR_WIDTH_OFFSET = 15;
> ```

SystemInformation.VerticalScrollBarWidth?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 641 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEXHknSJGhxXIpjq6l:-KjEXHknSJGhxXIpjq6m:b-7t41a4) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L641)):*
> ```Smalltalk
> 				//Calculate width for longest text line
> 				const int SCROLLBAR_WIDTH_OFFSET = 15;
> 				var longestTextRowWidth = stringRows.Max(textForRow => TextRenderer.MeasureText(textForRow, FONT).Width);
> ```

Is this ever different from the width we got (and discarded) above from measuring the whole text?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 643 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEgiwoltSwjxqyPWYS:-KjEgiwoltSwjxqyPWYT:bcva6jg) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L643)):*
> ```Smalltalk
> 				var longestTextRowWidth = stringRows.Max(textForRow => TextRenderer.MeasureText(textForRow, FONT).Width);
> 				var captionWidth = TextRenderer.MeasureText(caption, SystemFonts.CaptionFont).Width;
> 				var textWidth = Math.Max(longestTextRowWidth + SCROLLBAR_WIDTH_OFFSET, captionWidth);
> ```

Technically we don't need the scroll bar width UNLESS we exceed the available vertical space, do we? I could live with just having a little extra margin if we don't need one, though.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 651 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEYGPljbfrd3nyFHrV:-KjEYGPljbfrd3nyFHrW:b-sxu1f2) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L651)):*
> ```Smalltalk
> 				//Set calculated dialog size (if the calculated values exceed the maximums, they were cut by windows forms automatically)
> 				flexibleMessageBoxForm.Size = new Size(textWidth + marginWidth,
> 													   textHeight + marginHeight);
> ```

This makes it wide enough so the lines don't have to wrap and high enough to show it all...ideally.
But, if the width gets clipped, the lines will wrap. Then it will want more height, which might easily be available.
There's an overload of MeasureText, or a similar function, that measures it assuming it will be laid out at a certain maximum width. If you used that in the first measureText call, with the appropriate max width, I think you'd get the extra height needed for wrapping.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 782 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEbJma4I9mzJkHUA7L:-KjEbJma4I9mzJkHUA7M:b-tvjb8e) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L782)):*
> ```Smalltalk
> 						flexibleMessageBoxForm.button3.DialogResult = DialogResult.OK;
> 
> 						flexibleMessageBoxForm.CancelButton = flexibleMessageBoxForm.button3;
> ```

Why is button 3 the cancel button, since we've set its text to OK?

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 827 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEcEqkZuDkjr5zBCVX:-KjEcEqlefnDfsz_fdCw:b8hshfs) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L827)):*
> ```Smalltalk
> 				else if (buttonIndexToFocus == 2)
> 				{
> 					buttonToFocus = this.button2;
> ```

There are cases above (e.g., YesNo) where the two visible buttons are 1 and 3.
If the index is an index among visible buttons, the code here needs adjusting.
If it's relative to all the buttons, you shouldn't be limiting it to the visibleButtonCount above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-:-KjEZOefqxHj1DH3FLfo:bomi04t)*

tombogle:
I think that was in FW. We did poke around a bit looking for anything similar. Libpalaso does already have *another* alternative to the standard MessageBox called SimpleMessageDialog, but it's designed for a totally different purpose. The addition of a Help button option would probably be a nice enhancement. It would presumably mean some additional overloads of Show, so it would certainly be something that could be done in the future that wouldn't be a breaking change.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-:-KjF0gD-v-yCVtmuaun3:bb4o1vn)*

tombogle:
*[DistFiles/Palaso.en.tmx, line 98 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDevn-Oi0L8QZWGCcw:-KjF1VOke1bvCapz6QmG:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/DistFiles/Palaso.en.tmx#L98)):*

I wonder if these should be AbortButton, etc. That would be more consistent with OKButton, etc. where the ampersand is part of the string.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 11 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDe_tWGtuN-NBDW-Ke:-KjF1c51xyyI6xKm0PPv:bx05zam) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L11)):*

Should this be in Widgets? And named BetterMessageBox to follow the pattern of similar classes already in libpalaso?
</blockquote></details>

Widgets are all controls. This is a form.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

andrew-polk:
Reviewed 4 of 4 files at r3.
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[DistFiles/Palaso.en.tmx, line 100 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjF2SgM5CPl-VQ0W3oV:-KjF2SgM5CPl-VQ0W3oW:b-og652r) ([raw file](https://github.com/sillsdev/libpalaso/blob/53a4a448b8165bedf3f6b6012e14f683047da3bc/DistFiles/Palaso.en.tmx#L100)):*
> &

Needs to be &amp ; 

---

*[DistFiles/Palaso.es.tmx, line 42 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjF2mFpzIKwiVCbx-QC:-KjF2mFpzIKwiVCbx-QD:b-6yvggu) ([raw file](https://github.com/sillsdev/libpalaso/blob/53a4a448b8165bedf3f6b6012e14f683047da3bc/DistFiles/Palaso.es.tmx#L42)):*
> &amp;No

Shouldn't have &amp ;

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 96 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDfw4FQEXwDX7QEElN:-KjIwhJ_B3AbajmT2bqO:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L96)):*

>         /// 0.2 means:  The FlexibleMessageBox can be at most half as wide as the working area.
>         /// 1.0 means:  The FlexibleMessageBox can be as wide as the working area.
>         ///
>         /// Default is: 70% of the working area width.

so if .7 is 70%, why is .2 50%? I'm guessing a typo? Same with other comments below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDg9mWdrUZS6Z1pj9E:-KjIwdG-Hkea0Vm2wb4_:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L14)):*

There is a fair amount of trailing whitespace in this file. Let's go ahead and eliminate it since we are starting fresh.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 100 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDgdVbGiW2-lb6yYjD:-KjIwtgIx617RvaZMhTl:b-nz31s5) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L100)):*

Is all caps the standard for public static in libpalaso? Doesn't seem right to me...
</blockquote></details>

I actually re-wrote this to use public properties backed by private fields (named according to Palaso standards). I also pulled the code that enforces the min/max into the property setters, rather than having it call a method to do this every time it is used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 641 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEXHknSJGhxXIpjq6l:-KjJliCWZnlh7eWHkBGJ:b-rhe8gx) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L641)):*

Is this ever different from the width we got (and discarded) above from measuring the whole text?
</blockquote></details>

Seems unlikely. Don't see how it could be.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
Review status: 1 of 7 files reviewed at latest revision, 19 unresolved discussions.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 345 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDqFnuDuwnBDj-5QQL:-KjJ4XehsOGLi8WfLXXe:b-6rwhhf) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L345)):*

> Form-Designer generated code

Will this actually work in designer? Doesn't look like it. If not, is there any way to enable it?
If there is no way to enable it, maybe we should add a comment which explains that while this was originally generated, it must be modified manually in the future.
</blockquote></details>

No. I added a comment.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 348 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDjt8oRdupHxfMSsVs:-KjJ4msVzRmjQPfw8CfK:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L348)):*

> Erforderliche Designervariable.

Lots of German throughout here...
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 589 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDr9kruYFFFlKHZoj5:-KjJ5gfsUo2Mau39WXyD:bhv3tit) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L589)):*

> The corrected given working area factor.

This is confusing.
How about "The given working area factor, corrected" or simply "The corrected working area factor"?
</blockquote></details>

Too late. It's gone :-)

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 609 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDrbhCvPzjh6Q_45fP:-KjJ7dm570V1ndomq7ba:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L609)):*

> If no owner given: Center on current screen

I guess this is supposed to indicate centering on current screen is the default behavior? Could be clarified.
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 134 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEVcnUKtxhMykA6jYE:-KjIzblQisCCkBWhTDbR:b-50loaj) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L134)):*

Wants either to go away, or have a comment saying normal L10NSharp localization will be used if you do NOT set it.
</blockquote></details>

Added necessary comments (this is going to be needed by Paratext if they decide to use it, which Tim said they might).

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 616 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEWGe36izyvFzqZMQQ:-KjJ7geULtpdMkxcNzuQ:b-iobvmi) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L616)):*

And if owner isn't null? Surely we should center on the owner screen? Not leave the position unspecified?
</blockquote></details>

That's the default based on the value set for StartPosition by the Designer. I've improved the comments and method name and moved the check for null outside the method. Hopefully clearer now.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 629 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEWjCF2ogR8jngGJvg:-KjJBdeqxLXdp3B4rPCM:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L629)):*

Feels like it should be based on the width of the screen we decide to show it on. 80% of my primary monitor's width would not fit at all on either of the others.
Note that if you do this, you'll need to account for the fact that SetDialogStartPosition figures out the right screen but then assumes the width is already set.
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 629 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEfB-RK0yyA3XDWcgW:-KjJCLDYb0vpHeeg9lJy:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L629)):*

Feels odd to set the system maximum size of the dialog here. I can understand that this class has a maximum size that IT will ever make the dialog, but why shouldn't the USER be allowed to make it larger and see the whole message if he wants to?
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 640 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEXatp3M3LyR2aolV3:-KjJCX1tErpAfEcbLcbd:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L640)):*

SystemInformation.VerticalScrollBarWidth?
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 643 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEgiwoltSwjxqyPWYS:-KjJCmjWrzz1hLtGvo3s:bly7ahw) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L643)):*

Technically we don't need the scroll bar width UNLESS we exceed the available vertical space, do we? I could live with just having a little extra margin if we don't need one, though.
</blockquote></details>

True. That extra margin does (sort of) help in the (unlikely) case where the user just decreases the vertical height of the message box.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 643 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEgiwoltSwjxqyPWYS:-KjNEP3b8ZXFYbzNaCqX:bvxqqmv) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L643)):*

True. That extra margin does (sort of) help in the (unlikely) case where the user just decreases the vertical height of the message box.
</blockquote></details>

I've pretty much completely overhauled this method to make better use of the MeasureText method, deal with tabs, and handle some of the edge cases the original version was mishandling.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 651 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEYGPljbfrd3nyFHrV:-KjNEiUQhCcvyTSO-SUb:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L651)):*

This makes it wide enough so the lines don't have to wrap and high enough to show it all...ideally.
But, if the width gets clipped, the lines will wrap. Then it will want more height, which might easily be available.
There's an overload of MeasureText, or a similar function, that measures it assuming it will be laid out at a certain maximum width. If you used that in the first measureText call, with the appropriate max width, I think you'd get the extra height needed for wrapping.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[DistFiles/Palaso.en.tmx, line 100 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjF2SgM5CPl-VQ0W3oV:-KjNEqLH-leIDfooBMZO:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/53a4a448b8165bedf3f6b6012e14f683047da3bc/DistFiles/Palaso.en.tmx#L100)):*

> &

Needs to be &amp ; 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[DistFiles/Palaso.es.tmx, line 42 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjF2mFpzIKwiVCbx-QC:-KjNF1fUxi2DRCaTJP4o:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/53a4a448b8165bedf3f6b6012e14f683047da3bc/DistFiles/Palaso.es.tmx#L42)):*

> &amp;No

Shouldn't have &amp ;
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 124 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjDh-eVpR7mtY3o4jvE:-KjNRYqBYV1FCIu_STxB:btwlkjt) ([raw file](https://github.com/sillsdev/libpalaso/blob/20f2d7a2c4450d4ab94749a057bc385936b0fb9c/SIL.Windows.Forms/FlexibleMessageBox.cs#L124)):*

> "FlexibleMessageBoxBtn.Ok", "OK"

Isn't this supposed to be Common.OKButton?
Shouldn't there be an ampersand on OK?

Same for others.

FWIW, this property isn't used anywhere.
</blockquote></details>

This is only needed if some caller uses a non-L10NSharp approach to localization (e.g., Paratext). By using FlexibleMessageBoxBtn instead of Common, it should minimize the chance of creating a conflict with some other ID used for localizing something elsewhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 782 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEbJma4I9mzJkHUA7L:-KjNYFAN-4kCHZhmf_TH:bhrvkgd) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L782)):*

Why is button 3 the cancel button, since we've set its text to OK?
</blockquote></details>

For a simple message box with only an OK button, either Enter or Escape needs to close it. I just tested this and it seems to work properly as it is. I assume that without this, Enter would work but Esc wouldn't.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

andrew-polk:
:lgtm: with the admission that I didn't go through the sizing algorithm in detail.

---

Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[TestApps/SIL.Windows.Forms.TestApp/TestAppForm.cs, line 238 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjNZkfj9kdkfS95WdV2:-KjNZkfj9kdkfS95WdV3:bauh0x) ([raw file](https://github.com/sillsdev/libpalaso/blob/4e84f0e53de603f8c6376ca00f2507f642fbb06a/TestApps/SIL.Windows.Forms.TestApp/TestAppForm.cs#L238)):*
> ```Smalltalk
> 		}
> 
> 		private bool ShowFlexibleMessageBox(int option, string msg, string caption, LinkClickedEventHandler handler)
> ```

This is a fun idea!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-:-KjNajrpEVGMpqCvH6Sk:b-o7s5go)*

JohnThomson:
:lgtm: with one very minor question, probably not important

---

Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 565 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjOJs5CPDlleQC7f52k:-KjOJs5CPDlleQC7f52l:buq8fvx) ([raw file](https://github.com/sillsdev/libpalaso/blob/4e84f0e53de603f8c6376ca00f2507f642fbb06a/SIL.Windows.Forms/FlexibleMessageBox.cs#L565)):*
> ```Smalltalk
> 				flexibleMessageBoxForm.StartPosition = FormStartPosition.Manual;
> 				flexibleMessageBoxForm.Left = screen.Bounds.Left + screen.Bounds.Width / 2 - flexibleMessageBoxForm.Width / 2;
> 				flexibleMessageBoxForm.Top = screen.Bounds.Top + screen.Bounds.Height / 2 - flexibleMessageBoxForm.Height / 2;
> ```

Should we use WorkingArea rather than Bounds? (e.g., WorkingArea doesn't count a taskbar)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-:-KjOPCo1dH_hRBdFfWE8:b-yq5866)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 827 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEcEqkZuDkjr5zBCVX:-KjOaRUaUGwDKjzMdOL9:b81hhv6) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L827)):*

There are cases above (e.g., YesNo) where the two visible buttons are 1 and 3.
If the index is an index among visible buttons, the code here needs adjusting.
If it's relative to all the buttons, you shouldn't be limiting it to the visibleButtonCount above.
</blockquote></details>

I thinkyou must have mis-read the code. He always shows the highest-numbered buttons available. Button1 is only shown when one of the three-button options is chosen. However, you're absolutely correct that the logic is wrong. If you are showing only 1 button, it will never get focused. And if you are showing 2 buttons, you'll never get the desired button focused.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 827 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjEcEqkZuDkjr5zBCVX:-KjOfZjwJ8DHjLYo1Y8S:b-bq68p9) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L827)):*

I thinkyou must have mis-read the code. He always shows the highest-numbered buttons available. Button1 is only shown when one of the three-button options is chosen. However, you're absolutely correct that the logic is wrong. If you are showing only 1 button, it will never get focused. And if you are showing 2 buttons, you'll never get the desired button focused.
</blockquote></details>

I was too fast to speak. If you only have one button it always gets the focus, so even though the logic is wrong (and confusing!), it doesn't matter. But the two-button case is a problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 565 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjOJs5CPDlleQC7f52k:-KjPAY6LNoRrlh8JXTI6:bkm2wzf) ([raw file](https://github.com/sillsdev/libpalaso/blob/4e84f0e53de603f8c6376ca00f2507f642fbb06a/SIL.Windows.Forms/FlexibleMessageBox.cs#L565)):*

Should we use WorkingArea rather than Bounds? (e.g., WorkingArea doesn't count a taskbar)
</blockquote></details>

I suppose ideally we should try to emulate MessageBox. It probably doesn't matter a whole lot, but WorkingArea does feel like it would be more sensible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

tombogle:
Review status: 5 of 7 files reviewed at latest revision, 12 unresolved discussions.

---

*[SIL.Windows.Forms/FlexibleMessageBox.cs, line 881 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/494#-KjENu9F0bf9v79WjFX5:-KjPBVbwiTl43KWwPG8R:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1ea25564ea515d4f9a3c8309ded1a2697d264b5/SIL.Windows.Forms/FlexibleMessageBox.cs#L881)):*

> Public show function

It is internal, actually
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

andrew-polk:
Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/494)*

PR #503, opened 2017-05-09
Use NUnitAddin instead of NUnitTeamCity task on TC

ermshiperete:
The NUnitTeamCity task has issues when run with mono-4.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/503)

### DIFF
--- a/build/Palaso.proj
+++ b/build/Palaso.proj
@@ -104,12 +104,19 @@
 	<Target Name="RunNUnitTC" Condition="'$(teamcity_version)' != ''">
 		<ItemGroup>
 			<TestAssemblies Include="$(RootDir)/output/$(Configuration)/*.Tests.dll"/>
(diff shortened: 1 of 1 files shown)


andrew-polk:
:lgtm: 
It strikes me that the TC and non-TC tasks are so similar now, it might be worth extracting part of it to a common task, but that is certainly not a requirement.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/503#-:-Kjh7_-ZKMz6s1xLyvI6:b-htqgo8)*

andrew-polk:
If this works, we should do it in L10NSharp (and Pathway?), too.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/503#-:-Kjh8oEkqbSdzv5xSg5C:b-15yvt4)*

PR #513, opened 2017-05-11
Fix unit test on (single core?) machines by waiting for all image indices to load

hatton:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/513)


hatton:
Andrew reviewed on flowdock; merging to unbreak linux build.

PR #516, opened 2017-05-12
Removed some compiler warnings caused by code not being CLS compliant.

mikelothers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/516)

### DIFF
--- a/SIL.Core/Annotations/Annotation.cs
+++ b/SIL.Core/Annotations/Annotation.cs
@@ -6,7 +6,7 @@ namespace SIL.Annotations
 {
 	public class Annotatable : IAnnotatable, ICloneable<Annotatable>, IEquatable<Annotatable>
 	{
-		protected Annotation _annotation;
+		private Annotation _annotation;
 
 		[XmlAttribute("starred")]
 		public bool IsStarred
@@ -33,7 +33,13 @@ namespace SIL.Annotations
 			}
 		}
 
-		public virtual Annotatable Clone()
+		protected Annotation Annotation
+		{
+			get { return _annotation; }
+			set { _annotation = value; }
+		}
+
+	    public virtual Annotatable Clone()
 		{
 			var clone = new Annotatable();
 			clone._annotation = _annotation == null ? null : _annotation.Clone();
--- a/SIL.Core/Annotations/Annotation.cs
+++ b/SIL.Core/Annotations/Annotation.cs
@@ -6,43 +6,37 @@ namespace SIL.Annotations
 {
 	public class Annotatable : IAnnotatable, ICloneable<Annotatable>, IEquatable<Annotatable>
 	{
-		private Annotation _annotation;
-
 		[XmlAttribute("starred")]
 		public bool IsStarred
 		{
 			get
 			{
-				if (_annotation == null)
+				if (Annotation == null)
 				{
 					return false; // don't bother making one yet
 				}
-				return _annotation.IsOn;
+				return Annotation.IsOn;
 			}
 			set
 			{
 				if (!value)
 				{
-					_annotation = null; //free it up
+					Annotation = null; //free it up
 				}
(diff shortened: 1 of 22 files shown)


hatton:
Review status: 0 of 22 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Core/Annotations/Annotation.cs, line 41 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/516#-Kjy9NLOOZr8yrHA4qfL:-Kjy9NLOOZr8yrHA4qfM:bfhvawy) ([raw file](https://github.com/sillsdev/libpalaso/blob/dbcee7133e7a801e1c1fa80fd2ce2609a88f02de/SIL.Core/Annotations/Annotation.cs#L41)):*
> ```Smalltalk
> 			set { _annotation = value; }
> 		}
> ```

I think simply

    protected Annotation Annotation;

would be simpler and thus preferred in all of these situations, rather than using a backing field or even ````{get; set;}````

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/516)*

andrew-polk:
Reviewed 22 of 22 files at r1.

---

*[SIL.Core/Annotations/Annotation.cs, line 41 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/516#-Kjy9NLOOZr8yrHA4qfL:-KjyD70eosGsOURHj_NS:b-nzec79) ([raw file](https://github.com/sillsdev/libpalaso/blob/dbcee7133e7a801e1c1fa80fd2ce2609a88f02de/SIL.Core/Annotations/Annotation.cs#L41)):*

I think simply

    protected Annotation Annotation;

would be simpler and thus preferred in all of these situations, rather than using a backing field or even ````{get; set;}````
</blockquote></details>

I can't remember the arguments now, but in the past Tom has convinced me 
```
protected Annotation Annotation {get; set;}
```
is preferred to
```
protected Annotation Annotation;
```
But I agree it would be better to get rid of the backing fields.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/516)*

hatton:

---

*[SIL.Core/Annotations/Annotation.cs, line 41 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/516#-Kjy9NLOOZr8yrHA4qfL:-KjyE5LXY9BL3ixdYKS7:b-jgf3dv) ([raw file](https://github.com/sillsdev/libpalaso/blob/dbcee7133e7a801e1c1fa80fd2ce2609a88f02de/SIL.Core/Annotations/Annotation.cs#L41)):*

I can't remember the arguments now, but in the past Tom has convinced me 
```
protected Annotation Annotation {get; set;}
```
is preferred to
```
protected Annotation Annotation;
```
But I agree it would be better to get rid of the backing fields.
</blockquote></details>

The arguments that I know of are not compelling; they fall into the category of the way we programmers start to think in what-if's that get us divorced from reality. In short, that API's in libpalaso can't change. 
 Spelled out: 

1) someone subclasses the class, in an external project
2) someone modified the original class to add code to the get or set
3) the subclassing project gets the latest palaso, build fails, and...
4) the world falls apart.

or 

1) someone subclasses the class, in an external project, and needs a way to change the behavior of get/set.
2) they find that they would have to go back into libpalaso to make this possible
3) the world falls apart.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/516)*

andrew-polk:
:lgtm:

---

Reviewed 10 of 10 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/516#-:-KkBQxozNoOYB6riuuEf:bnfp4nl)*

PR #523, opened 2017-05-19
Some refactorings in XmlUtils

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/523)

### DIFF
--- a/SIL.Core.Tests/Xml/XmlUtilsTests.cs
+++ b/SIL.Core.Tests/Xml/XmlUtilsTests.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Text;
+using System.Threading;
 using System.Xml;
 using System.Xml.Linq;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


FoolRunning:
:lgtm: for the most part. See my comment. Not sure it's worth addressing.

---

Reviewed 2 of 2 files at r1.

---

*[SIL.Core/Xml/XmlUtils.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/523#-KkpVgbXjZBvMGUjTyZk:-KkpVgbXjZBvMGUjTyZl:bkbmrjj) ([raw file](https://github.com/sillsdev/libpalaso/blob/1f4c8d9d817caeeb9a3b7a8118b31f66e4156f1d/SIL.Core/Xml/XmlUtils.cs#L251)):*
> ```Smalltalk
> 		public static string MakeStringFromList(List<int> vals)
> 		{
> 			return MakeIntegerListValue(vals.ToArray());
> ```

This is the only thing that jumped out at me. This method is now less efficient than it was because of the array copy (especially for large lists). One option would be to create a private version that takes an **IEnumberable\<int\>**. However, since I'm not certain who uses this method, it may not be worth doing the work for the (probably small) gain of performance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/523#-:-KkpXkIvgWS2xu6Mbh1R:b-asp0dw)*

ermshiperete:

---

*[SIL.Core/Xml/XmlUtils.cs, line 251 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/523#-KkpVgbXjZBvMGUjTyZk:-KlJbsr4u-clJLDehlGL:b-q1wb3l) ([raw file](https://github.com/sillsdev/libpalaso/blob/1f4c8d9d817caeeb9a3b7a8118b31f66e4156f1d/SIL.Core/Xml/XmlUtils.cs#L251)):*

This is the only thing that jumped out at me. This method is now less efficient than it was because of the array copy (especially for large lists). One option would be to create a private version that takes an **IEnumberable\<int\>**. However, since I'm not certain who uses this method, it may not be worth doing the work for the (probably small) gain of performance.
</blockquote></details>

Good idea. Updated the PR. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/523)*

ermshiperete:
All builds on `master` branch succeeded.

FoolRunning:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/523#-:-KlOLZgOJKHvitlq_PiP:bnfp4nl)*

PR #526, opened 2017-05-22
Moved ReflectionHelper to SIL.Core so it doesn't depend on System.Windows.Forms

FoolRunning:
Moved **ReflectionHelper** to **SIL.Core** so it doesn't depend on **System.Windows.Forms** (and it makes more sense there).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/526)

### DIFF
--- a/SIL.Core/Reflection/ReflectionHelper.cs
+++ b/SIL.Core/Reflection/ReflectionHelper.cs
@@ -1,10 +1,8 @@
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Reflection;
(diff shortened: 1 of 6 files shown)


andrew-polk:
:lgtm: as long as no one objects to another breaking change. I'm okay with it, and now is the time to do it.

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/526#-:-Kklze3QkSPyvy2ht_r-:b-4n53wl)*

PR #528, opened 2017-05-23
Maintain backwards compatibility in SILAboutBox

gmartin7:
* handle parsing failures differently for Debug
   or Release builds
* uses HtmlAgilityPack to load AboutBox html and
   validate/manipulate it.
* do more returning of original html on error
   to maintain backwards compatibility
* do better job of ensuring correct meta tag

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/528)

### DIFF
--- a/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs
+++ b/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs
@@ -1,13 +1,58 @@
-﻿using NUnit.Framework;
+﻿using System;
+using NUnit.Framework;
 using SIL.Windows.Forms.Miscellaneous;
 
 namespace SIL.Windows.Forms.Tests.Miscellaneous
 {
+	/// <summary>
+	/// The SILAboutBox class has different behavior depending on whether it is run
+	/// in Debug or Release mode, therefore different test results depending on which version is running.
+	/// </summary>
 	[TestFixture]
 	public class SILAboutBoxTests
 	{
+
 		[Test]
-		public void ValidateHtml_ValidHtmlNotOverriden()
+		public void ValidateHtml_MissingBodyTagHandled()
+		{
+			const string htmlInput =
+				@"<html>
+					<meta charset='UTF-16' />
+					<some>stuff</some>
+				</html>";
+#if DEBUG
+			var ex = Assert.Throws<ApplicationException>(() => SILAboutBox.ValidateHtml(htmlInput));
+			Assert.That(ex.Message, Is.EqualTo("Html has no head or body and needs a charset meta tag."));
+#else
+			var result = SILAboutBox.ValidateHtml(htmlInput);
+			Assert.That(result, Is.EqualTo(htmlInput));
+#endif
+		}
+
+		[Test]
+		public void ValidateHtml_InvalidHtmlHandled()
+		{
+			const string htmlInput =
+				@"<html>
+					ad>
+					</head>
+					<body>
+						<some>stuff</some>
+					</body>
+				</html>";
+#if DEBUG
+			var ex = Assert.Throws<ApplicationException>(() => SILAboutBox.ValidateHtml(htmlInput));
+			Assert.That(ex.Message, Is.EqualTo("Start tag <head> was not found at line 3"));
+#else
+			var result = SILAboutBox.ValidateHtml(htmlInput);
+			Assert.That(result, Is.EqualTo(htmlInput));
+#endif
+		}
+
+		// These tests should behave the same whether in Debug or Release mode
+
+		[Test]
+		public void ValidateHtml_ValidHtml_CharsetNotOverriden()
 		{
 			const string htmlInput =
 				@"<html>
@@ -23,31 +68,33 @@ namespace SIL.Windows.Forms.Tests.Miscellaneous
 		}
 
 		[Test]
-		public void ValidateHtml_InvalidHtml_ReturnsMinimalHtml()
+		public void ValidateHtml_ValidHtmlNoHead_GetsHeadAndMeta()
 		{
 			const string htmlInput =
 				@"<html>
-					ad>
+					<body>
+						<some>stuff</some>
+					</body>
+				</html>";
+			var result = SILAboutBox.ValidateHtml(htmlInput);
+			Assert.That(result, Is.StringContaining("<head><meta charset=\"UTF-8\"></head>"));
+		}
+
+		[Test]
+		public void ValidateHtml_ValidHtmlMetaTagHasNoCharsetAttribute_GetsOne()
+		{
+			const string htmlInput =
+				@"<html>
+					<head>
+						<meta content='text/html' />
 					</head>
 					<body>
 						<some>stuff</some>
 					</body>
 				</html>";
 			var result = SILAboutBox.ValidateHtml(htmlInput);
-			Assert.That(result, Is.EqualTo(SILAboutBox.MinimalHtmlContents));
-		}
-
-		[Test]
-		public void ValidateHtml_ValidHtmlNoMetaTag_GetsOne()
-		{
-			const string htmlInput =
-				@"<html>
-					<body>
-						<some>stuff</some>
-					</body>
-				</html>";
-			var result = SILAboutBox.ValidateHtml(htmlInput);
-			Assert.That(result, Is.StringContaining("<head><meta charset=\"UTF-8\" /></head>"));
+			var trimmedResult = result.Replace(Environment.NewLine, string.Empty).Replace("\t", string.Empty);
+			Assert.That(trimmedResult, Is.StringContaining("<head><meta content='text/html' charset=\"UTF-8\"></head>"));
 		}
 	}
 }
--- a/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs
+++ b/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs
@@ -9,8 +9,10 @@ using System.Reflection;
 using System.Windows.Forms;
 using System.Xml;
 using L10NSharp;
+using HAP = HtmlAgilityPack;
 using SIL.Acknowledgements;
 using SIL.IO;
+using SIL.Reporting;
 
 namespace SIL.Windows.Forms.Miscellaneous
 {
@@ -226,52 +228,64 @@ namespace SIL.Windows.Forms.Miscellaneous
 			}
 		}
 
-		internal const string MinimalHtmlContents = "<html><head><meta charset='UTF-8' /></head><body></body><ul>" +
-								DependencyMarker + "</ul></html>";
-
 		/// <summary>
-		/// Make sure the About Box html is valid by loading it into an XmlDocument.
-		/// Also ensure that it uses Unicode, since even common things like copyright symbols
+		/// Make sure the About Box html is valid by loading it into an HtmlDocument.
+		/// Also try to ensure that it uses Unicode, since even common things like copyright symbols
 		/// won't be rendered correctly otherwise.
+		///
+		/// Behavior is different for Debug and Release versions. Release versions will Log an event if
+		/// there are parsing errors, but will return the original parameter in order to maintain backwards compatibility.
+		/// Debug versions will throw an exception so that programmers will be encouraged to update their html.
+		/// Tests will reflect this.
+		///
 		/// Internal for testing
 		/// </summary>
 		internal static string ValidateHtml(string htmlString)
 		{
-			XmlDocument dom;
-			XmlNode headNode;
-			try
+			// Bloom actually had missing html, head and body tags at one point.
+			var dom = new HAP.HtmlDocument();
+			dom.LoadHtml(htmlString);
+			if (dom.ParseErrors.Any())
 			{
-				dom = new XmlDocument {InnerXml = htmlString};
-				var charsetNode = dom.SelectSingleNode("//head/meta[@charset]");
-				if (charsetNode != null)
+				var err = dom.ParseErrors.First();
+				HandleParsingError(string.Format("{0} at line {1}", err.Reason, err.Line));
+				return htmlString; // maintain backwards compatibility even if there are parsing errors
+			}
+			var charsetNode = dom.DocumentNode.SelectSingleNode("//head/meta[@charset]");
+			if (charsetNode != null)
+			{
+				return htmlString; // don't overwrite any existing charset attribute
+			}
+			var headNode = dom.DocumentNode.SelectSingleNode("//head");
+			if (headNode == null)
+			{
+				var bodyNode = dom.DocumentNode.SelectSingleNode("//body");
+				if (bodyNode == null)
 				{
-					return htmlString; // don't overwrite any existing charset attribute
+					// What!? Someone created an AboutBox html file with no body element?!
+					HandleParsingError("Html has no head or body and needs a charset meta tag.");
+					return htmlString;
 				}
-				headNode = dom.SelectSingleNode("//head");
-				if (headNode == null)
-				{
-					var bodyNode = dom.SelectSingleNode("//body");
-					if (bodyNode == null)
-					{
-						// What!? Someone created an AboutBox html file with no body element?!
-						// Create a minimal file
-						return MinimalHtmlContents;
-					}
-					headNode = dom.CreateElement("head");
-					bodyNode.ParentNode.InsertBefore(headNode, bodyNode);
-				}
-				var metaNode = dom.CreateElement("meta");
-				var charsetAttr = dom.CreateAttribute("charset");
-				charsetAttr.Value = "UTF-8";
-				metaNode.SetAttributeNode(charsetAttr);
+				headNode = dom.CreateElement("head");
+				bodyNode.ParentNode.InsertBefore(headNode, bodyNode);
+			}
+			var metaNode = dom.DocumentNode.SelectSingleNode("//head/meta");
+			if (metaNode == null) // might be a meta tag with no charset attribute
+			{
+				metaNode = dom.CreateElement("meta");
 				headNode.AppendChild(metaNode);
-				return dom.InnerXml;
 			}
-			catch (XmlException ex)
-			{
-				Debug.WriteLine("Loading the AboutBox html threw an exception: " + ex.Message);
-				return MinimalHtmlContents;
-			}
+			metaNode.SetAttributeValue("charset", "UTF-8");
+			return dom.DocumentNode.OuterHtml;
+		}
+
+		private static void HandleParsingError(string message)
+		{
+#if DEBUG
+			throw new ApplicationException(message);
+#else
+			Logger.WriteEvent("Html for AboutBox generated parse errors. " + message);
+#endif
 		}
 
 		protected override void OnClosed(EventArgs e)
--- a/SIL.Windows.Forms/Properties/AssemblyInfo.cs
+++ b/SIL.Windows.Forms/Properties/AssemblyInfo.cs
@@ -43,3 +43,6 @@ using SIL.Acknowledgements;
 	Location = "./taglib-sharp.dll", LicenseUrl = "https://opensource.org/licenses/LGPL-2.1")]
 [assembly: Acknowledgement("gtk-sharp", Name = "gtk-sharp", Url = "http://www.mono-project.com/docs/gui/gtksharp/",
 	Location = "./gtk-sharp.dll", LicenseUrl = "https://opensource.org/licenses/MIT")]
+[assembly: Acknowledgement("HtmlAgilityPack", Name = "HtmlAgilityPack",
+	Url = "https://www.nuget.org/packages/HtmlAgilityPack/", LicenseUrl = "https://opensource.org/licenses/MS-PL",
+	Location = "./HtmlAgilityPack.dll")]
--- a/SIL.Windows.Forms/SIL.Windows.Forms.csproj
+++ b/SIL.Windows.Forms/SIL.Windows.Forms.csproj
@@ -381,6 +381,10 @@
     <Reference Include="Mono.Posix" />
   </ItemGroup>
   <ItemGroup>
+    <Reference Include="HtmlAgilityPack, Version=1.4.9.5, Culture=neutral, PublicKeyToken=bd319b19eaf3b43a, processorArchitecture=MSIL">
+      <HintPath>..\packages\HtmlAgilityPack.1.4.9.5\lib\Net45\HtmlAgilityPack.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="MarkdownDeep, Version=1.5.4615.26275, Culture=neutral, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\packages\MarkdownDeep.NET.1.5\lib\.NetFramework 3.5\MarkdownDeep.dll</HintPath>
--- a/SIL.Windows.Forms/packages.config
+++ b/SIL.Windows.Forms/packages.config
@@ -1,4 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="HtmlAgilityPack" version="1.4.9.5" targetFramework="net461" />
   <package id="MarkdownDeep.NET" version="1.5" targetFramework="net40-Client" />
 </packages>


andrew-polk:
Reviewed 1 of 1 files at r1.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 244 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqhrGm7j4WkZfFOouF:-KkqhrGm7j4WkZfFOouG:b-6zku07) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L244)):*
> This will throw if the htmlString is not valid XML

Actually, this feels wrong. There is valid HTML which isn't valid XML. For example: 
```
<html>
<head><title>foo</title></head>
<body>
<img src="a" alt="alt">
</body>
</html>
```
is invalid XML because the img tag is not closed.

Can we find a way to do this validation for HTML rather than XML?

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 245 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqhQuhazuayllpZraT:-KkqhQuhazuayllpZraU:bdf7448) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L245)):*
> Changes that may need to be made could include

Worth mentioning what broke Bloom? Namely no head and body tag?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 246 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqjE4Bs_mGP5B3rlP2:-KkqjE4Bs_mGP5B3rlP3:bncx485) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L246)):*
> &nbsp; -> &#160;

Shouldn't have to do this for HTML. This is related to my comment about XML vs HTML above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 246 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqjE4Bs_mGP5B3rlP2:-KkqjT8rtvSgU8qmQQ5n:b-ua19ya) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L246)):*
That should have read:

> & nbsp; -> &# 160

Shouldn't have to do this for HTML. This is related to my comment about XML vs HTML above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 244 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqhrGm7j4WkZfFOouF:-KkqoscNiPpQaZ9uSz8R:bbeuzzy) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L244)):*

> This will throw if the htmlString is not valid XML

Actually, this feels wrong. There is valid HTML which isn't valid XML. For example: 
```
<html>
<head><title>foo</title></head>
<body>
<img src="a" alt="alt">
</body>
</html>
```
is invalid XML because the img tag is not closed.

Can we find a way to do this validation for HTML rather than XML?
</blockquote></details>

Do you suppose it's worth trying to use libtidy?

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 245 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqhQuhazuayllpZraT:-KkqlPQAvo5oXSgzVz4b:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L245)):*

> Changes that may need to be made could include

Worth mentioning what broke Bloom? Namely no head and body tag?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 244 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqhrGm7j4WkZfFOouF:-KkqpnTSlfqNBQ0o0I3f:b-bmnmdt) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L244)):*
> Do you suppose it's worth trying to use libtidy?

Probably not.

Taking a step back...
We don't necessarily care that we have valid HTML. We have been living with it to this point with issue. This all really came about because the process by which we were trying to ensure we had the charset set as UTF-8 *happened* to also ensure we had valid XML (sorry I didn't catch the HTML vs XML issue when I first reviewed this. So how can we ensure the meta tag with charset="UTF-8" without having to have valid XML?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 244 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqhrGm7j4WkZfFOouF:-KkqqOenTz_lYbTPtvtN:bl35s5y) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L244)):*

> Do you suppose it's worth trying to use libtidy?

Probably not.

Taking a step back...
We don't necessarily care that we have valid HTML. We have been living with it to this point with issue. This all really came about because the process by which we were trying to ensure we had the charset set as UTF-8 *happened* to also ensure we had valid XML (sorry I didn't catch the HTML vs XML issue when I first reviewed this. So how can we ensure the meta tag with charset="UTF-8" without having to have valid XML?
</blockquote></details>

to this point with issue -> to this point without issue

I really wish you could edit these comments...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 246 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkqjE4Bs_mGP5B3rlP2:-KkrHx9RMoK_QDrqqBcB:b-s7pmwe) ([raw file](https://github.com/sillsdev/libpalaso/blob/543748824339991bfd904c6cd3edce389fd135fa/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L246)):*

That should have read:

> & nbsp; -> &# 160

Shouldn't have to do this for HTML. This is related to my comment about XML vs HTML above.
</blockquote></details>

HtmlAgilityPack worked great!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:
Reviewed 5 of 5 files at r2.

---

*[SIL.Windows.Forms/packages.config, line 3 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuJYybhxYIEe2r65rw:-KkuJYybhxYIEe2r65rx:b-mhcqjq) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/packages.config#L3)):*
> ```XML
> <?xml version="1.0" encoding="utf-8"?>
> <packages>
>   <package id="HtmlAgilityPack" version="1.4.9.5" targetFramework="net461" />
> ```

I think this means that all projects which use the SILAboutBox now need to distribute this dll. It is pretty small (130K), so that might be okay. But we may need to run it by others and/or see if anyone thinks of a better way to do what we are ultimately trying to accomplish (asserting that the charset is defined in a meta tag).

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 10 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuHEHkrfu4aJTF8uXh:-KkuHEHkrfu4aJTF8uXi:b2je22n) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L10)):*
> ```Smalltalk
> 	{
> 		[Test]
> 		public void ValidateHtml_MissingBodyTagReturnsMinimalHtml()
> ```

Is there a reason we don't want the original htmlInput here, too?
If someone already has aboutBox html they have been using which doesn't have a body tag, seems like we should treat it the same way as the invalid html path (i.e. maintain backward compatibility with a Debug failure or exception).

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-KkuGWRdec1ZvF3Hs4Aa:b-p03bsy) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*
> but a Debug.Fail makes sure that programmers become aware of the problem.

Does this mean the unit test will Debug.Fail every time it is run in Debug mode?
Doesn't seem like we can live with that.

Best option I can think of at the moment is two separate Asserts, one for #if DEBUG, one for the #else. That's probably not a bad idea. It documents that actually *is* the behavior we want.
So then, perhaps we actually want #if DEBUG in the actual code to throw an ApplicationException.

Open to other ideas.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 251 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuIAyJoI2muu9_mG63:-KkuIAyJoI2muu9_mG64:b-vtjrk8) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L251)):*
> ```Smalltalk
> 				{
> 					var err = dom.ParseErrors.First();
> 					throw new ApplicationException(string.Format("{0} at {1}", err.Reason, err.Line));
> ```

What kinds of things will cause this exception? Is this another case where we should just return the original string for backward compatibility?

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 266 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuIKZaK0n5JAYd3zMK:-KkuIKZaK0n5JAYd3zML:b-39emrw) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L266)):*
> ```Smalltalk
> 						// What!? Someone created an AboutBox html file with no body element?!
> 						// Create a minimal file
> 						return MinimalHtmlContents;
> ```

As I commented on the test, I think this should probably return the original string.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:

---

*[SIL.Windows.Forms/packages.config, line 3 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuJYybhxYIEe2r65rw:-KkvDjrABVxLFI2isIXM:bjolp0f) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/packages.config#L3)):*

I think this means that all projects which use the SILAboutBox now need to distribute this dll. It is pretty small (130K), so that might be okay. But we may need to run it by others and/or see if anyone thinks of a better way to do what we are ultimately trying to accomplish (asserting that the charset is defined in a meta tag).
</blockquote></details>

Perhaps when we get it ready to the best of our ability we can call for other opinions on Common flow.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 10 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuHEHkrfu4aJTF8uXh:-KkvAjtpw5V8Fz_Ksg8g:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L10)):*

Is there a reason we don't want the original htmlInput here, too?
If someone already has aboutBox html they have been using which doesn't have a body tag, seems like we should treat it the same way as the invalid html path (i.e. maintain backward compatibility with a Debug failure or exception).
</blockquote></details>

Done.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-KkvDJqzj-CatykQJ52g:bkvx193) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*

> but a Debug.Fail makes sure that programmers become aware of the problem.

Does this mean the unit test will Debug.Fail every time it is run in Debug mode?
Doesn't seem like we can live with that.

Best option I can think of at the moment is two separate Asserts, one for #if DEBUG, one for the #else. That's probably not a bad idea. It documents that actually *is* the behavior we want.
So then, perhaps we actually want #if DEBUG in the actual code to throw an ApplicationException.

Open to other ideas.
</blockquote></details>

The trick is to make sure that programmers see parse errors and such, but tests pass in both Debug and Release CI situations.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 251 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuIAyJoI2muu9_mG63:-KkvDgVo9f3bPIRFtzfl:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L251)):*

What kinds of things will cause this exception? Is this another case where we should just return the original string for backward compatibility?
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 266 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuIKZaK0n5JAYd3zMK:-KkvA0PTV7rKGXdLSej8:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L266)):*

As I commented on the test, I think this should probably return the original string.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:
Review status: 3 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-KkvZmarJNItoiSIFk0Q:b-96k153) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*

The trick is to make sure that programmers see parse errors and such, but tests pass in both Debug and Release CI situations.
</blockquote></details>

Was/am having trouble with flowdock and hadn't pushed the latest. What do you think now?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/packages.config, line 3 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuJYybhxYIEe2r65rw:-KkvbyMsMW_MM8LqXdw-:b-wnasc3) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms/packages.config#L3)):*

Perhaps when we get it ready to the best of our ability we can call for other opinions on Common flow.
</blockquote></details>

As long as we aren't spending time getting this way perfected when it will be replaced with something else.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-KkvWkYQLgitVMrZHxZ5:b-mwjkd7) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*

Was/am having trouble with flowdock and hadn't pushed the latest. What do you think now?
</blockquote></details>

I still don't see code which addresses this. Seems like the test needs to have #if DEBUG ... #else ...

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 251 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkvbLW5-0yV68gYcUnF:-KkvbLW5-0yV68gYcUnG:b-efpgil) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1d6387c61ed106fb96c793e53664244d3afdd99/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L251)):*
> ```Smalltalk
> 				if (dom.ParseErrors.Count() > 0)
> 				{
> #if DEBUG
> ```

Assuming this is even the direction we really want to go (I still think we need to pursue another approach to accomplishing the main goal until we decide it isn't feasible), the three paths which fallback to the original html should all be handled the same way (with the code as common as possible): alert devs, let it pass for prod (logging is okay).
Currently the three paths do three different things.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 271 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkvaZIGwnG2OwF8SYW5:-KkvaZIGwnG2OwF8SYW6:b99tzyg) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1d6387c61ed106fb96c793e53664244d3afdd99/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L271)):*
> but... will it work?

Sure. This was/is the case for Bloom, no?
The point is we don't want to break something a project is already using. If they are using it, it works. : )

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:
I can't connect to Flowdock or I'd post on Commons, requesting wider feedback.

gmartin7:
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-Kkw11T9J8XpHxPYLtPj:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*

I still don't see code which addresses this. Seems like the test needs to have #if DEBUG ... #else ...
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 251 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkvbLW5-0yV68gYcUnF:-Kkw15bG_SOnU6kFHpHF:b-u6dsur) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1d6387c61ed106fb96c793e53664244d3afdd99/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L251)):*

Assuming this is even the direction we really want to go (I still think we need to pursue another approach to accomplishing the main goal until we decide it isn't feasible), the three paths which fallback to the original html should all be handled the same way (with the code as common as possible): alert devs, let it pass for prod (logging is okay).
Currently the three paths do three different things.
</blockquote></details>

Treated the 2 error conditions the same, the third case that returns the original wasn't an error.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 271 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkvaZIGwnG2OwF8SYW5:-Kkw1HmAX1elEXWSvE-M:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1d6387c61ed106fb96c793e53664244d3afdd99/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L271)):*

> but... will it work?

Sure. This was/is the case for Bloom, no?
The point is we don't want to break something a project is already using. If they are using it, it works. : )
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:
Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-KkwEu0rJfR_ZzV1hiUj:b-p7d7sd) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*

Done.
</blockquote></details>

I would have tried to keep one test per scenario and just had the asserts wrapped in #if DEBUG... #else. I think you can accomplish this with Assert.Throws().
This has the advantage of less duplicated code and ensuring that if someone changes one test, they are changing both.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 251 at r3](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkvbLW5-0yV68gYcUnF:-KkwERZ_by_p5aYum5et:b-vejgyh) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1d6387c61ed106fb96c793e53664244d3afdd99/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L251)):*

Treated the 2 error conditions the same, the third case that returns the original wasn't an error.
</blockquote></details>

Yeah, I was reading it wrong. Thanks for the enhancement.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 243 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkwFWW8HDxEFuLi7u4q:-KkwFWW8HDxEFuLi7u4r:b-egel3p) ([raw file](https://github.com/sillsdev/libpalaso/blob/e041a0d0f2be70004554ab8c6333bbcba042cea1/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L243)):*
> ValidateHtml

I'm not sure this is a good name anymore. Aren't we actually just trying to ensure the charset is set?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:
Review status: 4 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs, line 42 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkuGWRdec1ZvF3Hs4A_:-KkwVWU-GDupcDvlR-6I:bff7pfj) ([raw file](https://github.com/sillsdev/libpalaso/blob/3771edb23e92c995fc47e0594d21edb6bb5ff4c1/SIL.Windows.Forms.Tests/Miscellaneous/SILAboutBoxTests.cs#L42)):*

I would have tried to keep one test per scenario and just had the asserts wrapped in #if DEBUG... #else. I think you can accomplish this with Assert.Throws().
This has the advantage of less duplicated code and ensuring that if someone changes one test, they are changing both.
</blockquote></details>

Unfortunately had to call the SUT differently in each case (DEBUG or RELEASE), but close enough not to cause problems, I think.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 243 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/528#-KkwFWW8HDxEFuLi7u4q:-KkwMY8A5TC7fVMZFmGM:b-rmnpxy) ([raw file](https://github.com/sillsdev/libpalaso/blob/e041a0d0f2be70004554ab8c6333bbcba042cea1/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L243)):*

> ValidateHtml

I'm not sure this is a good name anymore. Aren't we actually just trying to ensure the charset is set?
</blockquote></details>

Depends on your build, I guess. In Debug it throws an exception and tells you about your html parse errors, so I guess that's validation. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

andrew-polk:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/528)*

gmartin7:
Closing pull request since we've decided to rip out html validation and just put in a Debug.Assert to let programmers know they need a charset='UTF-8' in their html.

PR #530, opened 2017-05-25
BL-4662 simplify AboutBox html verification

gmartin7:
* if the SILAboutBox html uses the new
   Acknowledgements Provider marker, we assert that
   the html includes at a minimum a unicode charset attribute.
* redo Acknowledgements part of AboutBox, so links
   to CSS might still work.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/530)


andrew-polk:
Reviewed 3 of 3 files at r1.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 220 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl09Sw9LNXIIc7AAuDG:-Kl09SwAFdoP4Bl4ar4Y:b-udumfe) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L220)):*
> " charset="

is &lt;meta charset ='UTF8'&gt; valid? (with a space after "charset")
If yes, we should change this to just "charset" or a regex.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 239 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Ab06A2yVzJxyIPqc:-Kl0Ab06A2yVzJxyIPqd:b-5pddqy) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L239)):*
> if that doesn't work

How likely is this? What is the recourse for a project for which this happens yet they want to inject the dependency attribution?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

gmartin7:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 220 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl09Sw9LNXIIc7AAuDG:-Kl0D7k8PgKxnpCRz5JO:b569pid) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L220)):*

> " charset="

is &lt;meta charset ='UTF8'&gt; valid? (with a space after "charset")
If yes, we should change this to just "charset" or a regex.
</blockquote></details>

I'll change it to just " charset", since a charset attribute would have to have a space before it.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 239 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Ab06A2yVzJxyIPqc:-Kl0DrlRbdaxHN3LxoTu:bv7my5g) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L239)):*

> if that doesn't work

How likely is this? What is the recourse for a project for which this happens yet they want to inject the dependency attribution?
</blockquote></details>

I was surprised Bloom let me do this, actually. I think that directory is read-only, but the program wrote to it, no problem. As you can see from the catch, I've left this here in case of an UnauthorizedAccessException. But really I have no idea how prevalent that'll be.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

andrew-polk:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 239 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Ab06A2yVzJxyIPqc:-Kl0FOKWdiN4k0unKtrP:b1rousj) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L239)):*

I was surprised Bloom let me do this, actually. I think that directory is read-only, but the program wrote to it, no problem. As you can see from the catch, I've left this here in case of an UnauthorizedAccessException. But really I have no idea how prevalent that'll be.
</blockquote></details>

Bloom is installed in AppData which the user always (should never say always... or never...) has write access to. Most programs are installed in Program Files which the user doesn't have access to (or would have to authorize via UAC which would also be bad).
In my opinion, we need to (not mutually exclusive):
1) Test another app which installs in Program Files
2) Come up with a different strategy
3) Have a plan for when projects come across this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

gmartin7:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 239 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Ab06A2yVzJxyIPqc:-Kl0OTLWE0GyP-r7_EW4:bnvpfdp) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L239)):*

Bloom is installed in AppData which the user always (should never say always... or never...) has write access to. Most programs are installed in Program Files which the user doesn't have access to (or would have to authorize via UAC which would also be bad).
In my opinion, we need to (not mutually exclusive):
1) Test another app which installs in Program Files
2) Come up with a different strategy
3) Have a plan for when projects come across this
</blockquote></details>

3. is "make the new file in a different location and forget about CSS!"
Plus, this can't be a problem until somebody tries to use the DependencyMarker anyway!
To do 1., we have to have another app that installs in Program Files and already built on the new libpalaso code.
Plus, I've spent way to much time on this already. :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

JohnThomson:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 239 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Ab06A2yVzJxyIPqc:-Kl0K8xGpB6QQRKwdued:boq9ox9) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L239)):*

3. is "make the new file in a different location and forget about CSS!"
Plus, this can't be a problem until somebody tries to use the DependencyMarker anyway!
To do 1., we have to have another app that installs in Program Files and already built on the new libpalaso code.
Plus, I've spent way to much time on this already. :(
</blockquote></details>

Even for Bloom, this will not work when it is installed for all users in Program Files, which our installer does support. And even though it's possible for a normal Bloom installation, writing something to the directory that contains our own program feels wrong. It's asking for trouble with anti-virus things. A user OUGHT to be able to protect himself against viruses by locking down the Bloom directory once it's installed.
Some more options:
  4. Document that an About Box HTML file which uses the dependency injection may not depend on relative paths to files
  5. Insert a <base> tag at the start of the derived HTML file that tells it to look for relative files in the original folder.
  6. Document that an About Box HTML file which uses dependency injection must live in its own folder with all dependent files, and copy the whole folder to a temp folder.
  7. Document that an About Box HTML file which uses dependency injection may only use a few common kinds of relative links, search for matching links, and copy the appropriate files to a temp directory along with the temp file.

(4) limits user options and I think will completely prevent images in about boxes, so it's probably too restrictive.
Unfortunately I think (5) will run afoul of cross-domain rules and won't work.
(6) could work but is a nuisance, especially for anyone who doesn't need any dependencies.
I rather like (7). A fairly simple regular expression will search for src or rel followed by a value with no path separators...something like (src|rel)=(['"])([^/\\]*)\1 (or something similar...handle white space...). That will catch all references to images, stylesheets, and scripts, and if the bit of the RE that matches the filename corresponds to an existing file in the same folder as the HTML we can just copy it. Unless they're doing relative paths to different folders that will do it, and I think it's reasonable to have SOME restrictions in the interests of simplicity.

Definitely test whatever you come up with on an allUsers install of Bloom (after enabling the new feature)!

Sorry to be difficult. If we can't afford the time for (7) we can achieve (4) by just discarding the "write to the same directory" option. It doesn't mean forgetting about CSS altogether, just forgetting about putting CSS in its own file (and therefore about less etc.). (7) might be worth capturing in comments as a possible enhancement.

I think the install-folder option is too dangerous to keep. It will very likely work (since a programmer can write to his own output folder) until a program actually ships, at which point the code will switch to the other option and any links will quietly break. And quite likely no one thought to test the About box in a real installed release!

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 229 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Jgdxa5xgj-sN7RZi:-Kl0Jgdxa5xgj-sN7RZj:b91b6ma) ([raw file](https://github.com/sillsdev/libpalaso/blob/93087533c61194724a1dbd75c4d1b21339ecd30d/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L229)):*
> ```Smalltalk
> 				{
> 					var temporaryName = "temp." + Path.GetFileName(filePath);
> 					var tempPath = Path.GetDirectoryName(filePath) + Path.DirectorySeparatorChar + temporaryName;
> ```

Path.Combine is the usual idiom here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

gmartin7:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 239 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Ab06A2yVzJxyIPqc:-Kl0WFntJwYeULwGDA3Y:b-ly5yo9) ([raw file](https://github.com/sillsdev/libpalaso/blob/cfb7c37cf384a2570690f09c1faf2a0a8de6cbe0/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L239)):*

Even for Bloom, this will not work when it is installed for all users in Program Files, which our installer does support. And even though it's possible for a normal Bloom installation, writing something to the directory that contains our own program feels wrong. It's asking for trouble with anti-virus things. A user OUGHT to be able to protect himself against viruses by locking down the Bloom directory once it's installed.
Some more options:
  4. Document that an About Box HTML file which uses the dependency injection may not depend on relative paths to files
  5. Insert a <base> tag at the start of the derived HTML file that tells it to look for relative files in the original folder.
  6. Document that an About Box HTML file which uses dependency injection must live in its own folder with all dependent files, and copy the whole folder to a temp folder.
  7. Document that an About Box HTML file which uses dependency injection may only use a few common kinds of relative links, search for matching links, and copy the appropriate files to a temp directory along with the temp file.

(4) limits user options and I think will completely prevent images in about boxes, so it's probably too restrictive.
Unfortunately I think (5) will run afoul of cross-domain rules and won't work.
(6) could work but is a nuisance, especially for anyone who doesn't need any dependencies.
I rather like (7). A fairly simple regular expression will search for src or rel followed by a value with no path separators...something like (src|rel)=(['"])([^/\\]*)\1 (or something similar...handle white space...). That will catch all references to images, stylesheets, and scripts, and if the bit of the RE that matches the filename corresponds to an existing file in the same folder as the HTML we can just copy it. Unless they're doing relative paths to different folders that will do it, and I think it's reasonable to have SOME restrictions in the interests of simplicity.

Definitely test whatever you come up with on an allUsers install of Bloom (after enabling the new feature)!

Sorry to be difficult. If we can't afford the time for (7) we can achieve (4) by just discarding the "write to the same directory" option. It doesn't mean forgetting about CSS altogether, just forgetting about putting CSS in its own file (and therefore about less etc.). (7) might be worth capturing in comments as a possible enhancement.

I think the install-folder option is too dangerous to keep. It will very likely work (since a programmer can write to his own output folder) until a program actually ships, at which point the code will switch to the other option and any links will quietly break. And quite likely no one thought to test the About box in a real installed release!
</blockquote></details>

Going with 4. for now. Someone smarter can continue the fight later. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

gmartin7:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs, line 229 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-Kl0Jgdxa5xgj-sN7RZi:-Kl0WjhuG8Up-KBv7wpj:b638bxz) ([raw file](https://github.com/sillsdev/libpalaso/blob/93087533c61194724a1dbd75c4d1b21339ecd30d/SIL.Windows.Forms/Miscellaneous/SILAboutBox.cs#L229)):*

Path.Combine is the usual idiom here
</blockquote></details>

Of course. Silly me. But now that code is gone anyway! :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530)*

JohnThomson:
Changes look good as far as they go. Thoughts
- didn't we remove the validation because it would remove the need for a whole DLL? If so there should be a change to a csproj to remove the reference to that DLL, and probably some other change(s) so we no longer fetch it, ship it, etc., unless other libpalaso DLLs use it.
- Somewhere there should be a comment, maybe on the class, that describes how the about box is used and includes a description of the dependency injection. That should be updated to warn that you can't use both dependency injection and relative links.
- I was hoping you would incorporate my thoughts on options for getting around that restriction as some sort of "possible enhancements" comment so they aren't lost

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-:-Kl0Yn9wDp2QgxqJCqlE:b37q8mi)*

gmartin7:
1st bullet: All taken care of. It was only earlier amendments of this commit that added it, so no need to delete what this version doesn't add. You'll see in the .csproj file, however, that we delete the tests file I created for the ValidateHtml method in an earlier merged commit, since that method has gone away.
2nd bullet: Done.
3rd bullet: Done.

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-:-Kl0cG6pOOQ87oY_kcJR:b-6kt4ad)*

JohnThomson:
:lgtm:

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/530#-:-Kl3yGgYtZIq1gHhxntX:bnfp4nl)*

PR #534, opened 2017-05-30
Remove references to IrrkLang in project but not in code

gtryus:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/534)


papeh:
Looks good to me. Too bad @palaso-bob is using the wrong TeamCity configuration for this branch.

ermshiperete:
The commit message confused me. Might be good to mention that we don't need the reference to IrrKlang in these projects because that's referenced and implemented in SIL.Media.

jasonleenaylor:
I was also confused by the commit message. A better one would be "Remove unneeded references to iirkLang" as the subject with content mentioning the why as @ermshiperete suggested.

PR #535, opened 2017-05-30
Warning message ignored for 32-bit build and 64-bit build

gtryus:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/535)

### DIFF
--- a/SIL.Windows.Forms.Scripture/SIL.Windows.Forms.Scripture.csproj
+++ b/SIL.Windows.Forms.Scripture/SIL.Windows.Forms.Scripture.csproj
@@ -272,6 +272,7 @@
     <DocumentationFile>..\output\Release\SIL.Windows.Forms.Scripture.xml</DocumentationFile>
     <Optimize>true</Optimize>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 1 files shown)


papeh:
the commit message could use a comment stating what warning 1591 is and why we need to ignore it

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/535#-:-KlQPkbO1G57pP7-X3Tr:bj6iq13)*

gtryus:
The C# error is described here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591

Apparently the project is set up to generate documentation in XML format which requires that each type and each member have a comment for the documentation. Not all types and members have been given comments. This is how the code was for 32-bit. The warning was being ignored. When I created the 64-bit release build, I didn't have this error being ignored and so the release build (which is what Team City uses) is failing for lack of complete comments.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/535#-:-KlUGqTWoSHLiQDWZFrm:b6702rc)*

PR #539, opened 2017-06-01
For Mixed Platform use AnyCPU

gtryus:
This allows the assemblies to be used on 32-bit (x86) and 64-bit.
Two assemblies were left as x86: SIL.Media and SIL.Windows.Forms.Keyboard
SIL.Media because it has platform specific dependencies
and keyboard because a test fails when it isn't x86.

Mixed Platforms generally means AnyCPU on Mono configs too
Also corrected issues with ExtraCopyright proj files using
incorrect paths for some build by updating to look like
LanguageData project.
  Also reviewed x86, x64, and AnyCPU to make sure they
specify the selected Configuration and Platform.
  An exception is made for SIL.Media which doesn't have an AnyCPU
on Windows so Deubg, DebugStrongName, Release and
ReleaseStrongName use the x86 build.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/539)


papeh:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---

*[Palaso.sln, line 2 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/539#-Kl_WdtkQUPiCkci5oYM:-Kl_WdtlfGx8xiRn9e6B:b-mzlg74) ([raw file](https://github.com/sillsdev/libpalaso/blob/af1cada29348f1e2a6be8b2e62d4674a8128f284/Palaso.sln#L2)):*
> ```
> Microsoft Visual Studio Solution File, Format Version 12.00
> # Visual Studio 14
> ```

a step back in time?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/539#-:-Kl_X-u-6lWjZLuY7HfF:bnfp4nl)*

gtryus:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Palaso.sln, line 2 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/539#-Kl_WdtkQUPiCkci5oYM:-KldZuRxLhRtwty2Hg6R:bvaj875) ([raw file](https://github.com/sillsdev/libpalaso/blob/af1cada29348f1e2a6be8b2e62d4674a8128f284/Palaso.sln#L2)):*

a step back in time?
</blockquote></details>

I think this is added by Visual Studio 2015 which is what I have been using for all the development work since it works for FieldWorks. Do you think this will cause a problem for anyone?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/539)*

PR #545, opened 2017-06-21
Fixed a couple issues found with the Versification.Table class

FoolRunning:
Fixed not all versifications being returned by the VersificationTables method (i.e. no built-in versifications were ever being included).

Fixed built-in versifications not being cached so a new one was created each time. This resulted in significantly increased memory usage and decreased performance (because a deep-compare was then necessary).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/545)


FoolRunning:
Fixing another bug just found with versification that I'm going to add to this PR, so I'm closing it.

PR #546, opened 2017-06-21
Fixed a few bugs found in the Versification.Table class

FoolRunning:
Fixed not all versifications being returned by the VersificationTables method (i.e. no built-in versifications were ever being included).

Fixed built-in versifications not being cached so a new one was created each time. This resulted in significantly increased memory usage and decreased performance (because a deep-compare was then necessary).

Fixed a versification named a number (e.g. "30") throwing an exception (caused by **Enum.TryParse** accepting the string representation of a number as a valid value)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/546)

### DIFF
--- a/SIL.Scripture.Tests/ScrVersTests.cs
+++ b/SIL.Scripture.Tests/ScrVersTests.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using NUnit.Framework;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using SIL.Reflection;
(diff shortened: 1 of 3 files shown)


andrew-polk:
A minor niggle. Otherwise, :lgtm: 

---

Reviewed 3 of 3 files at r1.

---

*[SIL.Scripture.Tests/ScrVersTests.cs, line 949 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/546#-KnEhpOm65jEUB9RpZAy:-KnEhpOm65jEUB9RpZAz:b-euh1fd) ([raw file](https://github.com/sillsdev/libpalaso/blob/71caa3de5121193cd7a1ad20f7d48f54b367c561/SIL.Scripture.Tests/ScrVersTests.cs#L949)):*
> versification

libpalaso convention is that private members start with _.
That would have saved me a few seconds understanding this test. : )

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/546#-:-KnEi9Ym91ALqxIU9xs0:b-88r6zj)*

PR #550, opened 2017-06-30
[API] Mark XmlUtils.AppendAttribute as obsolete

ermshiperete:
The preferred way is to use `SetAttribute` which deals with both
setting an attribute and replacing an existing one.

`AppendAttribute` always appended the attribute to the end of
the collection (and the collection removed an already existing one
with the same name). `SetAttribute` will keep an existing attribute
and just change its value, thus keeping the order of attributes.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/550)

### DIFF
--- a/SIL.Core/Xml/XmlUtils.cs
+++ b/SIL.Core/Xml/XmlUtils.cs
@@ -550,16 +550,10 @@ namespace SIL.Xml
 		/// <summary>
 		/// Append an attribute with the specified name and value to parent.
 		/// </summary>
-		/// <param name="parent"></param>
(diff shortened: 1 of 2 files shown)


papeh:
👍

gtryus:
It would be good to inform those responsible for code bases that use AppendAttribute that you are making it obsolete.

ermshiperete:
Email announcing the upcoming change sent to the `sil-lsdev` group.

ermshiperete:
Good point. Updated version to 4.1.

PR #553, opened 2017-07-12
PTX-51943 ignore content of bad existing ldml file when writing new file

jwickberg:
Paratext got errors when trying to update an existing ldml file when the existing file was invalid because the content was not actually written to disk.

Added a try/catch block so invalid existing content will be ignored.

Added a Flush call to the writer to hopefully force the written data to disk and prevent this problem.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/553)

### DIFF
--- a/SIL.WritingSystems/LdmlDataMapper.cs
+++ b/SIL.WritingSystems/LdmlDataMapper.cs
@@ -763,11 +763,11 @@ namespace SIL.WritingSystems
 		{
 			if (element != null)
 			{
-                if (element.IsEmpty || (string.IsNullOrEmpty((string)element) && !element.HasElements && !element.HasAttributes))
-                {
-                    element.Remove();
-                    element = null;
-                }
+				if (element.IsEmpty || (string.IsNullOrEmpty((string)element) && !element.HasElements && !element.HasAttributes))
+				{
+					element.Remove();
+					element = null;
+				}
 			}
 		}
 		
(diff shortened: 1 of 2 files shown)


SIL.WritingSystems/LdmlDataMapper.cs
```diff
@@ -806,7 +806,15 @@ public void Write(string filePath, WritingSystemDefinition ws, Stream oldFile)
 						DtdProcessing = DtdProcessing.Parse
 					};
 					reader = XmlReader.Create(oldFile, readerSettings);
-					element = XElement.Load(reader);
+					try
+					{
+						element = XElement.Load(reader);
+					}
+					catch
```
ddaspit: Is there a specific exception you can catch here? Maybe XmlException?

SIL.WritingSystems/LdmlDataMapper.cs
```diff
@@ -819,6 +827,7 @@ public void Write(string filePath, WritingSystemDefinition ws, Stream oldFile)
 				using (var writer = XmlWriter.Create(filePath, writerSettings))
 				{
 					WriteLdml(writer, element, ws);
+					writer.Flush();
```
ddaspit: I don't think this Flush() call should be necessary. Doesn't Close() flush the buffers? In fact, the Close() shouldn't even be necessary since the XmlWriter is wrapped in a using statement.

jwickberg: The errors I've seen have been caused by the write being done, but the disk file not being written to - at least in Windows the write to disk is normally deferred for an unspecified time. From my reading of the Flush on the Writer, it seems to do the force the disk update.

I created a FlushingStream for Paratext and this has prevented many errors related to files just containing nulls.

SIL.WritingSystems/LdmlDataMapper.cs
jwickberg: I couldn't think of any exception I would want to handle differently. Things like OutOfMemory shouldn't be caught since with .Net 4.0 those errors are not catchable by default. The XmlException is the only exception I've seen so far, so that may be sufficient.

PR #562, opened 2017-08-11
BL-4684 ensure metadata links not clicked when imageInfo null

JohnThomson:
An attempt to make certain that this crash can't happen because the relevant controls can't be visible when the variable is null

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/562)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs
+++ b/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs
@@ -68,15 +68,26 @@ namespace SIL.Windows.Forms.ImageToolbox
 			{
 				try
 				{
+					var oldValue = _imageInfo;
+					// We've been having trouble with an obscure situation where one of the edit metadata links is
+					// clicked while _imageInfo is null. Wondering whether an exception might somehow be caught
+					// in such a way that SetupMetaDataControls is called here based on value being non-null,
+					// but _imageInfo never actually gets set to the new value. Seems a setter should always
+					// set the backing variable to the new value, so we now do this first.
+					_imageInfo = value;
+
 					if (value == null || value.Image == null)
 					{
 						_currentImageBox.Image = null;
 						_metadataDisplayControl.Visible = false;
+						// These two controls when clicked run code that requires _imageInfo to be non-null.
+						// Make sure they are hidden if it is set to null.
 						_invitationToMetadataPanel.Visible = false;
+						_editLink.Visible = false;
 					}
 					else
 					{
-						if(value.Image == _currentImageBox.Image)
+						if (value.Image == _currentImageBox.Image)
 						{
 							return;
 						}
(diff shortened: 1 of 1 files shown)


andrew-polk:
One question. Otherwise, :lgtm: 

---

Reviewed 1 of 1 files at r1.

---

*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 174 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/562#-KrVuGqezMBRlTQhoBh2:-KrVuGqfJb6ueqQ7v5E3:b-3tjc5p) ([raw file](https://github.com/sillsdev/libpalaso/blob/2d2312d7f07a32e4e90e638f451ef0865827906b/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L174)):*
> ```Smalltalk
> 		private void SetupMetaDataControls(Metadata metaData)
> 		{
> 			Guard.AgainstNull(_imageInfo, "_imageInfo");
> ```

I'm sure this is probably fine, but I wanted to throw out a question in case...

In the case where the first 'if' condition is true, we don't care if _imageInfo is null. So should the check be moved down? Or is it just simply wrong to call this method in that state?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/562#-:-KrVunwwIjb2JK4TuWhn:bijt0ar)*

JohnThomson:

---

*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 174 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/562#-KrVuGqezMBRlTQhoBh2:-KrXc2-T_7Cr7IntnYf5:b-5fitzs) ([raw file](https://github.com/sillsdev/libpalaso/blob/2d2312d7f07a32e4e90e638f451ef0865827906b/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L174)):*

I'm sure this is probably fine, but I wanted to throw out a question in case...

In the case where the first 'if' condition is true, we don't care if _imageInfo is null. So should the check be moved down? Or is it just simply wrong to call this method in that state?
</blockquote></details>

It's a good question. Three of four callers are passing _imagInfo.Metadata as argument; the fourth is passing value.metadata, where value is the one that has just been assigned to _imageInfo. So IN THEORY there is absolutely no possible way we can call this method with _imageInfo null. It should be a waste to have the guard at all. But we're getting crashes where these controls are visible and _imageInfo is null, which also shouldn't be possible. So I put it in for paranoia and in hopes of further localizing the weirdness if it somehow persists.
Now, it could at some future time make sense to call this with _imageInfo null and use it as a way to hide those controls. At that point the guard would need to be moved. But it is present to verify the theory that this method is NEVER called with _imageInfo null. So I am inclined to leave it as-is for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/562)*

PR #563, opened 2017-08-22
Fix LT-18510 Do not make a copy of an audio file

RaymondLuong3:
* If the file already exists in the AudioVisual directory, use that file instead of creating a new file.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/563)

### DIFF
--- a/SIL.Media/ShortSoundFieldControl.cs
+++ b/SIL.Media/ShortSoundFieldControl.cs
@@ -13,6 +13,7 @@ namespace SIL.Media
 		private ISimpleAudioSession _recorder;
 		private string _path;
 		private string _deleteButtonInstructions = "Delete this recording.";
+        public bool CopyFilesAlreadyInDefaultDir { get; set; }
 
 		public event EventHandler BeforeStartingToRecord;
 		public event EventHandler SoundRecorded;
@@ -30,6 +31,7 @@ namespace SIL.Media
 			_playButton.MouseEnter += delegate(object o, EventArgs args) { ButtonEnter(args); };
 			_playButton.MouseLeave += delegate(object o, EventArgs args) { ButtonLeave(args); };
 			_deleteButton.MouseLeave += delegate(object o, EventArgs args) { ButtonLeave(args); };
+            CopyFilesAlreadyInDefaultDir = true;
 		}
 
 		private void UpdateButtonAppearances(bool enable)
@@ -241,17 +243,19 @@ namespace SIL.Media
 					AutoUpgradeEnabled = true,
 					Filter = "sound files (*.wav, *.mp3)|*.wav;*.mp3"
 				};
-				if (DialogResult.OK != dlg.ShowDialog())
+				if (DialogResult.OK != dlg.ShowDialog() || Path == dlg.FileName)
 				{
 					return false;
 				}
 				if (File.Exists(Path))
 					File.Delete(Path);
 
-				SetPathToAvailableFilenameInDefaultDir(dlg.FileName);
-				// *Always* make a copy, since the only way to change the recording is to delete the old one, and there is no delete confirmation.
+                UpdatePathToSoundFile(dlg.FileName);
+				// (Unless CopyFilesAlreadyInDefaultDir is false) *Always* make a copy, since the only way
+				// to change the recording is to delete the old one, and there is no delete confirmation.
 				// (We don't want to make it that easy to delete a file that somebody else might be using)
-				File.Copy(dlg.FileName, Path);
+				if (!File.Exists(Path))
+					File.Copy(dlg.FileName, Path);
 			}
 			catch (Exception error)
 			{
@@ -260,6 +264,16 @@ namespace SIL.Media
 			UpdateScreen();
 			return true;
 		}
+
+        private void UpdatePathToSoundFile(string dlgPath)
+        {
+	        string mediaDir = System.IO.Path.GetDirectoryName(Path);
+            if (!CopyFilesAlreadyInDefaultDir && dlgPath.StartsWith(mediaDir))
+                Path = dlgPath;
+            // the file is outside the media directory, make a copy
+            else
+                SetPathToAvailableFilenameInDefaultDir(dlgPath);
+        }
 
 		/// <summary>
 		/// LT-15375 Don't copy the file to some weird name.


hatton:
@RaymondLuong3 what product are you working on? Is there an issue somewhere that write up the problem and this change in behavior?

RaymondLuong3:
I'm working on FieldWorks. Audio files are duplicated when they already exist in the AudioVisual directory of a project. You might be able to see it here. https://jira.sil.org/browse/LT-18510

hatton:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 252 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ks9vNbWljbGvSvobkDH:-Ks9vNbWljbGvSvobkDI:b-gbz975) ([raw file](https://github.com/sillsdev/libpalaso/blob/8b4ba1d7342ef6fea6380cf1e3f7ac7d5324f806/SIL.Media/ShortSoundFieldControl.cs#L252)):*
> ```Smalltalk
> 
> 				SetPathToAvailableFilenameInDefaultDir(dlg.FileName);
> 				// Copies the file into the default directory (AudioVisual/) if it is not already there.
> ```

I don't see any reference to "AudioVisual" in this library.  Perhaps that is specific to the client app you are working on? Perhaps you could write "e.g., AudioVisual in FLEx".

---

*[SIL.Media/ShortSoundFieldControl.cs, line 274 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ks9vRZCaXw9S7qY5vOs:-Ks9vRZCaXw9S7qY5vOt:b3rutj3) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ab19fb5a9727b1b15b29316bea0c3ceeb4b6bb/SIL.Media/ShortSoundFieldControl.cs#L274)):*
> ```Smalltalk
> 			var newPath = System.IO.Path.Combine(defaultDir, dlgName);
> 			if(File.Exists(newPath))
> 			{
> ```

Your PR here is removing some pretty intentional functionality. Your client app might not need it, but how can we know that no other apps are relying on this functionality (e.g. WeSay, which was the original client of this control)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

hatton:
@glasseyes  Daniel, can you look take over looking at this from the WeSay perspective? Thanks.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-:-Ks9vlrwv5ZLPO6UccBl:b-fczglr)*

RaymondLuong3:
@hatton I'm having trouble commenting on Reviewable. Thanks for your input, I realize that copying the file to the default directory is an intentional functionality. If WeSay depends on this (or other programs) I will concede to not make the change.

glasseyes:
The audio stuff is crashing in WeSay at the moment (like in https://jira.sil.org/browse/WS-398) in irrKlang. So I can't test what WeSay would do with these changes. I'm away next week but when I get back I'm planning to backport the fixes that are already in the libpalaso branch to libpalaso-2.6 (and possibly libpalaso-3.1) before looking at this. My gut feeling looking at the code is that the way it was done is deliberate so I would rather see a different way to do it.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-:-KsNkk07pEB6XrLufo0i:b-ts60w8)*

hatton:
@raymondluong3  How about you add a parameter that activates the change behavior you want. Set its default to the current behavior.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-:-KsOaQH7pxOdEggqpAVe:b-reac2d)*

RaymondLuong3:
@hatton Thanks for the suggestion. I will look into a way of setting a flag so that this SIL.Media assembly knows that FLEx is calling and will not duplicate the audio file. It looks tricky, hopefully I can figure it out.

RaymondLuong3:
Yikes! I mistakenly reverted merge #562. It should be fixed now. I'm am just finding my way around git.

papeh:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 277 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-KtNhdRW1DQ8MnPzvvuK:-KtNhdRW1DQ8MnPzvvuL:b4kuk0n) ([raw file](https://github.com/sillsdev/libpalaso/blob/19ae08fc9bf38f14e7f7ed392d6fa739e397370e/SIL.Media/ShortSoundFieldControl.cs#L277)):*
> ```Smalltalk
> 
> 			// Bypass the file disambiguation when the client application is FieldWorks
> 			bool clientIsFieldWorks = Application.ExecutablePath.ToLower().Contains("fieldworks");
> ```

I think it would be better to pass a parameter from the client application.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

papeh:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 248 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4i4lCfyosDIAQ_Bly:-Ku4i4lCfyosDIAQ_Blz:b-f1a1kx) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L248)):*
> ```Smalltalk
> 					Filter = "sound files (*.wav, *.mp3)|*.wav;*.mp3"
> 				};
> 				if (DialogResult.OK != dlg.ShowDialog())
> ```

not part of the request per se, but it would be good to add an idiot check here: if `Path == dlg.FileName`, we should also return false.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 276 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4iqyH5-A_VEjJGqZV:-Ku4iqyH5-A_VEjJGqZW:b-ym41ly) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L276)):*
> ```Smalltalk
>             else
>             {
>                 if (PathToMediaRootDir != String.Empty && dlgPath.Contains(PathToMediaRootDir))
> ```

`string.IsNullOrEmpty(PathToMediaRootDir)`
`dlgPath.StartsWith`

---

*[SIL.Media/ShortSoundFieldControl.cs, line 282 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4iZdnG9qtAEyZUa6Z:-Ku4iZdnG9qtAEyZUa6_:bocpw0j) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L282)):*
> ```Smalltalk
>                     SetPathToAvailableFilenameInDefaultDir(dlgPath);
>             }
>         }
> ```

This method can be refactored to have only one if statement

---

*[SIL.Media/ShortSoundFieldControl.cs, line 290 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4jLdfnYS6HSvHiOeD:-Ku4jLdfnYS6HSvHiOeE:bbp02d7) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L290)):*
> ```Smalltalk
> 		private void SetPathToAvailableFilenameInDefaultDir(string dlgPath)
> 		{
> 			var defaultDir = System.IO.Path.GetDirectoryName(Path);
> ```

r5 introduces a new member variable PathToMediaRootDir. That should always be the same as what we're calculating here, so we shouldn't need the new member variable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

RaymondLuong3:
*[SIL.Media/ShortSoundFieldControl.cs, line 290 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4jLdfnYS6HSvHiOeD:-KuW7-7gnKWAAsMBTtYC:bbpyxce) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L290)):*

r5 introduces a new member variable PathToMediaRootDir. That should always be the same as what we're calculating here, so we shouldn't need the new member variable.
</blockquote></details>

For FLEx, the defaultDir is the LinkedFiles/AudioVisual directory, but we want the PathToMediaRootDir to be the LinkedFiles directory. Was there something I am overlooking?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

RaymondLuong3:
@hatton Thanks for looking this over. Yes, I did see I was removing an intentional functionality. If this is not possible, I can request that FLEx accepts the previous functionality as correct behavior.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 252 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ks9vNbWljbGvSvobkDH:-KsAIBAA5_Ja25mA_ubj:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/8b4ba1d7342ef6fea6380cf1e3f7ac7d5324f806/SIL.Media/ShortSoundFieldControl.cs#L252)):*

I don't see any reference to "AudioVisual" in this library.  Perhaps that is specific to the client app you are working on? Perhaps you could write "e.g., AudioVisual in FLEx".
</blockquote></details>

Done.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 274 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ks9vRZCaXw9S7qY5vOs:-KsAIHThATkeUUksSYY7:bmhd3nb) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ab19fb5a9727b1b15b29316bea0c3ceeb4b6bb/SIL.Media/ShortSoundFieldControl.cs#L274)):*

Your PR here is removing some pretty intentional functionality. Your client app might not need it, but how can we know that no other apps are relying on this functionality (e.g. WeSay, which was the original client of this control)?
</blockquote></details>

If Wesay depends on this behavior, I will concede that this should not be changed

---

*[SIL.Media/ShortSoundFieldControl.cs, line 277 at r4](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-KtNhdRW1DQ8MnPzvvuK:-KuW6zn3tBbcpGJkGP_O:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/19ae08fc9bf38f14e7f7ed392d6fa739e397370e/SIL.Media/ShortSoundFieldControl.cs#L277)):*

I think it would be better to pass a parameter from the client application.
</blockquote></details>

Done.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 248 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4i4lCfyosDIAQ_Bly:-KuW6uJymlNxuNVv9nml:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L248)):*

not part of the request per se, but it would be good to add an idiot check here: if `Path == dlg.FileName`, we should also return false.
</blockquote></details>

Done.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 276 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4iqyH5-A_VEjJGqZV:-KuW6xPld8kd6AmoGbe3:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L276)):*

`string.IsNullOrEmpty(PathToMediaRootDir)`
`dlgPath.StartsWith`
</blockquote></details>

Done.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 282 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4iZdnG9qtAEyZUa6Z:-KuW6wVDwFl7hm7CsRC5:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L282)):*

This method can be refactored to have only one if statement
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-:-KsA7UAS6wPjG8S0KgJT:b-39uo4h)*

RaymondLuong3:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 274 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ks9vRZCaXw9S7qY5vOs:-KuW7nExPDMmXxWtoy2N:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ab19fb5a9727b1b15b29316bea0c3ceeb4b6bb/SIL.Media/ShortSoundFieldControl.cs#L274)):*
<details><summary><i>Previously, RaymondLuong3 (Raymond Luong) wrote…</i></summary><blockquote>

If Wesay depends on this behavior, I will concede that this should not be changed
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

papeh:
Could you upload the FLEx changes, too?

---

Reviewed 1 of 1 files at r6.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 290 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4jLdfnYS6HSvHiOeD:-Ku_-WmUpGcQQD0b5cKJ:b-huv59e) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L290)):*
<details><summary><i>Previously, RaymondLuong3 (Raymond Luong) wrote…</i></summary><blockquote>

For FLEx, the defaultDir is the LinkedFiles/AudioVisual directory, but we want the PathToMediaRootDir to be the LinkedFiles directory. Was there something I am overlooking?
</blockquote></details>

That's kind of a rare case, to link to an audio file outside the AudioVisual directory.  I would argue that this doesn't merit special handling, but if you want to support users who think they want to do this, I won't stop you.

r5 also introduced `CopyFilesAlreadyInDefaultDir`.  We may wish to consider whether we need both (presently, `CopyFiles...` has meaning only if `PathTo...` is set to a valid path)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-:-Ku_1Z9CygimCADdWR65:bwokgsk)*

RaymondLuong3:
Yes, let me revise it to match the latest correction.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-:-Kuenk406G05UXA75XVv:bttrmr0)*

RaymondLuong3:
*[SIL.Media/ShortSoundFieldControl.cs, line 290 at r5](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-Ku4jLdfnYS6HSvHiOeD:-KuenRROrE0XNXF-ryeJ:b-681rok) ([raw file](https://github.com/sillsdev/libpalaso/blob/b169b21435df8a982bb40f9078769614150d1578/SIL.Media/ShortSoundFieldControl.cs#L290)):*

That's kind of a rare case, to link to an audio file outside the AudioVisual directory.  I would argue that this doesn't merit special handling, but if you want to support users who think they want to do this, I won't stop you.

r5 also introduced `CopyFilesAlreadyInDefaultDir`.  We may wish to consider whether we need both (presently, `CopyFiles...` has meaning only if `PathTo...` is set to a valid path)
</blockquote></details>

This sounds reasonable to me. I'll limit audio files to be in the AudioVisual and subfolders

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

papeh:
Reviewed 1 of 1 files at r7.

---

*[SIL.Media/ShortSoundFieldControl.cs, line 258 at r6](https://reviewable.io:443/reviews/sillsdev/libpalaso/563#-KutXxqyFPl_VyyvJYUm:-KutXxqyFPl_VyyvJYUn:b-6qlz9m) ([raw file](https://github.com/sillsdev/libpalaso/blob/8e3c5c621fcfeb166f0ce35e10568dfae003d9fb/SIL.Media/ShortSoundFieldControl.cs#L258)):*
> The only way to change the recording is to delete the old one . . .  we don't want . . . to delete a file that somebody else might be using.

When we made extraneous copies, we could delete them without worrying, because they were *our* copies (save those that were afterwards linked by CmFiles). Now, if we link a file without making our own copy, changing or recording will either delete or record over the shared copy.  We could:

- Abort this change (keep creating extra copies)
- if !CopyFilesAlreadyInDefaultDir, don't bother deleting the old file (leaves a bunch of unmanaged files)
- Find or create a utility to clean up orphan files (CmFiles and AudioWS files) that can be run manually
- Track all AWS files with CmFiles (would require another Property, or FLEx would have to have sole control over deletion, which normally happens inside the library call)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/563)*

papeh:
Shelving for now. There is a workaround for the user, and it would be significant effort to implement the request without the potential to unintentionally delete files.

PR #564, opened 2017-09-06
empty spans in empty text cause parsing to fail

glasseyes:
WS-451 this causes WeSay to fail to load a project and exit

fix makes alternative an empty string instead of null text object
TryGetValue only fails if there is no text in the <span> or <text>

alternative = new LiftString(); creates a LiftString with Text = null
so alternative.Text.Length throws an exception

Also adds a test to comfirm that empty spans pass parsing

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/564)

### DIFF
--- a/Palaso.Lift.Tests/Parsing/ParserTests.cs
+++ b/Palaso.Lift.Tests/Parsing/ParserTests.cs
@@ -177,6 +177,14 @@ namespace Palaso.Lift.Tests.Parsing
 			SimpleCheckGetOrMakeEntry_InsertVersion("<lift V />", 0);
 		}
 
+		[Test]
+		public void EmptySpanOk()
+		{
+			_doc.LoadXml("<example><form lang='en-US-fonipa'><text><span lang='en-US'></span></text></form></example>");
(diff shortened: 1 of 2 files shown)


papeh:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[Palaso.Lift/Parsing/LiftMultiText.cs, line 409 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/564#-KtNdYR-468wx2hHTkZZ:-KtNdYR-468wx2hHTkZ_:b-rghgvb) ([raw file](https://github.com/sillsdev/libpalaso/blob/b186e0bbed4ba98dab0dc3452fb3b8db5cb86f6c/Palaso.Lift/Parsing/LiftMultiText.cs#L409)):*
> ```Smalltalk
> 			{
> 				// When you get to here you know that textNode.InnerText.Length=0
> 				alternative = new LiftString("");
> ```

Would it be better to  change the default constructor instead of this call?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/564#-:-KtNeDTuwhCocxL-fGT8:bnfp4nl)*

PR #565, opened 2017-09-07
BL-4780 Allow setting aliases for languages in LanguageLookupDialog

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/565)

### DIFF
--- a/SIL.Windows.Forms.WritingSystems/LanguageLookupControl.cs
+++ b/SIL.Windows.Forms.WritingSystems/LanguageLookupControl.cs
@@ -15,6 +15,7 @@ namespace SIL.Windows.Forms.WritingSystems
 	{
 		private readonly LanguageLookupModel _model;
 		private string _lastSearchedForText;
+		private Dictionary<string, string> _languageNameAliases = new Dictionary<string, string>();
 
 		public event EventHandler ReadinessChanged;
 
@@ -65,8 +66,14 @@ namespace SIL.Windows.Forms.WritingSystems
 			{
 				string oldLangTag = _model.LanguageTag;
 				_model.SelectedLanguage = value;
-				_desiredLanguageDisplayName.Text = _model.DesiredLanguageName;
-				if(_model.LanguageTag != oldLangTag)
+				var newName = _model.DesiredLanguageName;
+				string alias;
+				if (value != null && _languageNameAliases.TryGetValue(value.LanguageTag, out alias))
+				{
+					_model.DesiredLanguageName = newName = alias;
+				}
+				_desiredLanguageDisplayName.Text = newName;
+				if (_model.LanguageTag != oldLangTag)
 					UpdateReadiness();
 			}
 		}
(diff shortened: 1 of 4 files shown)


hatton:
John if I'm reading this right, you're leaving it to the app to have to know to make this fix. That doesn't seem right... this hard-won (!!!) knowledge about Chinese should flow to all our apps.

JohnThomson:
Don't merge yet...need to use the alias in one more place.

JohnThomson:
If you think getting these ideas was hard, just try persuading Ken and Jason and Marlon and whoever's doing WeSay that they are not only appropriate for literacy but also for linguistics and that all our programs should defy the Ethnologue and use different names for these codes...or persuade Ethnologue to change the names, which may get into ISO issues and changing something in a published standard...and then getting a new release of FLEx and WeSay and ?? out that does something about existing data that uses zh or cmn...2018? 2019? It took weeks if not months of work to get FLEx and WeSay handling Chinese language codes as well as they do now...a nightmare of messed up send/receive for years...FLEx wound up doing its own private fix (if the user chooses cmn FLEx substitutes zh-CN). I tried to get Ken interested in this issue early on because he knows all this history MUCH better than I do but so far he has sent me one short message and not added comments to the issue at all.
I'm not happy about a one-app solution but I don't even want to think about opening the can of worms involved in changing it for all our apps.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/565#-:-KtTGd1_VbG-hslUTkq0:b-7yvypj)*

JohnThomson:
(Think "agreeing on the next version of LIFT" * 3)

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/565#-:-KtTLQdvd5-5rbyvl-hw:bfocp8g)*

JohnThomson:
OK, I patched the PR to properly handle aliases when selecting an item in the list.
Maybe we could have a libpalaso method on LookupLanguage{Dialog,Control} something like SetLiteracyOverrides() which would at least make our choices available to others?

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/565#-:-KtTNxWiP2M2sx8yltqc:b-2crywq)*

hatton:
OK ok ok :-)

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/565#-:-KtTaCFG80I52C1oDIg1:b1e7eaf)*

andrew-polk:
I really like your idea of moving the logic to a method the client can call on the control/dialog. And, as I suggested on the Bloom PR, the logic for which codes map to which languages should be in a common data structure.
I think the method name should be more specific to Chines, though. Then, if there are more of these in the future, the client can pick and choose which to apply.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/565#-:-KtU9L0te-uYtb98dhJC:b-fo7n1v)*

andrew-polk:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/565#-:-KtXXzXP010Roo6tezFY:bnfp4nl)*

PR #568, opened 2017-09-19
Handle a problem loading an image index gracefully

hatton:
Previously, a failure would make the whole image gallery fail. With this, it will still load other indexes and keep as many of the bad index lines as it was able to parse before failing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/568)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageCollection.cs
+++ b/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageCollection.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -62,7 +62,8 @@ namespace SIL.Windows.Forms.ImageToolbox.ImageGallery
 			_indexOfWordsToRelativePath = new Dictionary<string, List<string>>();
 			_indexOfRelativeFilePathToKeywordsCsv = new Dictionary<string, string>();
 
-			using (var f = File.OpenText(FindIndexFileInFolder(FolderPath)))
+			var indexPath = FindIndexFileInFolder(FolderPath);
+			using (var f = File.OpenText(indexPath))
 			{
 				//skip header line, which was already read to make the index layout above
 				f.ReadLine();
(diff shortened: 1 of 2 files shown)


JohnThomson:
:lgtm: with minor and optional suggestions

---

Reviewed 2 of 2 files at r1.

---

*[SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageCollection.cs, line 88 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/568#-KuPxKMwI2yW5F3NMQJX:-KuPxKMwI2yW5F3NMQJY:b-v2ammx) ([raw file](https://github.com/sillsdev/libpalaso/blob/71df749a4f01e1e684445afa7d1ddb2da5e24b0f/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageCollection.cs#L88)):*
> ```Smalltalk
> 							if(_indexOfRelativeFilePathToKeywordsCsv.ContainsKey(relativePath))
> 								throw new ApplicationException($"{indexPath} has more than one row describing a file name '{relativePath}'. This is not allowed. You can use 'conditional formatting' with the condition 'duplicate' in a spreadsheet program to find duplicates.");
> 						}
> ```

Do you not want to re-throw any other ArgumentException that might occur?
I would also be inclined to save the problem path in a variable so you could report the first (or last) duplicate if any at the end of the loop, thus loading all non-duplicate and one of each duplicate.
But that would be icing on the cake. What you already have is a big improvement and it is fine with me to merge as-is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/568#-:-KuPyAnHyqLv2S7GOmq7:bup3cmm)*

hatton:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageCollection.cs, line 88 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/568#-KuPxKMwI2yW5F3NMQJX:-KuQ1a2POAvCyLZyBUxk:b-xvh0up) ([raw file](https://github.com/sillsdev/libpalaso/blob/71df749a4f01e1e684445afa7d1ddb2da5e24b0f/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageCollection.cs#L88)):*
> Do you not want to re-throw any other ArgumentException that might occur?

I do, thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/568)*

PR #571, opened 2017-09-20
Make new collection list via code instead of app.config

hatton:
... because app.config doesn't get used by the clients.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/571)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs
+++ b/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
@@ -228,6 +228,11 @@ namespace SIL.Windows.Forms.ImageToolbox.ImageGallery
 					_collectionDropDown.Visible = true;
 					_collectionDropDown.Text =
 						"Galleries".Localize("ImageToolbox.Galleries");
+					if(ImageToolboxSettings.Default.DisabledImageCollections == null)
+					{
+						ImageToolboxSettings.Default.DisabledImageCollections = new StringCollection();
+					}
+
 					foreach (var collection in _imageCollectionManager.Collections)
 					{
 					    if(ImageToolboxSettings.Default.DisabledImageCollections.Contains(collection.FolderPath))
--- a/SIL.Windows.Forms/app.config
+++ b/SIL.Windows.Forms/app.config
@@ -29,10 +29,7 @@
                 <value/>
             </setting>
             <setting name="DisabledImageCollections" serializeAs="Xml">
-                <value>
-                    <ArrayOfString xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-                        xmlns:xsd="http://www.w3.org/2001/XMLSchema" />
-                </value>
+                <value/>
             </setting>
         </SIL.Windows.Forms.ImageToolbox.ImageToolboxSettings>
         <SIL.Windows.Forms.Progress.LogBoxSettings>


StephenMcConnel:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/app.config, line 33 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/571#-KuVvvSY1KFYu0Glu2Pi:-KuVvvSY1KFYu0Glu2Pj:b-8z6jnp) ([raw file](https://github.com/sillsdev/libpalaso/blob/a3db84c85a05c2ebb1df5878d0af23a56c5e5e37/SIL.Windows.Forms/app.config#L33)):*
> ```
>             <setting name="DisabledImageCollections" serializeAs="Xml">
>                 <value>
>                     <ArrayOfString xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
> ```

Should there be a comment informing the future programmer that this is a mere placeholder for a
list added explicitly in ImageGalleryControl.cs?

---

*[SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs, line 236 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/571#-KuVvKX97uXemOqeW8b7:-KuVvKX97uXemOqeW8b8:bm3n74f) ([raw file](https://github.com/sillsdev/libpalaso/blob/c054df2ea79f8ad3982c1faf856301cca7837c96/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs#L236)):*
> ```Smalltalk
> 					}
> 
> 					foreach (var collection in _imageCollectionManager.Collections)
> ```

This works fine for an empty collection.  What if we get a non-empty list of collections that should be disabled?  Perhaps a comment here telling the programmer to explicitly add them here and not in app.config?  Or maybe that's an obvious inference.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/571)*

hatton:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs, line 236 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/571#-KuVvKX97uXemOqeW8b7:-KuVyWAKzO-2OfxcVk9m:b9o7ki2) ([raw file](https://github.com/sillsdev/libpalaso/blob/c054df2ea79f8ad3982c1faf856301cca7837c96/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs#L236)):*
> This works fine for an empty collection.  What if we get a non-empty list of collections that should be disabled?  

Sorry, I don't understand your question.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/571)*

gmartin7:
Reviewed 2 of 2 files at r1.

---

*[SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs, line 236 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/571#-KuVvKX97uXemOqeW8b7:-Ku_I0s4fg74zCrd2kGT:bj4f8at) ([raw file](https://github.com/sillsdev/libpalaso/blob/c054df2ea79f8ad3982c1faf856301cca7837c96/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs#L236)):*

> This works fine for an empty collection.  What if we get a non-empty list of collections that should be disabled?  

Sorry, I don't understand your question.
</blockquote></details>

My understanding is that Steve wants to leave a comment here so that when we DO have a collection to be disabled we know (a bit more clearly after we've forgotten what we did here) what to do about it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/571)*

hatton:

---

*[SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs, line 236 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/571#-KuVvKX97uXemOqeW8b7:-Ku_JtaOrJfrgC07TfiN:bydiywg) ([raw file](https://github.com/sillsdev/libpalaso/blob/c054df2ea79f8ad3982c1faf856301cca7837c96/SIL.Windows.Forms/ImageToolbox/ImageGallery/ImageGalleryControl.cs#L236)):*

My understanding is that Steve wants to leave a comment here so that when we DO have a collection to be disabled we know (a bit more clearly after we've forgotten what we did here) what to do about it.
</blockquote></details>

OK, I think the disconnect here is that *we* would never do this. This is a user setting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/571)*

PR #588, opened 2017-11-07
Fix RobustFile.Copy to handle empty files without crashing

StephenMcConnel:
See https://issues.bloomlibrary.org/youtrack/issue/BL-5191 for an
example of this actually happening.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/588)

### DIFF
--- a/SIL.Core.Tests/IO/RobustFileTests.cs
+++ b/SIL.Core.Tests/IO/RobustFileTests.cs
@@ -202,5 +202,28 @@ namespace SIL.Tests.IO
 				RobustFile.BufferSize = oldBufSize;
 			}
 		}
+
+		[Test]
+		public void CopyEmptyFile()
+		{
+			// Create an empty file.
+			var emptyFile = Path.GetTempFileName();
+			Assert.IsTrue(RobustFile.Exists(emptyFile));
+			var info = new FileInfo(emptyFile);
+			Assert.AreEqual(0, info.Length);
+
+			var destName = emptyFile + "-xyzzy";
+			// This was throwing a System.ArgumentOutOfRangeException exception before being fixed.
+			RobustFile.Copy(emptyFile, destName, true);
+			Assert.IsTrue(RobustFile.Exists(destName));
+			var infoDest = new FileInfo(destName);
+			Assert.AreEqual(0, infoDest.Length);
+
+			// Clean up after ourselves.
+			RobustFile.Delete(emptyFile);
+			Assert.IsFalse(RobustFile.Exists(emptyFile));
+			RobustFile.Delete(destName);
+			Assert.IsFalse(RobustFile.Exists(destName));
+		}
 	}
 }
--- a/SIL.Core/IO/RobustFile.cs
+++ b/SIL.Core/IO/RobustFile.cs
@@ -33,15 +33,14 @@ namespace SIL.IO
 
 		// This is just a guess at a buffer size that should make copying reasonably efficient without
 		// being too hard to allocate.
-		internal static int BufferSize = 1024 * 256; // internal so we can tweak in unit tests
+		internal static int BufferSize = FileStreamBufferSize; // internal so we can tweak in unit tests
 
 		public static void Copy(string sourceFileName, string destFileName, bool overwrite = false)
 		{
 			RetryUtility.Retry(() =>
 			{
 				// This is my own implementation; the .NET library uses a native windows function.
-				var bufSize = (int)Math.Min(BufferSize, new FileInfo(sourceFileName).Length);
-				var buffer = new byte[bufSize];
+				var buffer = new byte[BufferSize];
 				using (var input = new FileStream(sourceFileName, FileMode.Open, FileAccess.Read))
 				{
 					using (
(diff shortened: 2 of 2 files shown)


andrew-polk:
Reviewed 2 of 2 files at r1.

---

*[SIL.Core/IO/RobustFile.cs, line 49 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/588#-KyNOZbylt0lhWf7sRwE:-KyNOZbzmqAgFP2Mm499:b-f0l1r0) ([raw file](https://github.com/sillsdev/libpalaso/blob/6f81ea69be5442562b352b10fefe7cffe39caa8b/SIL.Core/IO/RobustFile.cs#L49)):*
> ```Smalltalk
> 				// (See https://issues.bloomlibrary.org/youtrack/issue/BL-5191.)
> 				if (bufSize == 0)
> 					bufSize = 8;
> ```

Given that the buffer size can be larger than the file size with no problem, can we just use BufferSize instead of min'ing it with file length?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/588)*

StephenMcConnel:

---

*[SIL.Core/IO/RobustFile.cs, line 49 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/588#-KyNOZbylt0lhWf7sRwE:-KyNTtPCgikTWo3DLbpG:buigskt) ([raw file](https://github.com/sillsdev/libpalaso/blob/6f81ea69be5442562b352b10fefe7cffe39caa8b/SIL.Core/IO/RobustFile.cs#L49)):*

Given that the buffer size can be larger than the file size with no problem, can we just use BufferSize instead of min'ing it with file length?
</blockquote></details>

I'm sure we could, but I wanted to make the minimal change needed to the method.  I'd want JohnT's feedback on that since he's the original author as far as I can tell.  (unless you are and it's hidden from blame by minor edits) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/588)*

JohnThomson:

---

*[SIL.Core/IO/RobustFile.cs, line 49 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/588#-KyNOZbylt0lhWf7sRwE:-KyR6m6m2L113LJCjCw9:b391lv7) ([raw file](https://github.com/sillsdev/libpalaso/blob/6f81ea69be5442562b352b10fefe7cffe39caa8b/SIL.Core/IO/RobustFile.cs#L49)):*

I'm sure we could, but I wanted to make the minimal change needed to the method.  I'd want JohnT's feedback on that since he's the original author as far as I can tell.  (unless you are and it's hidden from blame by minor edits) 
</blockquote></details>

I should have given this more thought originally. According to https://www.infoworld.com/article/3212988/application-development/how-to-not-use-the-large-object-heap-in-net.html, memory blocks smaller than 85K will be allocated in the normal heap, not the large-object one. That feels to me like a good thing, so 256K is too big...especially to allocate in order to copy say a 1K file. Even finding say a 40K block might take some garbage collection. It's not obvious to me that anything bigger than around 4K (the private FileStream.DefaultBufferSize) will save significant time...ideally we'd have time for some experiments. If I had some reason for using a buffer larger than that, I wish I'd commented it!
Given what I know now, I'd be inclined to drop the BufferSize constant, drop the min function (and hence the not-zero code), and just always allocate FileStreamBufferSize...unless you feel like doing some performance measurements to see whether a larger buffer helps for larger files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/588)*

StephenMcConnel:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[SIL.Core/IO/RobustFile.cs, line 49 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/588#-KyNOZbylt0lhWf7sRwE:-KyRIeexgf-A4X8BsJUN:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/6f81ea69be5442562b352b10fefe7cffe39caa8b/SIL.Core/IO/RobustFile.cs#L49)):*

I should have given this more thought originally. According to https://www.infoworld.com/article/3212988/application-development/how-to-not-use-the-large-object-heap-in-net.html, memory blocks smaller than 85K will be allocated in the normal heap, not the large-object one. That feels to me like a good thing, so 256K is too big...especially to allocate in order to copy say a 1K file. Even finding say a 40K block might take some garbage collection. It's not obvious to me that anything bigger than around 4K (the private FileStream.DefaultBufferSize) will save significant time...ideally we'd have time for some experiments. If I had some reason for using a buffer larger than that, I wish I'd commented it!
Given what I know now, I'd be inclined to drop the BufferSize constant, drop the min function (and hence the not-zero code), and just always allocate FileStreamBufferSize...unless you feel like doing some performance measurements to see whether a larger buffer helps for larger files.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/588)*

andrew-polk:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/588#-:-KyRJZpl4HUSuti3EieU:bnfp4nl)*

PR #598, opened 2018-01-30
Add word wrapping to BitmapButton control

StephenMcConnel:
This addresses a problem reported in Bloom localization
(https://issues.bloomlibrary.org/youtrack/issue/BL-5506).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/598)

### DIFF
--- a/SIL.Windows.Forms/Widgets/BitmapButton.cs
+++ b/SIL.Windows.Forms/Widgets/BitmapButton.cs
@@ -92,6 +92,16 @@ namespace SIL.Windows.Forms.Widgets
 			set{_TextDropShadow = value;}
 		}
 		/// <summary>
+		/// Enable the wrapping of the button text
+		/// </summary>
+		[Browsable(true),
+			Category("Appearance"),
+			Description("allows the text to wrap words in its box (if no shadow)"),
+			System.ComponentModel.RefreshProperties(RefreshProperties.Repaint)
+		]
+		public bool TextWordWrap { get; set; }
+
+		/// <summary>
 		/// Enables the dashed focus rectangle
 		/// </summary>
 		[Browsable(true),
@@ -339,6 +349,7 @@ namespace SIL.Windows.Forms.Widgets
 
 			ImageAttributes = new ImageAttributes();
 			DisabledTextColor = System.Drawing.Color.DimGray;
+			TextWordWrap = false;
 		}
 
 		// Expose the protected method in case some user wants to change the
@@ -645,9 +656,11 @@ namespace SIL.Windows.Forms.Widgets
 			if( (State == BtnState.Pushed) && (OffsetPressedContent) )
 				   rect.Offset(1,1);
 			//
-			// caculate bounding rectagle for the text
+			// calculate bounding rectangle for the text
 			//
 			System.Drawing.SizeF size = txt_Size(e.Graphics,this.Text,this.Font);
+			if (TextWordWrap && !TextDropShadow && size.Width > (float)this.Width)
+				size = CalculateBestWrappingSize(e.Graphics);
 			//
 			// calculate the starting location to paint the text
 			//
@@ -659,7 +672,10 @@ namespace SIL.Windows.Forms.Widgets
 			{
 				using(var solidBrush = new SolidBrush(DisabledTextColor))
 				{
-					e.Graphics.DrawString(this.Text, this.Font, solidBrush, pt.X + 1, pt.Y + 1);
+					if (TextWordWrap && !TextDropShadow)
+						e.Graphics.DrawString(this.Text, this.Font, solidBrush, new RectangleF(pt.X+1, pt.Y+1, size.Width, size.Height));
+					else
+						e.Graphics.DrawString(this.Text,this.Font, solidBrush, pt.X + 1, pt.Y + 1);
 				}
 			}
 				//
@@ -690,7 +706,10 @@ namespace SIL.Windows.Forms.Widgets
 				//
 				using(var solidBrush = new SolidBrush(this.ForeColor))
 				{
-					e.Graphics.DrawString(this.Text,this.Font, solidBrush,pt.X,pt.Y);
+					if (TextWordWrap && !TextDropShadow)
+						e.Graphics.DrawString(this.Text, this.Font, solidBrush, new RectangleF(pt.X, pt.Y, size.Width, size.Height));
+					else
+						e.Graphics.DrawString(this.Text,this.Font, solidBrush,pt.X,pt.Y);
 				}
 			}
 		}
@@ -828,6 +847,17 @@ namespace SIL.Windows.Forms.Widgets
 			System.Drawing.SizeF size = g.MeasureString(strText,font);
 			return (size);
 		}
+		/// <summary>
+		/// Calculate the size that is best for wrapping text to fit the available width.
+		/// </summary>
+		private System.Drawing.SizeF CalculateBestWrappingSize(Graphics g)
+		{
+			// Try to wrap with the maximum available width.  This may cause words to split in the
+			// middle, or more than two lines to be displayed.  But that's still better than showing
+			// partial data in one line.
+			return g.MeasureString(this.Text, this.Font, this.Width);
+		}
+
 		/// <summary>
 		/// Calculates the rectangular region used for text display.
 		/// </summary>


papeh:
Reviewed 1 of 1 files at r1.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BiKxMayFXjcJRhy9E:-L4BiKxMayFXjcJRhy9F:b6zd3ve) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L106)):*
> ```Smalltalk
> 		{
> 			get{return _TextWordWrap;}
> 			set{_TextWordWrap = value;}
> ```

I know everybody else is doing it this way, but is there a good reason not to make this an auto-implement property?

If you choose not to auto-imp, perhaps we should `get { return _TextWordWrap && !TextDropShadow; }`, since that's the only way it's used, and is consistent with the comment.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 863 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoZ75MzAOdT7ychmh:-L4BoZ75MzAOdT7ychmi:b6v47el) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L863)):*
> ```Smalltalk
> 			// We need to wrap if possible.  Try each space as a possible break point.
> 			SizeF sizePrev = sizeOneLine;
> 			for (int idx = this.Text.IndexOf(' '); idx > 0; idx = this.Text.IndexOf(' ', idx+1))
> ```

`idx > 0` will not work with leading spaces

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 874 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4Bpx1Da0NqmCe7oQbW:-L4Bpx1Da0NqmCe7oQbX:brrn943) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L874)):*
> ```Smalltalk
> 					// quit.  If we have a previous splitting point that looks better, use it.
> 					// Otherwise, measure with this splitting point's width and hope for the
> 					// best.
> ```

I don't think we should return until we have tried each splitting point.  The best place to wrap is usually the first space after the midpoint. Take for example, "I congratulate my students": The first wrapping point may be too wide, but the second will provide a narrower size.
I recommend something similar to
```
if (sizeWrap.Width >= sizePrev.Width) // we have passed the middle of the string, so sizes are getting wider
{
  if (sizeWrap.Width < Width) // wrapping beyond the midpoint fits
    return sizeWrap;
  else
    return sizePrev;
}
```


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598)*

JohnThomson:

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 103 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoBbPSgyKkDnuvPeZ:-L4BoBbPSgyKkDnuvPe_:bbi5vgs) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L103)):*
> ```Smalltalk
> 			System.ComponentModel.RefreshProperties(RefreshProperties.Repaint)
> 		]
> 		public bool TextWordWrap
> ```

Is there a reason for the backing field? Why not just public bool TextWordWrap {get; set;}?

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 863 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoZ75MzAOdT7ychmh:-L4BwYBAhi8T2NLK0nEp:bfzl2mf) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L863)):*

`idx > 0` will not work with leading spaces
</blockquote></details>

It seems to me that this algorithm could benefit from unit tests. I find myself struggling to figure out whether it will do the right thing with trailing spaces as well as leading ones, multiple adjacent spaces, newlines. What will/should happen if the first space results in a size wider than the button? Should we make an arbitrary break...so at least all the text can be seen...if there is no good break?
If this weren't for localization, most of these could be ignored...the programmer would fix any input that gave a bad-looking button. But this is for localization, so it probably won't be tested before it ships.
Are there non-Roman issues we should think about? For example, IIRC Thai and possibly Chinese don't use spaces. Do languages like Arabic have any other characters that signal a good place to break?
Another option to think about: I believe the Windows.Forms API has an overload of DrawString (or it might be DrawText) that wraps the text to fit a given rectangle. Look for one that takes an options argument, one of which is for wrapping. Could you just use that with a rectangle that is the width of the button and two lines high, and let the system figure out a good break point? e.g., https://docs.microsoft.com/en-us/dotnet/framework/winforms/advanced/how-to-draw-wrapped-text-in-a-rectangle. (Or maybe that's what you're doing...this code just finds a better rectangle. But consider the options for TextFormatFlags. If WordBreak doesn't give a useful split, there may be another option that is some help.)
Another option would to use a binary split to find the smallest width at which the text can be drawn in two lines.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 884 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4C-0H4Vls_tlfmSCDh:-L4C-0H4Vls_tlfmSCDi:b-c9vvhu) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L884)):*
> ```Smalltalk
> 			if (sizePrev != sizeOneLine)
> 				return g.MeasureString(this.Text, this.Font, 1+(int)sizePrev.Width);
> 			return sizeOneLine;
> ```

Would it do any better on strings that can't be split at spaces if we returned a rectangle the width of the button and two lines high?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598)*

JohnThomson:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598)*

StephenMcConnel:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 103 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoBbPSgyKkDnuvPeZ:-L4C4iG9rx_IxoFPQH_P:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L103)):*

Is there a reason for the backing field? Why not just public bool TextWordWrap {get; set;}?
</blockquote></details>

Done.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BiKxMayFXjcJRhy9E:-L4C4n6FY2HwzUlt1RFm:b-p8txt2) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L106)):*

I know everybody else is doing it this way, but is there a good reason not to make this an auto-implement property?

If you choose not to auto-imp, perhaps we should `get { return _TextWordWrap && !TextDropShadow; }`, since that's the only way it's used, and is consistent with the comment.
</blockquote></details>

made it auto-implement.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 863 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoZ75MzAOdT7ychmh:-L4CBiLwfJkazRg766H8:bj4ogoc) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L863)):*

It seems to me that this algorithm could benefit from unit tests. I find myself struggling to figure out whether it will do the right thing with trailing spaces as well as leading ones, multiple adjacent spaces, newlines. What will/should happen if the first space results in a size wider than the button? Should we make an arbitrary break...so at least all the text can be seen...if there is no good break?
If this weren't for localization, most of these could be ignored...the programmer would fix any input that gave a bad-looking button. But this is for localization, so it probably won't be tested before it ships.
Are there non-Roman issues we should think about? For example, IIRC Thai and possibly Chinese don't use spaces. Do languages like Arabic have any other characters that signal a good place to break?
Another option to think about: I believe the Windows.Forms API has an overload of DrawString (or it might be DrawText) that wraps the text to fit a given rectangle. Look for one that takes an options argument, one of which is for wrapping. Could you just use that with a rectangle that is the width of the button and two lines high, and let the system figure out a good break point? e.g., https://docs.microsoft.com/en-us/dotnet/framework/winforms/advanced/how-to-draw-wrapped-text-in-a-rectangle. (Or maybe that's what you're doing...this code just finds a better rectangle. But consider the options for TextFormatFlags. If WordBreak doesn't give a useful split, there may be another option that is some help.)
Another option would to use a binary split to find the smallest width at which the text can be drawn in two lines.
</blockquote></details>

I verified that a simple measurement with the width limited to what is available works as well or better as this complicated loop (which likely wasn't quite correct anyway).

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 874 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4Bpx1Da0NqmCe7oQbW:-L4CBvch--ONDfPcz0-_:b-x5ymmc) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L874)):*

I don't think we should return until we have tried each splitting point.  The best place to wrap is usually the first space after the midpoint. Take for example, "I congratulate my students": The first wrapping point may be too wide, but the second will provide a narrower size.
I recommend something similar to
```
if (sizeWrap.Width >= sizePrev.Width) // we have passed the middle of the string, so sizes are getting wider
{
  if (sizeWrap.Width < Width) // wrapping beyond the midpoint fits
    return sizeWrap;
  else
    return sizePrev;
}
</blockquote></details>

no longer relevant with simplified one-line method.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 884 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4C-0H4Vls_tlfmSCDh:-L4C1Irf-yO-9zlilOVy:b-5c7h0m) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L884)):*

Would it do any better on strings that can't be split at spaces if we returned a rectangle the width of the button and two lines high?
</blockquote></details>

I've changed the code in such a way that it might split words in the middle if it has to, or maybe show three lines instead of two.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598)*
<!-- Sent from Reviewable.io -->

StephenMcConnel:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 863 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoZ75MzAOdT7ychmh:-L4CEvBs-nAR-2M4xB7o:b9mddov) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L863)):*

I verified that a simple measurement with the width limited to what is available works as well or better as this complicated loop (which likely wasn't quite correct anyway).
</blockquote></details>

I haven't figured out how to do unit tests on this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598)*

JohnThomson:
:lgtm:.  Since Hasso also had some very good comments, I'll let him take another look, too.

I wish we could make the new behavior the default...it's hard to imagine it NOT being desirable. But I suppose that would be a breaking change? Is there somewhere we could document that it would be a good breaking change the next time we make a release at that level?

It's also tempting to regard the old behavior...drawing the text clipped...as a bug and just fix it without adding a new property to control it.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-:-L4CSYLo-SbNp8Jg0Ttm:b9r9cwl)*

StephenMcConnel:
I did it this way because the original author used ShadowedText by default, and I'm not at all sure how word wrap would interact with shadowed text, so I made them mutually exclusive.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-:-L4CTqVoq2LMismywc2S:bruft45)*

papeh:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 863 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4BoZ75MzAOdT7ychmh:-L4CooyFjCoP9i22LVs7:b-ku93d7) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L863)):*

I haven't figured out how to do unit tests on this.
</blockquote></details>

Now, since the method is down to one line and one caller, does it really deserve to be a separate method?

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 352 at r2](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-L4CnzQqel19IUdcr7vF:-L4CnzQqel19IUdcr7vG:bsvl902) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9cac383292f9f416ad8510c6201b2f774af56d3/SIL.Windows.Forms/Widgets/BitmapButton.cs#L352)):*
> ```Smalltalk
> 			ImageAttributes = new ImageAttributes();
> 			DisabledTextColor = System.Drawing.Color.DimGray;
> 			TextWordWrap = false;
> ```

theoretically not necessary

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/598#-:-L4Cp8bz8x7rKgRJ_bJ-:bnfp4nl)*

StephenMcConnel:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 863 at r1](https://beta.reviewable.io/reviews/sillsdev/libpalaso/598#-L4BoZ75MzAOdT7ychmh:-L4CpYxX3Uz26v_uHvfB:btkz4cr) ([raw file](https://github.com/sillsdev/libpalaso/blob/f945afc8d81704a46ba63c8c7a0b2cdfcbb0d573/SIL.Windows.Forms/Widgets/BitmapButton.cs#L863)):*

Now, since the method is down to one line and one caller, does it really deserve to be a separate method?
</blockquote></details>

The other method for measuring text (just above this one) is also called only once and contains only one line.   So I figured to maintain the pattern.  Having a semi-meaningful method name and a method comment also help to document what is going on.

---

*[SIL.Windows.Forms/Widgets/BitmapButton.cs, line 352 at r2](https://beta.reviewable.io/reviews/sillsdev/libpalaso/598#-L4CnzQqel19IUdcr7vF:-L4CpR7jAjqB-iiuxpOo:b-soolth) ([raw file](https://github.com/sillsdev/libpalaso/blob/d9cac383292f9f416ad8510c6201b2f774af56d3/SIL.Windows.Forms/Widgets/BitmapButton.cs#L352)):*

theoretically not necessary
</blockquote></details>

True, but I figured being explicit doesn't hurt future maintainers.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/libpalaso/598)*

PR #600, opened 2018-01-31
BL-5416 Improve RobustIO.LoadXElement NR path names

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/600)

### DIFF
--- a/SIL.Core.Tests/IO/RobustIOTests.cs
+++ b/SIL.Core.Tests/IO/RobustIOTests.cs
@@ -5,8 +5,10 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using System.Xml;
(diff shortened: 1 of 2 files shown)


hatton:
Looks good, and I don't suppose these failed tests are related, but let's see if that gets sorted first.

I tend to think this method you fixed doesn't belong in this class, btw. But at least now it will be fixed.

andrew-polk:
The failing tests were fixed in another PR. But if we want to see this PR build successfully, we need to rebase on that change.

JohnThomson:
I believe I've successfully rebased the change...do we need to do something to start a new test build?

andrew-polk:
They started automatically. And all relevant builds succeeded.

PR #602, opened 2018-02-08
Add GuidProvider and DateTimeProvider

ermshiperete:
Together with `ReproducibleDateTimeProvider` and `ReproducibleGuidProvider` this helps in testing by allowing to get specific date/time or GUID in the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/602)

### DIFF
--- a/Palaso.TestUtilities/Palaso.TestUtilities.csproj
+++ b/Palaso.TestUtilities/Palaso.TestUtilities.csproj
@@ -127,6 +127,8 @@
     <Compile Include="FluentAssertXmlTests.cs" />
     <Compile Include="LdmlContentForTests.cs" />
     <Compile Include="LiftContentForTests.cs" />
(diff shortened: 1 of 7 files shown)


andrew-polk:
:lgtm:

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/602#-:-L4urRsLabt8WMmP2x8U:bnfp4nl)*

ermshiperete:
@jasonleenaylor  The DateTimeProvider and GuidProvider are expected to be used in production code - when we want a method to be testable and be able to produce a defined result we'd use e.g. DateTimeProvider.Current.Now instead of DateTime.Now. In the unit test we can set up things so that ReproducibleDateTimeProvider gets used and then we know the exact output to expect from the method under test.

PR #618, opened 2018-02-23
Convert SIL.Scripture to .NET Standard 2.0

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/618)


SIL.Scripture/SIL.Scripture.csproj
```diff
@@ -1,421 +1,21 @@
...
-      <RequiredTargetFramework>3.5</RequiredTargetFramework>
-    </Reference>
-    <Reference Include="System.Data" />
-    <Reference Include="System.Xml" />
-  </ItemGroup>
-  <ItemGroup>
-    <Compile Include="..\GlobalAssemblyInfo.cs">
-      <Link>Properties\GlobalAssemblyInfo.cs</Link>
-    </Compile>
-    <Compile Include="BCVRef.cs" />
```
jasonleenaylor: Ok, so by what magic are the .cs files included in this project?

andrew-polk: No idea. It must just include all .cs files. Yay!

jasonleenaylor:
So follow this thought experiment with me: Working on a new feature you checkout a new branch, modify an interface, and add a new class that uses the interface. Now you want to go back to master to fix a bug so you checkout master and build. What happens? I was asking @ddaspit the same question on his PR.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L6NiQDSEFtx6cB63VIA:bnvcib0)*

JohnThomson:
In my experience, going back to master will not delete the new file in which (I presume) you created the new class. However, it will revert the .csproj file which lists the files to compile, so the new file won't be in the build, and thus its reference to the modified interface won't be a problem, and all will be well except that you will be in some danger of accidentally checking in the new class as part of your bug fix.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L6Np-ObLrDyqiQWlTo8:b-uerv9j)*

jasonleenaylor:
@johnthomson Right, but in this PR the .csproj file doesn't list any files to compile. So in the future it would try and build any .cs files it found giving confusing errors. So either developers would have to get into a habit of running git clean -fdx (which is a good way of losing uncommitted new files) or we will be periodically having to deal with this confusion.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L6NrfSfh63HKLV50q41:b2t4bdv)*

jasonleenaylor:
So https://docs.microsoft.com/en-us/dotnet/core/tools/csproj explains the difference. I think this is going to bite us on branch changes. There is a way to shut off the new default behavior it looks like. What do you guys think? @ddaspit @andrew-polk ?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L6Nt1BoYIInQpudDvpC:b8bbahv)*

andrew-polk:
If I'm correct, you're talking about creating a new file in one branch, not committing, switching to another branch, and desiring the new file which belongs to the first branch not have any effect. Doesn't seem like an important scenario to me, personally.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L6NxS4JKBkt7EzgrC7n:b8v6sla)*

ermshiperete:
I just tried. If the new file in the branch is committed to git and then you switch to a different branch, git will remove the new file from the working directory. I think git doesn't remove the files if they are not checked in or if they are ignored. I'd vote that we try the new behaviour - makes the csproj file much shorter.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L6NwjqZztOfjbo_WMMa:b-1fro7o)*

jasonleenaylor:
OK, If this doesn't cause any trouble for normal branch switching scenarios then there is no need to use special tags to keep the file list. I know that I've seen files left behind before, maybe it was an old version of git.

ddaspit:
I didn't explicitly close this. It must have closed when I merged in my PR, I'm not sure why.

andrew-polk:
Likely got deleted if the netstandard branch was deleted (which this was targetting). I retargeted to master.

andrew-polk:
Going to take some work to clean this up... probably won't happen right away.

ddaspit:
LGTM. Make sure that all of the different configurations (x86, x64, Mixed Platforms, Any CPU) are setup correctly. You also might want to put a description in SIL.Scripture.csproj, so that the NuGet package has a description.

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L7giJTy-E1QUOzhbClu:bg66jxr)*

andrew-polk:
> Make sure that all of the different configurations (x86, x64, Mixed Platforms, Any CPU) are setup correctly.

Double checked.

> You also might want to put a description in SIL.Scripture.csproj, so that the NuGet package has a description.

Done.

---

Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/618#-:-L7isKaU2K1usVh2Onl4:b-ft9fi7)*

PR #620, opened 2018-02-23
Keep GTK file chooser on top of program window and calling dialog

StephenMcConnel:
The code also prevents the user from accidentally (or maliciously) invoking multiple file
choosers at the same time.  The GTK dialog doesn't seem to want to behave as a true
modal dialog with respect to the Mono/C# program.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/620)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs
+++ b/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs
@@ -45,43 +45,55 @@ namespace SIL.Windows.Forms.ImageToolbox
 				ImageChanged(sender, e);
 		}
 
+		private bool _forceModal;
+
 		private void OnGetFromFileSystemClick(object sender, EventArgs e)
 		{
-			SetMode(Modes.SingleImage);
-			// The primary thing that OpenFileDialogWithViews buys us is the ability to default to large icons.
-			// OpenFileDialogWithViews still doesn't let us read (and thus remember) the selected view.
-			using (var dlg = new OpenFileDialogWithViews(OpenFileDialogWithViews.DialogViewTypes.Large_Icons))
+			if (_forceModal)
+				return;		// The GTK file chooser on Linux isn't acting modal, so we simulate it.
+			_forceModal = true;
+			try
 			{
-				if (string.IsNullOrEmpty(ImageToolboxSettings.Default.LastImageFolder))
+				SetMode(Modes.SingleImage);
+				// The primary thing that OpenFileDialogWithViews buys us is the ability to default to large icons.
+				// OpenFileDialogWithViews still doesn't let us read (and thus remember) the selected view.
+				using (var dlg = new OpenFileDialogWithViews(OpenFileDialogWithViews.DialogViewTypes.Large_Icons))
 				{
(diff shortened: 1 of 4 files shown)


StephenMcConnel:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs, line 55 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/620#-L63he75-GUFzEjO7fkK:-L63he75-GUFzEjO7fkL:bkfdd87) ([raw file](https://github.com/sillsdev/libpalaso/blob/9a4cee8bb732eac9e318e594c5a69007dcf40b47/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs#L55)):*
> ```Smalltalk
> 				return;		// The GTK file chooser on Linux isn't acting modal, so we simulate it.
> 			_forceModal = true;
> 			try
> ```

The only actual change here is adding an outer try/finally block to ensure that _forceModal gets reset to false when the method finishes.  Indenting the code by 1 tab seems to confuse these fancy diff displays no end.  (They just aren't fancy enough!)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/620)*

JohnThomson:
Reviewed 4 of 4 files at r1.

---

*[SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs, line 48 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/620#-L6HYYfTknxwjs_u-2qc:-L6HYYfTknxwjs_u-2qd:b-9dwhif) ([raw file](https://github.com/sillsdev/libpalaso/blob/9a4cee8bb732eac9e318e594c5a69007dcf40b47/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs#L48)):*
> ```Smalltalk
> 		}
> 
> 		private bool _forceModal;
> ```

How about 'dialogIsOpen'? The current name suggests to me an option for whether to make it modal or not.

---

*[SIL.Windows.Forms/ImageToolbox/OpenFileDialogWithViews.cs, line 88 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/620#-L6HYr8pVV_O0WMPmCUo:-L6HYr8pVV_O0WMPmCUp:bn6mzpl) ([raw file](https://github.com/sillsdev/libpalaso/blob/9a4cee8bb732eac9e318e594c5a69007dcf40b47/SIL.Windows.Forms/ImageToolbox/OpenFileDialogWithViews.cs#L88)):*
> ```Smalltalk
> 					if (DialogAdapters.CommonDialogAdapter.UseGtkDialogs)
> 					{
> 						DialogAdapters.CommonDialogAdapter.WindowType = Platform.IsCinnamon ?
> ```

Might deserve a comment? One thing I found myself wondering was, is Cinnamon definitely the one and only flavor of Linux that requires this, or just the only one out of a handful we test with?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/620)*

StephenMcConnel:
The original change this is based on has been reverted, so I'm closing this PR.

PR #633, opened 2018-03-06
Tweak SuperToolTip to not constantly create/destroy windows

StephenMcConnel:
See https://issues.bloomlibrary.org/youtrack/issue/BL-5705 for the
motivation of this change.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/633)

### DIFF
--- a/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs
+++ b/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs
@@ -168,14 +168,7 @@ namespace SIL.Windows.Forms.SuperToolTip
 			if (controlTable[owner].UseSuperToolTip)
 			{
 				CloseTooltip();
-				if (_useFadding)
-				{
-					FadeIn();
-				}
-
 				winData.SuperInfo = controlTable[owner].SuperToolTipInfo;
-				window.Size = winData.Size;
-
 				if (winData.SuperInfo.OffsetForWhereToDisplay != default(Point))
 				{
 					window.Location = owner.PointToScreen(winData.SuperInfo.OffsetForWhereToDisplay);
@@ -184,7 +177,10 @@ namespace SIL.Windows.Forms.SuperToolTip
 				{
 					window.Location = owner.PointToScreen(new Point(0, owner.Height));
 				}
-
+				if (_useFadding)
+				{
+					FadeIn();
+				}
 				winData.GotFocus += OnGotFocus;
 				window.Show(owner);
 				winData.GotFocus -= OnGotFocus;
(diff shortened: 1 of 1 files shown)


hatton:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/633#-L6x7DQI_yys6LKXdZaX:-L6x7DQI_yys6LKXdZaY:b3yv6zn) ([raw file](https://github.com/sillsdev/libpalaso/blob/6ae110b4a74d12d71e4bd2ee4ebc782f537a2429/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs#L216)):*
> ```Smalltalk
> 					window.SizeChanged += CheckSizeChange;
> 			}
> 			window.MouseLeave += OnToolTipMouseLeave;
> ```

So... i gather you found a scenario where window was not null. If it was not, does that mean we end up with two SizeChanged event handlers, and two entries of winData in window.Controls? Is that a potential problem? Might warrant a comment-answer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/633)*

StephenMcConnel:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/633#-L6x7DQI_yys6LKXdZaX:-L6xHKoB4HAHFpm3yW7V:b-9aiad5) ([raw file](https://github.com/sillsdev/libpalaso/blob/6ae110b4a74d12d71e4bd2ee4ebc782f537a2429/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs#L216)):*

So... i gather you found a scenario where window was not null. If it was not, does that mean we end up with two SizeChanged event handlers, and two entries of winData in window.Controls? Is that a potential problem? Might warrant a comment-answer.
</blockquote></details>

I've tweaked the code a bit more to remove redundancy and maybe make it clearer.  I also added some comments to the method and the preceding one, and renamed this method slightly.  The window is null initially, and this method is used to create it as well as to reinitialize it each time the tooltip is closed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/633)*

hatton:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/633#-L6x7DQI_yys6LKXdZaX:-L6xU9sxkmm5fE83IO1M:b-24vnj1) ([raw file](https://github.com/sillsdev/libpalaso/blob/6ae110b4a74d12d71e4bd2ee4ebc782f537a2429/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs#L216)):*

I've tweaked the code a bit more to remove redundancy and maybe make it clearer.  I also added some comments to the method and the preceding one, and renamed this method slightly.  The window is null initially, and this method is used to create it as well as to reinitialize it each time the tooltip is closed.
</blockquote></details>

So, you're expecting that winData has always been Dispose()'d of by the time you reassign it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/633)*

StephenMcConnel:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/633#-L6x7DQI_yys6LKXdZaX:-L6xUjKD4X0xjqOlfS5M:b-7hhpq1) ([raw file](https://github.com/sillsdev/libpalaso/blob/6ae110b4a74d12d71e4bd2ee4ebc782f537a2429/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs#L216)):*

So, you're expecting that winData has always been Dispose()'d of by the time you reassign it?
</blockquote></details>

Yes.  The CloseTooltip() method explicitly calls winData.Dispose() (and then sets the variable to null) before calling CreateOrInitTooltipWindow().

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/633)*

hatton:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/633#-L6x7DQI_yys6LKXdZaX:-L6xWvoQiff2QqQAOMRd:b-1gf3l6) ([raw file](https://github.com/sillsdev/libpalaso/blob/6ae110b4a74d12d71e4bd2ee4ebc782f537a2429/SIL.Windows.Forms/SuperToolTip/SuperToolTip.cs#L216)):*

Yes.  The CloseTooltip() method explicitly calls winData.Dispose() (and then sets the variable to null) before calling CreateOrInitTooltipWindow().
</blockquote></details>

Ok, put another way, you're expecting that we can't have 2 CreateTooltipWindow() calls without an intervening CloseTooltip(). Fine by me, if that's correct.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/633)*

PR #646, opened 2018-03-13
St382 reports blank Genre correctly by removing angle brackets

gtryus:
Angle brackets around <Unknown> were preventing the use of the item from the controlled vocabulary

Card: https://trello.com/c/9eUfILJG

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/646)

### DIFF
--- a/SIL.Archiving.Tests/IMDI30Tests.cs
+++ b/SIL.Archiving.Tests/IMDI30Tests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Globalization;
 using System.Linq;
 using NUnit.Framework;
@@ -305,6 +305,14 @@ namespace SIL.Archiving.Tests
 			Assert.IsNull(countries.FindByText("Unknown"));
 			Assert.IsNull(countries.FindByText("Unspecified"));
 			Assert.IsNull(countries.FindByText("Undefined"));
+		}
+
+		[Test]
+		public void SetSessionGenre_ContainsAngleBrackets_LatinOnly()
+		{
+			var session = new Session();
+			session.Genre = "<Unknown>";
+			Assert.AreEqual("Unknown", session.Genre);
 		}
 	}
 }
--- a/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs
+++ b/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
@@ -1870,7 +1870,7 @@ namespace SIL.Archiving.IMDI.Schema
 			}
 			set
 			{
-				MDGroup.Content.Genre = value.ToVocabularyType(false, ListType.Link(ListType.ContentGenre));
+				MDGroup.Content.Genre = value.Replace("<","").Replace(">","").ToVocabularyType(false, ListType.Link(ListType.ContentGenre));
 			}
 		}
 


hatton:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Archiving/SIL.Archiving.csproj, line 299 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/646#-L7WXyzPKCMGZ1TjDGLI:-L7WXyzPKCMGZ1TjDGLJ:b-tppetz) ([raw file](https://github.com/sillsdev/libpalaso/blob/9522e96ecde67e256d1c1ffff6da8293d410fb38/SIL.Archiving/SIL.Archiving.csproj#L299)):*
> ```XML
>   </ItemGroup>
>   <ItemGroup>
>     <Reference Include="Enchant.Net, Version=1.4.3.1, Culture=neutral, processorArchitecture=MSIL">
> ```

l've asked on Slack about this, it is presumably required due to some change in libpalaso. Maybe they will fix that? Be a drag if all apps now have to start shipping a spelling checker even if they don't support spell checking?

---

*[SIL.Archiving/IMDI/Schema/IMDI_3_0.cs, line 1873 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/646#-L7WY_O9t6tfTcEXF4mw:-L7WY_O9t6tfTcEXF4mx:bxjqm7t) ([raw file](https://github.com/sillsdev/libpalaso/blob/9522e96ecde67e256d1c1ffff6da8293d410fb38/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs#L1873)):*
> ```Smalltalk
> 			set
> 			{
> 				MDGroup.Content.Genre = value.Replace("<","").Replace(">","").ToVocabularyType(false, ListType.Link(ListType.ContentGenre));
> ```

Instead of removing characters, or even dealing with these as they come up, it would be good if the code instead run *all* outputs through an xml encoder? https://stackoverflow.com/questions/1132494/string-escape-into-xml

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/646)*

gtryus:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Archiving/SIL.Archiving.csproj, line 299 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/646#-L7WXyzPKCMGZ1TjDGLI:-L7_-eel-bZkgIXQqDxH:bikcw24) ([raw file](https://github.com/sillsdev/libpalaso/blob/9522e96ecde67e256d1c1ffff6da8293d410fb38/SIL.Archiving/SIL.Archiving.csproj#L299)):*

l've asked on Slack about this, it is presumably required due to some change in libpalaso. Maybe they will fix that? Be a drag if all apps now have to start shipping a spelling checker even if they don't support spell checking?
</blockquote></details>

Yes, I was discussing this with Jason this morning too. I am not totally sure why it wouldn't build without adding this dependency. I wondered if it was related to the comment from Damian: "You need to use git clean". I had done a rebuild from Visual Studio but perhaps that doesn't rebuild everything. I had to upgrade to VS2017 and I had to add C to it as this was not installed by default but was needed for ForceAbort (or some such C program that seemed to be added to SIL library). So in the end, I got it to build and at one point it was complaining that it wouldn't build without this but I am not sure if it is needed by everyone using SIL.Windows.Forms or not. It is a SIL.Windows.Forms dependency.

---

*[SIL.Archiving/IMDI/Schema/IMDI_3_0.cs, line 1873 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/646#-L7WY_O9t6tfTcEXF4mw:-L7_0dOMHzdicbCXV91H:bubexvp) ([raw file](https://github.com/sillsdev/libpalaso/blob/9522e96ecde67e256d1c1ffff6da8293d410fb38/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs#L1873)):*

Instead of removing characters, or even dealing with these as they come up, it would be good if the code instead run *all* outputs through an xml encoder? https://stackoverflow.com/questions/1132494/string-escape-into-xml
</blockquote></details>

yeah, as I commented on the card, I don't think this will work. The problem is the UI puts these angle cahracters in and tries to use them in the lookup. I could put the code in the "ToVocabulary" logic where they would apply to all attempts on mapping to vocabulary but I think this is the only place users have reported that they are needed. So I didn't know if it made sense to create a general solution for a specific problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/646)*

gtryus:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Archiving/SIL.Archiving.csproj, line 299 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/646#-L7WXyzPKCMGZ1TjDGLI:-L7_JKGAO6I1IWeOKmbG:bwlfq6y) ([raw file](https://github.com/sillsdev/libpalaso/blob/9522e96ecde67e256d1c1ffff6da8293d410fb38/SIL.Archiving/SIL.Archiving.csproj#L299)):*
<details><summary><i>Previously, gtryus wrote…</i></summary><blockquote>

Yes, I was discussing this with Jason this morning too. I am not totally sure why it wouldn't build without adding this dependency. I wondered if it was related to the comment from Damian: "You need to use git clean". I had done a rebuild from Visual Studio but perhaps that doesn't rebuild everything. I had to upgrade to VS2017 and I had to add C to it as this was not installed by default but was needed for ForceAbort (or some such C program that seemed to be added to SIL library). So in the end, I got it to build and at one point it was complaining that it wouldn't build without this but I am not sure if it is needed by everyone using SIL.Windows.Forms or not. It is a SIL.Windows.Forms dependency.
</blockquote></details>

I removed these lines from the commit because using git clean -dxf and rebuilding worked without these modifications. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/646)*

PR #648, opened 2018-03-14
Remove unnecessary package from 4.6.1 version

ermshiperete:
The 4.6.1 version of SIL.Core doesn't need the
System.Runtime.InteropServices.RuntimeInformation nuget package.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/648)


ermshiperete:
Stupid compiler on my machine - it build successfully on my machine even though the dependency is still needed!

PR #649, opened 2018-03-14
if induce or simulator in stderr then not error

glasseyes:
chorus 2.6 builds found that the exception induced by InduceChorusFailure in FailureSimulator as part of a test was being detected as a false positive on stderr. So, with this change, if induce or simulator is found on stderr then log as a warning instead of error

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/649)


ermshiperete:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/649#-:-L7YkN9T7xr-jCH1Wap5:bnfp4nl)*

PR #653, opened 2018-03-14
St384 - Allow calling programs access to the IMDI file object

gtryus:
Allow calling programs to access the file object so they
can attach additional information such as recording equipment

card: https://trello.com/c/IlOdLSNy

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/653)

### DIFF
--- a/SIL.Archiving.Tests/IMDI30Tests.cs
+++ b/SIL.Archiving.Tests/IMDI30Tests.cs
@@ -320,8 +320,8 @@ namespace SIL.Archiving.Tests
 		{
 			var session = new Session();
 			var number = session.MDGroup.Content.Keys.Key.Count;
-			session.AddKeyValuePair("keyword", "holiday");
-			session.AddKeyValuePair("keyword", "emotion");
+			session.AddContentKeyValuePair("keyword", "holiday");
+			session.AddContentKeyValuePair("keyword", "emotion");
 			Assert.AreEqual(number + 2, session.MDGroup.Content.Keys.Key.Count);
 		}
 	}
--- a/SIL.Archiving.Tests/IMDI30Tests.cs
+++ b/SIL.Archiving.Tests/IMDI30Tests.cs
@@ -314,5 +314,15 @@ namespace SIL.Archiving.Tests
 			session.Genre = "<Unknown>";
 			Assert.AreEqual("Unknown", session.Genre);
 		}
+
+		[Test]
+		public void SessionAddKeyValuePair_TwoKeywords_numberUp2()
+		{
+			var session = new Session();
+			var number = session.MDGroup.Content.Keys.Key.Count;
+			session.AddKeyValuePair("keyword", "holiday");
+			session.AddKeyValuePair("keyword", "emotion");
+			Assert.AreEqual(number + 2, session.MDGroup.Content.Keys.Key.Count);
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


papeh:
:LGTM:

---

Reviewed 4 of 4 files at r1.

---

*[SIL.Archiving/IMDI/Schema/IMDI_3_0.cs, line 1885 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/653#-L83tNrKuNaDglIAy4Kl:-L83tNrKuNaDglIAy4Km:b-s9ugqq) ([raw file](https://github.com/sillsdev/libpalaso/blob/ddd660e23dfc301ae99a35c4bccf65ce14a9d27e/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs#L1885)):*
> ```Smalltalk
> 					return writtenResourceType;
> 			}
> 			return null;
> ```

I hope we don't plan to use this frequently over large sets of files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/653#-:-L83tWjPHEtSAXrNLj9F:bmxilld)*

gtryus:

---

*[SIL.Archiving/IMDI/Schema/IMDI_3_0.cs, line 1885 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/653#-L83tNrKuNaDglIAy4Kl:-L88DbGPnMokxf7YvInL:b-3th1g2) ([raw file](https://github.com/sillsdev/libpalaso/blob/ddd660e23dfc301ae99a35c4bccf65ce14a9d27e/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs#L1885)):*

I hope we don't plan to use this frequently over large sets of files.
</blockquote></details>

Not sure if you have a better way to propose. The issue is that SIL.ArchivingFiles are added to the session but in the adding process they are converted to a derived class IMDISessionFiles. There are a number (<10) that will be made to this method for each file as various fields are added to this IMDISessionFile that are not supported by SIL.ArchivingFile.

This was my second alternative solution. First I returned the IMDISessionFile from the add method but this made the interface for SIL.ArchivingFile return an IMDISessionFIle which was somewhat counter intuitive (and probably breaks some object oriented rule somewhere).

So for the sake of clarity and simplicity I pass in the file name from each method that needs it and search for the file in the list of files. It is simple to understand and fairly simple to maintain even if it isn't the most efficient way to do it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/653)*

jasonleenaylor:
*[SIL.Archiving/IMDI/Schema/IMDI_3_0.cs, line 1827 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/653#-L88Ic_MSBARquJO_bMw:-L88Ic_MSBARquJO_bMx:bhdtj45) ([raw file](https://github.com/sillsdev/libpalaso/blob/ddd660e23dfc301ae99a35c4bccf65ce14a9d27e/SIL.Archiving/IMDI/Schema/IMDI_3_0.cs#L1827)):*
> ```Smalltalk
> 
> 		/// <remarks/>
> 		public void AddGroupKeyValuePair(string key, string value)
> ```

value is an unfortunately chosen variable name, maybe in the future this could be refactored, but it doesn't need to hold anything up here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/653)*

PR #656, opened 2018-03-16
Convert SIL.DblBundle to .NET Standard 2.0 library

andrew-polk:


ddaspit:
Reviewed 7 of 7 files at r1.

---

*[SIL.DblBundle/Bundle.cs, line 62 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/656#-L8BLd2nswahK83l2Z_H:-L8BLd2nswahK83l2Z_I:b-pex457) ([raw file](https://github.com/sillsdev/libpalaso/blob/1df9c16790235466f4e41d89e32873e0466bdf73/SIL.DblBundle/Bundle.cs#L62)):*
> ```Smalltalk
> 	public static class Bundle
> 	{
> 		public static Func<string, string, string> GetLocalizedMessage = (id, englishString) => englishString;
> ```

Instead of using a static class, could you pass this function in using a constructor or property? I'm not sure how easy this would be, but it would make the code somewhat cleaner, since you wouldn't have to rely on a static. You could use a factory pattern to create the bundles if you don't want to pass in this function every time you create the bundle.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/656)*

andrew-polk:
*[SIL.DblBundle/Bundle.cs, line 62 at r1](https://beta.reviewable.io/reviews/sillsdev/libpalaso/656#-L8BLd2nswahK83l2Z_H:-L8cqXSk6RYPRTwt0_zm:bg1j79k) ([raw file](https://github.com/sillsdev/libpalaso/blob/1df9c16790235466f4e41d89e32873e0466bdf73/SIL.DblBundle/Bundle.cs#L62)):*
<details><summary><i>Previously, ddaspit (Damien Daspit) wrote…</i></summary><blockquote>

Instead of using a static class, could you pass this function in using a constructor or property? I'm not sure how easy this would be, but it would make the code somewhat cleaner, since you wouldn't have to rely on a static. You could use a factory pattern to create the bundles if you don't want to pass in this function every time you create the bundle.
</blockquote></details>

Thanks.
I decided to make a more generic solution. There were two classes in this assembly which needed a way to localize methods, and I notice that SIL.Archiving will need something similar. So I made a generic implementation in SIL.Core and an L10NSharp-specific one in SIL.Windows.Forms.
See what you think.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/libpalaso/656)*

ddaspit:
:lgtm:

---

Reviewed 7 of 7 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/libpalaso/656#-:-L8jat4a-qHeC9xAJdoN:bnfp4nl)*

PR #659, opened 2018-03-19
St389 - file names only allow one period

gtryus:
card: https://trello.com/c/P2knXht8

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/659)

### DIFF
--- a/SIL.Archiving.Tests/SimpleObjectTests.cs
+++ b/SIL.Archiving.Tests/SimpleObjectTests.cs
@@ -1,4 +1,4 @@
-﻿using System.Linq;
+using System.Linq;
 using NUnit.Framework;
 using SIL.Archiving.Generic;
 
@@ -30,6 +30,13 @@ namespace SIL.Archiving.Tests
 			Assert.AreEqual(2, hs.Count);
 			Assert.AreEqual(firstValue, hs.First(ls => ls.Iso3LanguageId == "eng").Value);
 			Assert.AreEqual(frenchValue, hs.First(ls => ls.Iso3LanguageId == "fra").Value);
+		}
+
+		[Test]
+		public void ToLatinOnly_MultiplePeriod_OnePeriod()
+		{
+			var result = "Langauge.wav.meta".ToLatinOnly("_", "+", ".");
+			Assert.AreEqual("Langauge_wav.meta", result);
 		}
 	}
 }
--- a/SIL.Archiving/Extensions.cs
+++ b/SIL.Archiving/Extensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Globalization;
 using System.Text;
 using System.Windows.Forms;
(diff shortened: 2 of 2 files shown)


SIL.Archiving/Extensions.cs
```diff
@@ -112,13 +112,17 @@ public static void AppendDelimiter(this StringBuilder sb, string value, string d
...
 
+			int i = 0;
 			foreach (var c in s)
 			{
 				var test = c.ToString(CultureInfo.InvariantCulture);
 
 				if ((test == charReplacement) || (test == spaceReplacement))  // do not replace the replacement characters
 					returnVal += test;
+				else if (c == '.' && i != lastDot) // only allow one dot in the file name
+					returnVal += spaceReplacement;
```
rmunn: Wouldn't it be better to use `charReplacement` rather than `spaceReplacement` for a period, for consistency with the rest of the logic in this function?

SIL.Archiving.Tests/SimpleObjectTests.cs
```diff
@@ -31,5 +31,12 @@ public void LanguageStringComparer_CorrectlyComparesLanguageStrings()
 			Assert.AreEqual(firstValue, hs.First(ls => ls.Iso3LanguageId == "eng").Value);
 			Assert.AreEqual(frenchValue, hs.First(ls => ls.Iso3LanguageId == "fra").Value);
 		}
+
+		[Test]
+		public void ToLatinOnly_MultiplePeriod_OnePeriod()
+		{
+			var result = "Langauge.wav.meta".ToLatinOnly("_", "+", ".");
+			Assert.AreEqual("Langauge+wav.meta", result);
```
rmunn: I would expect a `_` rather than a `+` here (see my comments on the `ToLatinOnly` function).

gtryus:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SIL.Archiving/Extensions.cs, line 125 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/659#-L7zqZuuKy0uTssn_EGG-r1-125:-L88G2aumf9NlSAXfSfG:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/5f188b0cb0875bc95cd89734623e6cb55886153b/SIL.Archiving/Extensions.cs#L125)):*
<details><summary><i>Previously, rmunn (Robin Munn) wrote…</i></summary><blockquote>

Wouldn't it be better to use `charReplacement` rather than `spaceReplacement` for a period, for consistency with the rest of the logic in this function? 
</blockquote></details>

Done.

---

*[SIL.Archiving.Tests/SimpleObjectTests.cs, line 39 at r1](https://reviewable.io:443/reviews/sillsdev/libpalaso/659#-L7zqZuuKy0uTssn_EGF-r1-39:-L88G3UD69QwLk9PPvhg:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/5f188b0cb0875bc95cd89734623e6cb55886153b/SIL.Archiving.Tests/SimpleObjectTests.cs#L39)):*
<details><summary><i>Previously, rmunn (Robin Munn) wrote…</i></summary><blockquote>

I would expect a `_` rather than a `+` here (see my comments on the `ToLatinOnly` function).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/libpalaso/659)*

PR #698, opened 2018-05-02
[linux] Fix location detection for RAMP (LT-16894)

ermshiperete:
RAMP installs in /opt/RAMP/ramp. LocateInProgramFiles() previously
didn't check subdirectories when running on Linux and so didn't find
RAMP even when installed. This change optionally checks in subdirectories.

This kind of changes API (or at least behavior) on Linux, but this
probably doesn't have noticeable effects for most projects.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/698)

### DIFF
--- a/Palaso.Tests/IO/FileLocatorTests.cs
+++ b/Palaso.Tests/IO/FileLocatorTests.cs
@@ -114,6 +114,77 @@ namespace Palaso.Tests.IO
 			Assert.DoesNotThrow(() => FileLocator.LocateInProgramFiles(findFile, true, "!~@blah"));
 		}
 
+		[Test]
+		[Platform(Include = "Linux")]
+		public void LocateInProgramFiles_DeepSearch_FindsFileInSubdir()
+		{
+			// This simulates finding RAMP which is installed as /opt/RAMP/ramp. We can't put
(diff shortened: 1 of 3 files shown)


jasonleenaylor:
Ok, so this is LGTM but why is it that the linux and windows implementations need to be different? The method comment says that the extra parameter for subfolders is illegal on linux but I don't understand why that would be so.

---







*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/698#-:-LBX8RqtALL1LLAhid1Q:b-h9q6rk)*

marksvc:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/698#-:-LBXO5eZ1abyQ9y2Tspe:bg7igk9)*

ermshiperete:
@jasonleenaylor That I don't understand either. I was wondering if I should just make the method behave the same on Windows and Linux, but that would make it an even more breaking change, although probably nobody would notice.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/libpalaso/698#-:-LBXWQYUE-YPVLcRUkVl:b17v7dj)*

PR #723, opened 2018-06-12
Fix exponential retry & delay problem with DeleteDirectoryAndContents

hatton:
We were getting hangs in Bloom when a multi-level directory had a file that couldn't be deleted. Turns out in February 2018 the libpalaso deletion code was changed to essentially require 40^(n-1) * 2 seconds wait, where n = level of nesting (n=1 if no child directories).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/723)

### DIFF
--- a/SIL.Core/IO/RobustIO.cs
+++ b/SIL.Core/IO/RobustIO.cs
@@ -41,6 +41,8 @@ namespace SIL.IO
 		{
 			// ReSharper disable EmptyGeneralCatchClause
 
+			var failedToDeleteAChildDirectory = false;
+
 			if (!Platform.IsWindows)
 			{
 				// The Mono runtime deletes readonly files and directories that contain readonly files.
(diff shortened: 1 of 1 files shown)


JohnThomson:
This is a big improvement. Couple of thoughts: how much do we care (a) about not delaying more than 2s, (b) trying hard to delete anything we can?
If a child directory fails, we will only try once to delete each file in the parent.
If multiple child directories fail, we will give them 2s each.
If there are a lot of files to try to delete, especially if several fail, each iteration might take more than 50ms.
One strategy: keep track of elapsed time since the method was called, and pass in a time limit. Instead of stopping after 40 iterations, stop when the time limit has elapsed. Allow child directories something like 10% of the remaining available time to try to delete themselves. (They might even get a second chance to try, if the outer loop completes in less than 2s.)
Probably not worth it...up to you. I'm fine with committing as is.

---

Review status: 0 of 1 files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/723#-:-LEpG4K6-Gy1Cnrd8xJw:b4ipuu0)*

JohnThomson:
Reviewed 1 of 1 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/723)*

hatton:
Yeah, I was thinking the same... this would be better if it took an argument of how long, total, to wait. But then I thought well, actually some directories with little mechanical disks could take a really long time, maybe you only want to specify some kind of "time spent retrying"... that's more complicated. Then I thought well, if it's really long, you should be using the OS's deletion dialog instead. And so forth. In the meantime, yeah I'd like just to fix this bug.

---

Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/723#-:-LEp_gjACzcaK2uPovG9:bwy8l7h)*

JohnThomson:
Fine with me. It's clearly a great improvement! There seem to be some merge conflicts or something, but once that's sorted out feel free to merge.

---

Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/723#-:-LEtshxo1NsUvE2oIX9R:bz0vzzm)*

PR #742, opened 2018-08-14
PTX-16877 make VerseRef a struct

jwickberg:
This is a breaking change, but class is not widely used.

We are changing VerseRef from a class to a struct to prevent problems with shared references to the same instance.

Minimal changes were done for this. The main changes involved using a new IsDefault property to check for a struct that hasn't been set since a null check can't be used.

The XmlSerializationHelper Deserialize methods were changed to support a struct. Needed to remove 'where class' restriction and use 'default(T)' rather than null.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/742)

### DIFF
--- a/SIL.Core/Xml/XmlSerializationHelper.cs
+++ b/SIL.Core/Xml/XmlSerializationHelper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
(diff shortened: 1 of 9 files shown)


andrew-polk:
___
*[SIL.Scripture/VerseRef.cs, line 155 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/742#-LJuTTeA5h55L6zNFM_u:-LJuTTeA5h55L6zNFM_v:b-ilp4ku) ([raw file](https://github.com/sillsdev/libpalaso/blob/2d20e43f7bdee7684bc9b55132aa8affe95e9067/SIL.Scripture/VerseRef.cs#L155)):*
> ```Smalltalk
> 		/// Checkes to see if a VerseRef hasn't been set - all values are the default.
> 		/// </summary>
> 		public bool IsDefault => bookNum == 0 && chapterNum == 0 && verseNum == 0 && versification == null;
> ```

The constructor initializes chapterNum and verseNum to -1 but here you check for 0.

jwickberg:
Due to git mistakes, I'm closing this and will create a new request.

PR #743, opened 2018-08-15
PTX-16877 make VerseRef a struct

jwickberg:
This is a a replacement for PR#742.

Updates from that request are:
1. A fix for a deserialization problem - verse ref without versification wasn't getting default versification.
2. Changed the signature of IScrVers.ChangeVersification to use a ref argument to make function like it used to.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/743)


jwickberg:
___
*[SIL.Scripture/IScrVers.cs, line 64 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/743#-LJytM0n3wcqbkL2_LeX:-LJyvtDDBqfJipAkLns6:b-gifhvr) ([raw file](https://github.com/sillsdev/libpalaso/blob/b15e088fa8ce52360c2b812ef593c78cd0a4db09/SIL.Scripture/IScrVers.cs#L64)):*

Not really a request, just more of a question...
Is it possible to make this use a ref as well? It would be more like the original code and also more consistent with `ChangeVersification`
</blockquote></details>

Yes - it would have been possible and be more consistent. If there are other changes, I'll do this one also.

jwickberg:
The main problem this change can create that won't be caught at compile time is a sequence like this:

methodA calls methodB passing a VerseRef
methodB changes the passed VerseRef
methodA expects to see the change

To get this to work, methodB has to be changed to have a ref argument

jwickberg:
___
*[SIL.Scripture/VerseRef.cs, line 1347 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/743#-LK2RAa8AMp1PYUCsFrM:-LK2iU1N77L05H4gDlVM:bailm1d) ([raw file](https://github.com/sillsdev/libpalaso/blob/1035d933b574c3b2e6ac7048c061b3527da9da67/SIL.Scripture/VerseRef.cs#L1347)):*

You don't have to do this as calling the method should create a copy.  May not be worth changing, though.
</blockquote></details>

I got an error when using "this" as the argument, so introduced the temp variable to get rid of the error.

jwickberg:
Pushed update with requested changes

PR #744, opened 2018-08-16
Don't reference NDesk.DBus from SIL.Core.Desktop

marksvc:
So when FW runs ILRepack on SIL.Core.Desktop it doesn't try to include
NDesk.DBus on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/744)


andrew-polk:
Looks like this will require some other changes as the Windows build fails.

marksvc:
Thanks Andrew. As I look at the list of TC builds it's not clear which one is actually relevant as it looks like none of them built from master. Which Windows build are you referring to? (Or do you mean that you tested this on a local Windows machine to determine this?) Thanks!

marksvc:
Haha, I didn't see the scroll bar on the list of build jobs. I found the failing master branch build :)

marksvc:
Cancelling PR. Not correct solution.

PR #745, opened 2018-08-22
Increased version number to 5.0

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/745)


JohnThomson:
Is this because of some breaking change we should all know about?

jwickberg:
The version number is being increased because of PR#743 - changing the VerseRef class to a struct.

The change only affects projects using SIL.Scripture, but best practice requires changing the major version number since this is a breaking change.

PR #746, opened 2018-08-23
BL-6370 BetterLabel uses TextRenderer functions for better NonRoman

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/746)


andrew-polk:
@hatton 
Do you think we need to hold off merging until we can test this out on Linux?
TextRenderer.DrawText is already used several places in libpalaso, but I don't know specifically if any of those are used in clients which run on Linux.

hatton:
Let's give it a go :-)

PR #748, opened 2018-08-27
Fix: LT-19356 FLEx crashes when trying to send in a suggestion

JamesPrabu:
* Changed the code to avoid the crash

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/748)


SIL.Core/Email/MAPIHelper.cs
```diff
@@ -92,7 +92,7 @@ IntPtr GetRecipients(out int recipCount)
 			int size = Marshal.SizeOf(typeof(MapiRecipDesc));
 			IntPtr intPtr = Marshal.AllocHGlobal(m_recipients.Count * size);
 
-			int ptr = (int)intPtr;
+			IntPtr ptr =intPtr;
```
jasonleenaylor: style mistake: space after =

PR #755, opened 2018-09-13
Make ArchivingLanguage get EnglishName from Ethnologue

JohnThomson:
If we don't have one from any other source.
Note: makes SIL.Archiving depend on SIL.WritingSystems

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/755)

### DIFF
--- a/SIL.Archiving/ArchivingLanguage.cs
+++ b/SIL.Archiving/ArchivingLanguage.cs
@@ -1,7 +1,8 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using SIL.WritingSystems;
 
 namespace SIL.Archiving
 {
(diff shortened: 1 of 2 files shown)


SIL.Archiving/ArchivingLanguage.cs
```diff
@@ -72,6 +73,15 @@ public string EnglishName
 					_englishName = returnVal;
 				}
 
+				if (string.IsNullOrEmpty(_englishName))
+				{
+					// Not very efficient, but this is not very performance-critical.
+					// And it's undesirable to crash if someone uses a language that's not windows-standard.
+					var lookup = new LanguageLookup();
+					var lang = lookup.GetLanguageFromCode(_iso3Code);
+					_englishName = lang?.DesiredName;
```
tombogle: Might want to add a comment noting that the DesiredName might not actually be an _English_ name (at least from what I can tell). I assume this won't be a problem (e.g., it won't be assumed that the name can be presented using a font that covers only the basic characters needed to display English text)?

PR #762, opened 2018-10-16
HT-290: Improved performance of log box when there is a high volume of verbose messages.

tombogle:
This makes UI more responsive and prevents a hang and internal error: "Transition into COM context ... for this RuntimeCallableWrapper failed..."

### DIFF
--- a/SIL.Windows.Forms/Progress/LogBox.cs
+++ b/SIL.Windows.Forms/Progress/LogBox.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Drawing;
+using System.Text;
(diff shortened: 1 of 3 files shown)


hatton:
@tombogle While we're in here... what is thread-safety status of this? The SafeInvoke moves things to the UI thread, so could a worker thread be writing to _stringBuilderPendingVerboseMessages only to have it cleared by the last line of FlushPendingVerboseMessageBuffer, and thus lose a message? Wouldn't be the end of the world :-)

tombogle:
I've added a comment to indicate that FlushPendingVerboseMessageBuffer must be invoked on UI thread (which it already was) to ensure thread safety.

PR #765, opened 2018-10-30
Part 1 of Update language info using new alltags.json

glasseyes:
This change is meant to include the latest Ethnologue data so that applications can use recent language codes and fix the SLDR code.

NRSI are doing the work of processing the Ethnologue data and iana and their tag data to produce alltags.json (which is alltags.json.txt in the project). Any data issues are their responsibility.

Both the Sldr class and the LanguageLookup class read this file to get what they want

Things to bear in mind that are changes from the current setup:
* The alltags data does not have Dialect names from the Ethnologue, which was used before.
* There is one region/country per tag not multiple
* The script is included in the code where it isn't implicit
* Macrolanguages are not marked separately, and some are included. If there is a problem with a particular macrolanguage then that is a data problem.
* Some of the data has private use tags longer than 8, IetfLanguageTag has relaxed that requirement to allow the data - so it is no longer conformant to RFC-5646
* StandardSubtags gets updated with language codes and scripts that are in alltags but not the iana registry.

This is incomplete:
* The alltags.json doesn't yet have a permanent internet location so there are still changes to make for updating to the latest file.
* We need requested feedback about the Ethiopian language names
* I still need to remove LanguageData and the Ethnologue resources as they aren't needed any more

These are extra to what is currently in the PR so initial review and testing of the PR would be helpful.

Please test this with your application or with SIL.Windows.Forms.TestApp -> Language

Search results are different from before. Please check if they make sense. Any problems are usually a data issue for NRSI.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/765)


SIL.Windows.Forms.WritingSystems.Tests/LanguageLookupModelTests.cs
```diff
@@ -24,7 +24,7 @@ public void IncludeRegionCodes_SetToFalseSearchForChinese_ReturnsTaiwanAndMainla
 			model.IncludeRegionalDialects = false;
 			model.SearchText = "chinese";
 			string[] codes = model.MatchingLanguages.Select(li => li.LanguageTag).ToArray();
-			Assert.That(codes, Contains.Item("zh-CN"));
```
jasonleenaylor: We will need to verify that existing projects which contain this previously valid code will continue to work for them.

SIL.WritingSystems/Resources/alltags.txt
```diff
@@ -1,2191 +0,0 @@
-*aa = aa-Latn | *aa-ET = aa-Latn-ET
```
tombogle: Daniel, I'm vaguely assuming you're on top of this since you're part of the Keyman team, but just in case, note that the following Keyman files refer to and/or access this file:
\resources\standards-data\alltags\readme.md
windows\src\buildtools\build_standards_data\Keyman.System.BuildLibPalasoAllTags.pas
windows\src\buildtools\build_standards_data\Makefile

glasseyes: I wasn't aware of that but Keyman gets it from the NRSI github rather than libpalaso so there isn't a problem. I'm making a minor PR to rename the class to make that explicit.

glasseyes:
merging so developers can start using it more, it is not ready for public deployment yet

tombogle:
See this thread: https://sil-language-software.slack.com/archives/C87FMEEE6/p1542217039003800

PR #773, opened 2018-11-20
Fix and improve global writing system store

ermshiperete:
- Allow to set path for GlobalWritingSystemRepo in unit tests

  Get the default path to the GlobalWritingSystemRepo from a field. This allows unit tests to set the field through reflection so that they can have the global writing system repo in a defined state. This is necessary when testing S/R where changes in the global WS store might cause additional commits to appear.

- Don't update timestamp when copying to global store

  Updating DateModified causes extra commits when doing a S/R because the WS gets copied from the global to the local store, introducing a modified timestamp and thus a change. This change adds an optional flag to `IWritingSystem.Create()` whether to clone the Id of the writing system.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/773)

### DIFF
--- a/GlobalAssemblyInfo.cs
+++ b/GlobalAssemblyInfo.cs
@@ -9,5 +9,5 @@ using System.Reflection;
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 
-[assembly: AssemblyFileVersion("5.0.0.0")]
-[assembly: AssemblyVersion("5.0.0.0")]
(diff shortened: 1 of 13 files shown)


ermshiperete:
This change modifies the API - does it require an update of the major version number?

ermshiperete:
> The version number still needs to be bumped for the LanguageData changes, so maybe this bump can be together with that?

Good idea.

PR #776, opened 2018-11-28
Don't complain if the Sldr is initialized repeatedly.

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/776)


jasonleenaylor:
I agree this does make things easier on the clients, but I'm still a bit concerned about the clean-up problem. Some kind of ref counting would be nice, but I won't block this if it is on some products release path.

JohnThomson:
It's been so long now I don't even recall which specific problem I wanted it for. Reference counting feels a bit like overkill, if a component that needs to use the SLDR creates a component that also uses it, most likely it won't discard that component until it's also done with the SLDR. And it would be easy to add if ever actually needed. But we could also just close this without merging; everyone seems to have found some workaround by now.

jasonleenaylor:
9 months ago John said he couldn't remember why we had done this. This can be reimplemented in the future if anyone discovers that they care.

PR #784, opened 2019-01-17
Fix crashes in using LanguageLookupDialog

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/784)

### DIFF
--- a/SIL.Windows.Forms.WritingSystems/LanguageLookupModel.cs
+++ b/SIL.Windows.Forms.WritingSystems/LanguageLookupModel.cs
@@ -72,6 +72,13 @@ namespace SIL.Windows.Forms.WritingSystems
 			}
 		}
 
+		/// <summary>
+		/// Filter out tags that contain a region marker unless the caller has already specified that region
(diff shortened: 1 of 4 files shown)


SIL.Windows.Forms.WritingSystems/LanguageLookupModel.cs
```diff
@@ -81,7 +81,14 @@ private bool RegionalDialectsFilter(LanguageInfo li)
...
 
-			return string.IsNullOrEmpty(IetfLanguageTag.GetRegionPart(li.LanguageTag));
+			// written this way to avoid having to catch predictable exceptions as the user is typing
+			string language;
+			string script;
+			string region;
+			string variant;
+			if (IetfLanguageTag.TryGetParts(li.LanguageTag, out language, out script, out region, out variant))
+				return string.IsNullOrEmpty(region);
+			return true;
```
jasonleenaylor: What does this return true mean? It seems like it should be return false to me.

JohnThomson:
I'd like Jason to take another look, too.

PR #548, opened 2017-06-26
Use system sound play instead of IrrKlang

JohnThomson:
Also adds a way to test interactively
Therefore makes TestApp x86 (since SIL.Media is)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/548)

PR #785, opened 2019-01-22
Fix: LT-19431 Audio won't play .mp3 files in FW9.0.4

JamesPrabu:
* Changed the code to support mp3 files
* Added unit test to test the method

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/785)


SIL.Media.Tests/AudioRecorderTests.cs
```diff
@@ -430,6 +431,22 @@ public void Play_DoesPlay ()
 			}
 		}
 
+		[Test]
+		public void Play_DoesPlayMp3()
+		{
+			using (var file = TempFile.FromResource(Resources.finished, ".mp3"))
```
jasonleenaylor: Since we are mocking everything no temp file is necessary. Just a path ending in .mp3 should work.

jasonleenaylor:
___
*[SIL.Media/AudioIrrKlangSession.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/785#-LWv8W718MQHLvcjAfBX:-LWvSv5J6K1R-GjIahAT:byojlu) ([raw file](https://github.com/sillsdev/libpalaso/blob/1c1f2d67abb577b3675ea2a453f6ba4921a9b3a5/SIL.Media/AudioIrrKlangSession.cs#L32)):*

It seems like this PR gives up on irrklang for mp3s and instead says "supply your own player". And yet it is still called an "AudioIrrKlangSession". That doesn't sound to good. Irrklang claims to be able to play mp3s, I suppose that didn't work? If we go ahead with this work-around, there should be heavy commenting explaining all this.
</blockquote></details>

Hmm, if IrrKlang is supposed have code to play mp3s then there must be a better solution here.

The SoundPlayer that is currently used in this class has nothing to do with irrklang and can only play .wav files.

jasonleenaylor:
___
*[SIL.Media/AudioIrrKlangSession.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/785#-LWv8W718MQHLvcjAfBX:-LWvUTSy1GGJtzcbBAJZ:bqaaxus) ([raw file](https://github.com/sillsdev/libpalaso/blob/1c1f2d67abb577b3675ea2a453f6ba4921a9b3a5/SIL.Media/AudioIrrKlangSession.cs#L32)):*

Hmm, if IrrKlang is supposed have code to play mp3s then there must be a better solution here.

The SoundPlayer that is currently used in this class has nothing to do with irrklang and can only play .wav files.
</blockquote></details>

Looks like we stopped using IrrKlang for playback a while back due to crashes.
 .mp3 playback was lost at that point.

jasonleenaylor:
___
*[SIL.Media/AudioIrrKlangSession.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/785#-LWv8W718MQHLvcjAfBX:-LWvWCoACwh9nc6eH2VJ:b-qo4va7) ([raw file](https://github.com/sillsdev/libpalaso/blob/1c1f2d67abb577b3675ea2a453f6ba4921a9b3a5/SIL.Media/AudioIrrKlangSession.cs#L32)):*

Looks like we stopped using IrrKlang for playback a while back due to crashes.
 .mp3 playback was lost at that point.
</blockquote></details>

https://github.com/sillsdev/libpalaso/pull/548

jasonleenaylor:
___
*[SIL.Media.Tests/SIL.Media.Tests.csproj, line 359 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/785#-LYgwXr0AS55x1UqRQcz:-LYgwXr0AS55x1UqRQd-:bbw8kpw) ([raw file](https://github.com/sillsdev/libpalaso/blob/c0083efb754152cc96a66a5df1383be7eda542d3/SIL.Media.Tests/SIL.Media.Tests.csproj#L359)):*
> ```XML
>     <Content Include="Resources\2Channel.wav" />
>     <None Include="Resources\short.mp3" />
>     <Content Include="Resources\sos-morse-code_daniel-simion.mp3" />
> ```

Remove, leftover from intermediate work.

PR #786, opened 2019-02-01
Create nuget packages

ermshiperete:
- All projects switched to the new sdk-style project files
- Added old-style `*-Designer.csproj` files for projects that implement dialogs or custom controls so that it’s possible to edit the forms/controls in Visual Studio’s designer
- Projects now require Visual Studio 2017, <= VS2015 won’t work. Using VS Code or Rider will also work. On Linux we require Mono 5.
- Check platform at runtime instead of compile-time - no more `#if MONO` statements in the code; same binary will work cross-platform.
- All projects are only compiled for `AnyCPU`
- Refactored SIL.Media:
  - Defined new `IRecordingDevice` interface which `IAudioRecorder.SelectedDevice` now returns. This interface is implemented in the `RecordingDevice` class of both NAudio and AlsaAudio. Since this is a breaking change the libpalaso version number went up to 7.0.
  - Load the correct 32- or 64-bit version of irrKlang.NET4 at runtime.
- (Almost) all assemblies are now strong-named. There’s only one or two where strong-naming is not possible because of mixed-platform dependencies.
- Added a [`CHANGELOG.md`](https://keepachangelog.com) file. Relevant changes and bugfixes should be documented in here. What gets added will show up as release notes on the nuget packages on nuget.org. What should go in this document is:
  - Any changes to the API
  - Bugfixes since the last released version
  - Other relevant changes that affect the consumers of the library
- Added [GitVersion](https://gitversion.readthedocs.io) to calculate the version number

See the [Google doc](https://docs.google.com/document/d/13KgAhU-UMVgbjRqVLbRpnBGMF45ph5JBuI8FzHNGRDA/edit?usp=sharing) for a more detailed discussion of the changes and the reasons why things were done a certain way.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/786)


tombogle:
The irrKlang stuff looks okay, but it doesn't really seem to belong in this PR

SIL.Core.Tests/packages.config
```diff
@@ -1,4 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="NUnit" version="2.6.4" targetFramework="net461" />
-</packages>
```
tombogle: Just curious: how/where does NUnit get referenced now? (Or why/how is it not needed?)

SIL.Media/FFmpegRunner.cs
```diff
@@ -4,511 +4,512 @@
...
-#endif
+			}
 
-#if !MONO
-            string withApplicationDirectory = GetPathToBundledFFmpeg();
+			string withApplicationDirectory = GetPathToBundledFFmpeg();
 
-            if (!string.IsNullOrEmpty(withApplicationDirectory) && File.Exists(withApplicationDirectory))
-                return withApplicationDirectory;
+			if (!string.IsNullOrEmpty(withApplicationDirectory) && File.Exists(withApplicationDirectory))
```
tombogle: Just below this line is a comment: "nb: this is sensitive to whether we are compiled against win32 or not, not just the host OS, as you might guess.
Is that (still) true? If so, what happens now that we are not doing conditional compilation?

ermshiperete:
> The irrKlang stuff looks okay, but it doesn't really seem to belong in this PR

Yes, it doesn't really fit in this PR, but it requires an interface change with an increment in version number, which I don't want to do separately.

Palaso.sln
```diff
@@ -153,130 +154,130 @@ Global
...
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugMonoStrongName|x86.Build.0 = DebugMonoStrongName|x86
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|Any CPU.ActiveCfg = DebugStrongName|Any CPU
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|Any CPU.Build.0 = DebugStrongName|Any CPU
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|Mixed Platforms.ActiveCfg = DebugStrongName|Any CPU
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|Mixed Platforms.Build.0 = DebugStrongName|Any CPU
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|x64.ActiveCfg = DebugStrongName|x64
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|x64.Build.0 = DebugStrongName|x64
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|x86.ActiveCfg = DebugStrongName|x86
-		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.DebugStrongName|x86.Build.0 = DebugStrongName|x86
+		{5DE33CD7-60CB-4B9F-A123-A83C1C686E47}.Debug|x64.ActiveCfg = Debug|Any CPU
```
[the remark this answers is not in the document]
ermshiperete: yes, eventually I'll delete those. Just haven't gotten that far yet. I should have applied the WIP label on this PR :smile:

SIL.Core.Tests/packages.config
ermshiperete: For SDK style csproj files they are referenced in the .csproj file.

SIL.Media/FFmpegRunner.cs
ermshiperete: Ahh, thanks. I didn't see that comment.

I don't understand this comment - it doesn't seem to match what the code is doing. Depending on whether we run as 32- or 64-bit process we get a different value for ProgramFiles. But then we look in both "Program Files" and "Program Files (x86)" for ffmpeg.exe and take whatever we find. I don't think this change changed anything how the method works. Correct me if I miss something.

tombogle:
> The irrKlang stuff looks okay, but it doesn't really seem to belong in this PR

Now that I've had a chance to digest this whole thing, I guess I kind of see why/how the irrKlang stuff had to be included as part of the whole removal of conditional compilation for Mono/Windows.

jasonleenaylor:
___
*[SIL.Media/FFmpegRunner.cs, line 58 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/786#-LYkjSE3FAn6vPDeR3GT-r2-58:-Latsu3q7crPD7lNygCA:bs911ht) ([raw file](https://github.com/sillsdev/libpalaso/blob/2daac931921a14354e37fa2e2b5c26a54d4affc6/SIL.Media/FFmpegRunner.cs#L58)):*

Ahh, thanks. I didn't see that comment.

I don't understand this comment - it doesn't seem to match what the code is doing. Depending on whether we run as 32- or 64-bit process we get a different value for ProgramFiles. But then we look in both "Program Files" and "Program Files (x86)" for ffmpeg.exe and take whatever we find. I don't think this change changed anything how the method works. Correct me if I miss something.
</blockquote></details>

The comment is explaining why the developer did some trickery below with the progFileDirs.
So it doesn't really matter. Also the code is dumb, or rather predates .net 4

var progFileDirs = new List<string>() {Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86),
Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles)}; ought to do the trick.

jasonleenaylor:
___
*[SIL.Archiving/Generic/AccessProtocol/ELARAccessProtocol.cs, line 1 at r14](https://reviewable.io/reviews/sillsdev/libpalaso/786#-LaVCJELBVMetgR9mmzn:-LatwF7M4U07fsoe8PdC:b-e7jgic) ([raw file](https://github.com/sillsdev/libpalaso/blob/98677d258e9ad37bf6f1fc8dc43af239b93b59f6/SIL.Archiving/Generic/AccessProtocol/ELARAccessProtocol.cs#L1)):*

No, but they are not included in the old .csproj files, so they don't get build at all. Nobody seems to miss them and so it seems save to remove them. 

It's strange, I can't find the commit that removed it from the .csproj file - it got added in 2013. But it's definitely not in any of the versions since libpalaso-2.6. And the link in the class comment doesn't work anymore. So I suggest that we remove the files, and if it turns out that we still need them, someone will complain and then we can re-add them.


</blockquote></details>

Ok, unless @tombogle or @gtryus chime in we can just remove them.

rmunn:
One other thing that needs to be adjusted is the default configuration in the `build/TestBuild.sh` script, which defaults (if nothing is specified on the command line) to "DebugMonoStrongName", a configuration that is no longer present in the `.csproj` files. Changing that to "Debug" allowed me to build the project without errors. I'm sure similar adjustments need to happen in the `TestBuild*.bat` files, but as my current dev machine is Linux-only, I can't test those batch files right now.

ermshiperete:
@rmunn It's true that we decided:

> Long-term goal is to have independent version numbers for the different assemblies.

But we also said: 

> For practical reasons (minimize cost) the nuget packages initially will still be in the same git repo and all will share the same version number. 

(see end of [Libpalaso Breakup document](https://docs.google.com/document/d/1XDBFFeu8xZ2oRUru2iYUENnnVbBwWYTRqvMlYNS7jsg/edit?usp=sharing)). IMO it makes sense to merge the changes to create nuget packages, and then make it right in a later change. Lock-step versions is no worse than what we currently have.

rmunn:
Update to my earlier: https://github.com/AArnott/Nerdbank.GitVersioning/issues/102 makes me not exactly a fan of how GitVersioning does things either (I strongly disagree with his opinion re: what the SemVer spec requires here). If we do use GitVersioning, we'd probably want to fork it to change that design decision. Again, I don't have time to discuss this in detail right now, unfortunately.

tombogle:
___
*[SIL.Archiving/Generic/AccessProtocol/ELARAccessProtocol.cs, line 1 at r14](https://reviewable.io/reviews/sillsdev/libpalaso/786#-LaVCJELBVMetgR9mmzn:-LjMKk7L0QlDRC4Fe_Vk:b-4opxvn) ([raw file](https://github.com/sillsdev/libpalaso/blob/98677d258e9ad37bf6f1fc8dc43af239b93b59f6/SIL.Archiving/Generic/AccessProtocol/ELARAccessProtocol.cs#L1)):*

Ok, unless @tombogle or @gtryus chime in we can just remove them.
</blockquote></details>

Pretty sure these were added as some kind of a false start. They would have been for SayMore and SayMore has no references to them, nor to AccessProtocolBase or IAccessProtocol (neither of which seem to exist anywhere). Should be safe to delete them.

PR #789, opened 2019-02-14
Improved the password notice in the settings protection dialog.

tombogle:
Apps can now override the default password if needed and also specify the support page URL.
This should be 100% compatible with existing apps and not break anything. However, apps that ship their own localized versions of Palaso TMX file will want to fix any existing translations for SettingsProtection.PasswordNotice. (I rewrote the wording such that googling the new param 2 should still work -- even better than before -- but it might sound/look weird if the wording is not adjusted.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/789)

### DIFF
--- a/DistFiles/Palaso.es.tmx
+++ b/DistFiles/Palaso.es.tmx
@@ -571,12 +571,21 @@
 	  </tuv>
 	</tu>
 	<tu tuid="SettingsProtection.PasswordNotice">
-	  <note>Don't forget to have the {0}, {1} and {2} in your translated string!</note>
+	  <note>The localization for this should be kept in sync with "SettingsProtection.PasswordNoticeWithSupportUrl". Param 0: Factory password; Param 1: product name</note>
 	  <tuv xml:lang="en">
-		<seg>Factory password for these settings is "{0}".  If you forget it, you can always google for "{1}" and "{2}"</seg>
+		<seg>Factory password for these settings is \"{0}\". If you forget it, you can always visit the {1} support page."</seg>
 	  </tuv>
 	  <tuv xml:lang="es">
-		<seg>Contraseña de fábrica para estos ajustes es "{0}".  Si lo olvida, siempre puedes usar Google para buscar "{1}" y "{2}"</seg>
+		<seg>Contraseña de fábrica para estos ajustes es "{0}". Si lo olvida, siempre puedes encontrarlo en la página de apoyo de {1}.</seg>
+	  </tuv>
+	</tu>
+	<tu tuid="SettingsProtection.PasswordNoticeWithSupportUrl">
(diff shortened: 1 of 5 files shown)


tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtection.cs, line 97 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/789#-LYhPdFe9XIjToe6dlzJ:-LYl8A0c8jz61T7b-Uit:bsubszp) ([raw file](https://github.com/sillsdev/libpalaso/blob/0e8480842088eef04bfb07fadb9ab58d58cb27f3/SIL.Windows.Forms/SettingProtection/SettingsProtection.cs#L97)):*

Is there a way to make this and the url a parameter set in code rather than a setting?
Seems like that would be much easier to deal with and more discoverable for developers.
</blockquote></details>

That would have been my preference, too. The way it is actually made it harder for me to find the relevant bits of code and figure out how the plumbing worked. I don't know why it was done this way. The only possible advantage I could see would be _maybe_ the possibility of the user overriding it in the config file. So if, for example, the local support situation was such that a different support URL was needed (for support in another language, because the government blocked ours, to facilitate the use of a local network to avoid Internet access, etc.), a computer support person could tweak it. Seems highly unlikely, and I'm not even sure it would work, but that might be what motivated this decision. I kept it this way because a) I wasn't sure about the original motivation and b) to prevent breaking anything (or changing the API, forcing us to move to v. 7).

tombogle:
Although I agree in principle with the perceived benefit of telling the user they can just google the two keywords, in practice it doesn't currently work for either of the two products using it. The HearThis search only give 4 results, none of which is useful (though I'm vaguely hopeful that eventually Google's bot will find the one on our community site). The Bloom search gives many pages of results, but only turns up one promising hit on page 1. That link takes them to github, which is probably not what we're hoping for. While it's true that the user might feel intimidated by the URL, in actual fact, both products make it quite easy for the user to find the support page without actually having to remember the URL.
Another option would be to actually have a "Forgot password?" link on the password page to take them to the relevant support page. That might be enough of a barrier to a novice user, while still giving the administrator a way out. We could hide the link until they attempt a bogus password. Or instead of an actual link, we could remind them that they can find the password on the support page. That would raise the bar slightly.

tombogle:
Even if/when they show up on Google, it will be quite difficult for us to predict/guarantee that they will show up in the number 1 slot (or anything close to that). It's just too fragile of a strategy to be reliable.
The two new params could be set directly, but I figured that it probably made sense to keep them together. And again, I was trying to avoid an unnecessary API change (especially right now with Eberhard working on the massive NuGet change). Technically, I guess in one sense this is a breaking change anyway, since it will require an actual change to HearThis to get the same behavior. So maybe my reasoning is faulty.

tombogle:
When I went to make the requested code changes, I realized the password override was a YAGNI. I had mis-read the code and thought HT would need it. So I just removed that capability altogether.

PR #792, opened 2019-02-22
Fix private use scripts showing up as normal

jasonleenaylor:
* Scripts added by SLDR from the languageTags.json
  were not retaining their private use identity
* Also cleaned up some comments and added some new
  comments relating to the private use convention

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/792)

### DIFF
--- a/SIL.WritingSystems/StandardSubtags.cs
+++ b/SIL.WritingSystems/StandardSubtags.cs
@@ -38,7 +38,8 @@ namespace SIL.WritingSystems
 			var variants = new List<VariantSubtag>();
 			foreach (string ianaSubtagAsString in ianaSubtagsAsStrings)
 			{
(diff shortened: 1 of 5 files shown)


SIL.WritingSystems/StandardSubtags.cs
```diff
@@ -179,7 +181,7 @@ private static IEnumerable<string> GetVariantPrefixes(string[] subTagComponents)
 
 		public static void AddScript(string script, string description)
 		{
-			var scriptTag = new ScriptSubtag(script, description, false, false);
+			var scriptTag = new ScriptSubtag(script, description, IsPrivateUseScriptCode(script), false);
 			RegisteredScripts.Add(scriptTag);
```
StephenMcConnel: This looks like the one critical line of this change as far as I can tell.

PR #795, opened 2019-02-26
Do not show Script/Region/Variant for Implicit Script

jasonleenaylor:
* If a language tag is using an implicit script it should
  not push the user into the Script/Region/Variant options.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/795)


JohnThomson:
OK, LGTM. I'll let you decide whether to merge or wait and see what others say...also whether any of the failing checks matter.

PR #812, opened 2019-03-19
Update to latest langtags.json

jasonleenaylor:
* Rename variables
* Handle json parse problems in SLDR cache
* Parse new region format

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/812)


SIL.WritingSystems/LanguageLookup.cs
```diff
@@ -30,13 +30,15 @@ public class LanguageLookup
...
+			string langTagsContent = LanguageRegistryResources.langTags;
 
-			List<AllTagEntry> rootObject = JsonConvert.DeserializeObject<List<AllTagEntry>>(allTagsContent);
+			//REVIEW: Shouldn't this use the SLDR cache version?
+			List<AllTagEntry> rootObject = JsonConvert.DeserializeObject<List<AllTagEntry>>(langTagsContent);
 
 			foreach (AllTagEntry entry in rootObject)
 			{
-				if (!entry.deprecated && !entry.tag.StartsWith("x-")) // tags starting with x- have undefined structure so ignoring them as well as deprecated tags
+				// tags starting with x- have undefined structure so ignoring them as well as deprecated tags				if (!entry.deprecated && !entry.tag.StartsWith("x-") && !entry.tag.StartsWith("_"))
```
jasonleenaylor: oops, the if statement shouldn't be commented out.

glasseyes:
Some of the Sldr tests were stil referring to alltags so that caused 2 of the 4 failures. I've pushed a change for that.
The AllTagsTime in the resource stores the date of the embedded json file which is used to request a newer one if available from the ldml server. That should be updated when a new langtags is included. Feel free to update or modify the code to something simpler to maintain.

The other 2 test failures are to do with Ethopian language names and deprecated tags so must pass before this can be merged.

PR #813, opened 2019-03-21
Save DefaultCollationType from PM to WritingSystem

jasonleenaylor:
* Fix bug where the DefaultCollationType wasn't being
  saved into the actual writing system when it was set in
  the WritingSystemSetupModel
* Add unit test

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/813)


jasonleenaylor:
Hmm, using this seems to cause some other odd behavior, let's not merge it yet.

jasonleenaylor:
Turns out that this is a bad idea - It looks reasonable and was testable, but the two types here have different purposes and expected values.

PR #814, opened 2019-03-28
Improve NumberingSystemDefinition API

jasonleenaylor:
* Add the NumberingSystemDefinition to the PM
* Add digit access (for use by view code)

Note: So far Flex is the only user of this model

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/814)

### DIFF
--- a/SIL.Windows.Forms.WritingSystems/WritingSystemSetupModel.cs
+++ b/SIL.Windows.Forms.WritingSystems/WritingSystemSetupModel.cs
@@ -870,6 +870,19 @@ namespace SIL.Windows.Forms.WritingSystems
 			}
 		}
 
+		public NumberingSystemDefinition CurrentNumberingSystemDefinition
+		{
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
___
*[SIL.WritingSystems/NumberingSystemDefinition.cs, line 47 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/814#-Lb7JwkB2XXmigob5ZTN:-Lb91ulG4zadtPI7f2Yj:b-yl1o9v) ([raw file](https://github.com/sillsdev/libpalaso/blob/e9e3bdda52a0d034199faabda5dc27b3840bc68f/SIL.WritingSystems/NumberingSystemDefinition.cs#L47)):*

Are we sure it's always decimal? Or how does this work for non-decimal systems? Or is this only used for base-ten numbering systems?
</blockquote></details>

All CLDR data assumes base 10 numbering systems so that is all we will support for now.

PR #816, opened 2019-04-04
If SLDR is offline (like today) ignore tests which need it

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/816)


FoolRunning:
I reviewed the changes and they look fine. However, I don't have the knowledge to know if it's a good idea or not to always skip those tests that you changed to not run on TC.

PR #817, opened 2019-04-11
Change WritingSystemRepository interaction with the Id field

jasonleenaylor:
* Stop treating the Canonicalized LanguageTag as a stable identifier
* Allow for multiple writing systems in the repository to have the same
  Canonicalized LanguageTag data
* Update the tag cleaner to handle some of the clean-up that
  Canonicalize did.
* Don't rename .ldml files if a writing system id has changed
  but is still equivalent
* Don't canonicalize language tags on WritingSystemDefinition creation

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/817)

### DIFF
--- a/SIL.WritingSystems.Tests/GlobalWritingSystemRepositoryTests.cs
+++ b/SIL.WritingSystems.Tests/GlobalWritingSystemRepositoryTests.cs
@@ -322,7 +322,7 @@ namespace SIL.WritingSystems.Tests
 		}
 
 		[Test]
-		public void AllWritingSystems_LdmlCheckingSetEmptyandGetWSId()
+		public void AllWritingSystems_LdmlCheckingSetEmptyCanNotSave()
 		{
 			using (var tf = new TemporaryFolder("GlobalWritingSystemRepositoryTests"))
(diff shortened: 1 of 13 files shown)


jasonleenaylor:
___
*[SIL.WritingSystems/WritingSystemDefinition.cs, line 123 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/817#-LckJmtHEGOi4uQt66kg:-LckJmtICOMjvSWz5od6:b-655dtd) ([raw file](https://github.com/sillsdev/libpalaso/blob/bec74b5f6ce1cc3de5276d503d1d6725913449b6/SIL.WritingSystems/WritingSystemDefinition.cs#L123)):*
> ```Smalltalk
> 				throw new ArgumentException("The language tag is invalid.", languageTag);
> 			_numberingSystem = NumberingSystemDefinition.Default;
> 			// If we canonicalize the tag when we read it in it will cause ldml files to move unexpectedly when we save.
> ```

I need to verify the necessity of this change, it was made early in the process and may now be unnecessary.

PR #820, opened 2019-05-02
Improve LanguageLookup to work better with new langtags data

jasonleenaylor:
* Add Country match into sort
* Add Match on any Name into sort
* Include iana names into data considered by the lookup
* Add class to handle loading a single json string into
  a string collection (to handle the iana data)
* Add debugging help to LanguageInfo
* Modernize Exception handling in IetfLanguageTagTests

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/820)

### DIFF
--- a/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs
+++ b/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs
@@ -206,24 +206,24 @@ namespace SIL.WritingSystems.Tests
 
 		#region Create
 		[Test]
-		[ExpectedException(typeof(ArgumentException), ExpectedMessage = "code [a] is invalid", MatchType = MessageMatch.Contains)]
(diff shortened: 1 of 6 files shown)


jasonleenaylor:
___
*[SIL.WritingSystems/LanguageLookup.cs, line 143 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/820#-LdtS37s7BnovBTB4TvP:-LdttH9E0X3UECpTJLZk:b-ppwwls) ([raw file](https://github.com/sillsdev/libpalaso/blob/d53d04839fedda8729384a046864aa68c75be1fc/SIL.WritingSystems/LanguageLookup.cs#L143)):*

The old code trimmed the name before adding it. Is that not needed?
</blockquote></details>

apparently not, but who knows what kind of junk data we will get back in the langtags in the future. I'll add the trim back.

PR #827, opened 2019-06-11
Report useful error from bad font content in ldml

jasonleenaylor:
At least one user was found to have a duplicated font in
their ldml data (we don't know how it was duplicated)

* Throw an exception with the actual font name and .ldml file
  when a duplicate is found during the parse.
* Add a unit test to verify this.
* Fix a typo that I noticed in a random file

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/827)

### DIFF
--- a/SIL.WritingSystems.Tests/LdmlDataMapperTests.cs
+++ b/SIL.WritingSystems.Tests/LdmlDataMapperTests.cs
@@ -1277,6 +1277,25 @@ namespace SIL.WritingSystems.Tests
 			}
 		}
 
+		[Test]
+		public void DuplicatedFontInLdml_ReportsBadData()
+		{
+			using (var roundTripOut = new TempFile())
+			using (var tempFile = new TempFile())
+			{
+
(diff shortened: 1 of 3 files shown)


jasonleenaylor:
___
*[SIL.WritingSystems/LdmlDataMapper.cs, line 475 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/827#-Lh6uOIpEjzO7fsfztZb:-Lh6zaVc1taoAvatVkl3:bxdid2k) ([raw file](https://github.com/sillsdev/libpalaso/blob/b16a4719d3a859daabd2e56a7cc340045bc617dc/SIL.WritingSystems/LdmlDataMapper.cs#L475)):*

So the idea is this would have helped the support person know what to look for? 
Doesn't seem like this would be helpful to the user.
</blockquote></details>

Yes. The program can't know which of the font definitions for the font name is correct (and they could be different). The ldml file is invalid, and if we can find the bug that made it we need to fix that. Just dropping the second one is an option, but not a great one in my opinion.

PR #829, opened 2019-06-26
LT-19632: Share WS Data w/ SLDR by default

papeh:
If the user has not set a preference for the project,
and has not set the FEEDBACK=off environment variable,
then enable sharing of Writing System data with SLDR
by default.
https://jira.sil.org/browse/LT-19632

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/829)


papeh:
making this change in LibLCM instead

PR #832, opened 2019-07-03
Avoid getting the list of recording devices twice when updating.

tombogle:
This is mainly for (massive) performance improvement, but it also avoids a race condition.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/832)

### DIFF
--- a/SIL.Media/Naudio/UI/RecordingDeviceIndicator.cs
+++ b/SIL.Media/Naudio/UI/RecordingDeviceIndicator.cs
@@ -1,4 +1,4 @@
-﻿#if !MONO
+#if !MONO
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


tombogle:
JohnT was wondering why this produced such a massive improvement. It turns out that getting the list of devices is fairly slow. Apparently, enough so, that getting the list twice every second, in addition to everything else we're doing causes the code to re-execute immediately or nearly immediately after it completes, which sucks up 100% of the CPU. By getting the list once and passing it in to SetKnownRecordingDevices, we leave some CPU cycles for other important stuff. We weren't expecting it drop off as much as it did, and we're not sure why it helped that much, but it was really significant. Note that callers should still make sure that the checking for a mic change is suspended during other performance-critical operations, especially when recording or doing other things that make a mic change an invalid or undesirable thing.

PR #834, opened 2019-07-09
Refactor Progress

ermshiperete:
Follow-up of the big nuget PR (#786)

- move Progress related classes in separate files
- refactor ConsoleProgressTests to test what the name says
- add comments to ProgressState classes that will eventually be deprecated

This change doesn't deprecate ProgressState yet because it is still used in too many libpalaso classes, and deprecating it would add a whole bunch of new warnings when compiling libpalaso. Instead I'd recommend to replace the usage of ProgressState over time.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/834)


ermshiperete:
Not ready yet. Depends on #836.

ermshiperete:
This change adds a whole bunch of new warnings - everywhere where we use `ProgressState` in existing libpalaso code.

PR #838, opened 2019-07-15
 Fix global WS repo to compare ws ids case insensitive (LF-297) 

ermshiperete:
While writing system ids in general were already case insensitive, the global writing system repository used two fields that compared by case. This is a problem on Linux where the file system is case
sensitive. This change modifies `GlobalWritingSystemRepository` to use case-insensitive dictionaries to store the ws ids. This fixes LF-297.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/838)

### DIFF
--- a/SIL.WritingSystems.Tests/GlobalWritingSystemRepositoryTests.cs
+++ b/SIL.WritingSystems.Tests/GlobalWritingSystemRepositoryTests.cs
@@ -11,12 +11,17 @@ namespace SIL.WritingSystems.Tests
 	[TestFixture]
 	public class GlobalWritingSystemRepositoryTests
 	{
+		private static TemporaryFolder CreateTemporaryFolder(string testName)
+		{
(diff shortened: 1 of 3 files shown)


ermshiperete:
> If through S/R or some other means the folder ends up with en-us and en-US what is the result after we save? My guess is that both files will remain on disc but only the first one loaded will be changed. 

Yes, both remain on disc and the second one will be ignored.

> If this S/R then went back to Windows could it be that the second file contents is the one retained?

That's conceivable, although the two writing systems would be in different projects.

> Another thing to think about is the dreaded "case-folding" exception on Mercurial on Windows.

I can't imagine how this could be possible with the current (fixed) code. The only way we can get two files that differ only by case is that two projects on two different machines have a ws id that differs by case. By fetching the other project from LD they can get the case-differing file which might end up in the global store. But I don't think with the fixed code it's possible that these two files end up in Mercurial.

So I'd suggest that we merge this fix and deal with the other problems if a user experiences them.

PR #841, opened 2019-07-23
BL-7224 Give language lookup click priority

gmartin7:
* 2 UI desires are conflicting here:
   1) update the Script and Variant label (to the right of the Script and Variant link)
       with the language clicked on in the list view
   2) keep script/region/variant info from the Script and Variant dialog when
       returning to the main list view.
* At this point, we are making the first a priority. If the user selects their
   language, then goes and adds more detail in the Script and Variant dialog,
   they can click 'OK' in the outer dialog and keep all that info. But clicking on a
   different language will reset that information.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/841)

### DIFF
--- a/SIL.Windows.Forms.WritingSystems/LanguageLookupControl.cs
+++ b/SIL.Windows.Forms.WritingSystems/LanguageLookupControl.cs
@@ -146,9 +146,10 @@ namespace SIL.Windows.Forms.WritingSystems
 				ListViewItem item = _listView.Items[_listView.SelectedIndices[0]];
 				var oldLangInfo = SelectedLanguage;
 				var newLangInfo = (LanguageInfo) item.Tag;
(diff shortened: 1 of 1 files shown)


andrew-polk:
Jason does have some feedback. He's looking.

jasonleenaylor:
Ok, so this is the bug encountered in FieldWorks which removing this code will probably re-introduce.

The existing language is say en-Arab-GB and a user clicks on our 'Change Language' button that brings up the LanguageLookupControl. It is pre-populated with the existing language name. The row for 'English' is selected as the top search result. This wipes out the Script and Region.

The user never goes to the Script/Variant dialog (we don't have that option turned on for this use case) and clicks 'Ok' and without ever making a choice their Script and Region are lost.

PR #845, opened 2019-07-26
Made LogBox enforce the MaximumLength value

tombogle:
MaximumLength is now exposed so the client can set it (mainly to support testing).
The logic is in place (with passing unit tests) to truncate any message that would exceed the limit and output a special error message.
WIP: Failing (and very slow) test for the more interesting case where the default maximum is used. In this case, an access violation occurs.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/845)

### DIFF
--- a/SIL.Windows.Forms.Tests/Progress/LogBox/LogBoxTests.cs
+++ b/SIL.Windows.Forms.Tests/Progress/LogBox/LogBoxTests.cs
@@ -48,6 +48,7 @@ namespace SIL.Windows.Forms.Tests.Progress.LogBox
 				sb.Append('a', Int32.MaxValue - (hugestr.Length + 1) * 999 - 1);
 				progress.WriteVerbose(sb.ToString());
 				progress.WriteVerbose(".");
+				Assert.IsTrue(progress.ErrorEncountered);
 				Assert.IsFalse(progress.Rtf.Contains("."));
 				Assert.IsTrue(progress.Rtf.Contains("Maximum length exceeded!"));
 			}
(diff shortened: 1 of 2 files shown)


tombogle:
I have removed the WIP label now, but I'm still not 100% sure this is the best approach. I think it is an improvement, but I'd feel much better with a fail-safe solution.

tombogle:
FWIW, I believe this control is used only by HearThis, Phonology Assistant, and Glyssen. I have corrected the issue in my Glyssen utility that was causing it to dump an excessive number of messages, so this is not a top priority for me now, but I believe my work (and the new unit tests) are an improvement, so it would be good to get the changes reviewed. If anyone can suggest a way to make it even better, that would be awesome.

tombogle:
I wanted it to write the error to the non-verbose box when the verbose one fills up because it feels like it's probably a good thing for the user to know, even if they are only looking at the terse view. Unless I'm mistaken (and I think my unit tests show I'm not), the error message should only be output when the overflow first happens. So in the terse box it will continue to write non-verbose messages after notifying the user of the problem. IIRC, it will also display a second error if/when the terse box fills up.

tombogle:
I'm mistaken. I'll try to fix it.

tombogle:
___
*[SIL.Windows.Forms/Progress/LogBox.cs, line 354 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/845#-LlDwad2813FxIBWdMvx:-LlYVL79-119q6GNGhLO:b-7637sb) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5e5cf5e7cac0a4c7c7019f94ca2c9657149c123/SIL.Windows.Forms/Progress/LogBox.cs#L354)):*

Maybe, if enough characters to hold message? Code above makes sure it will fit once, but when you try to add it the second time, you might have half a message's worth? Overflowing by that much probably doesn't matter, but in that situation the user doesn't need to see it twice.
</blockquote></details>

I think my latest check-in solves this now.

tombogle:
___
*[SIL.Windows.Forms/Progress/LogBox.cs, line 309 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/845#-LlDtY_14PRdAXmoyAjs:-LlYW-00EUs1KxYOlEFO:bg2f2t3) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5e5cf5e7cac0a4c7c7019f94ca2c9657149c123/SIL.Windows.Forms/Progress/LogBox.cs#L309)):*

It's safe to use rtfBox here, inside the delegate? (Of course this isn't new code...)
</blockquote></details>

Apparently, but maybe there's some situation where it wouldn't be. I can't think of any good reason not to use rtfBoxForDelegate, and it might avoid a future problem, so I'll change it.

PR #858, opened 2019-09-04
PTX-18778 Prevent errors when initializing keyboards

jwickberg:
Prevent bad input language settings from creating errors when initializing keyboards

This may be related to InKey keyboards - some systems have problems when InKey is running and Paratext is started.

The actual error is occurring in .Net code, so just needed to try accessing the value causing the error and ignore input languages that fail.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/858)

### DIFF
--- a/SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs
+++ b/SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs
@@ -81,7 +81,7 @@ namespace SIL.Windows.Forms.Keyboarding.Windows
 
 			foreach (InputLanguage inputLanguage in InputLanguage.InstalledInputLanguages)
 			{
-				if (InvalidLanguageName(inputLanguage))
+				if (HasInvalidLanguageName(inputLanguage))
 					continue;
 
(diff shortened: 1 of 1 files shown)


SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs
```diff
@@ -144,6 +147,20 @@ public void UpdateAvailableKeyboards()
...
 
+		private bool InvalidLanguageName(InputLanguage inputLanguage)
+		{
+			// We have had some input languages get errors when trying to get the layout name, so just want to skip those
+			try
+			{
+				string test = inputLanguage.LayoutName;
+				return false;
+			}
+			catch (Exception)
```
ermshiperete: Would it be better to catch a specific exception?

I wonder if the exception we get here is the one we caught previously which got refactored/removed in bfedd4ae811261b585e13bc7823a62781a8c11cb.

jwickberg: There is no documented exception for InputLanguage.LayoutName, so that's why I just caught anything.

We are still accessing the property that caused the Paratext problem report, so even with the other changes that I did recently, this error is still possible.

PR #863, opened 2019-09-18
SP-1816: Crash if archiving target directory is not accessible

phillip-hopper:
Let the user know the directory is not accessible, without a fatal exception.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/863)

### DIFF
--- a/SIL.Archiving/ArchivingDlgViewModel.cs
+++ b/SIL.Archiving/ArchivingDlgViewModel.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -546,6 +546,29 @@ namespace SIL.Archiving
 
 		/// <remarks/>
 		public Dictionary<string, MessageType> AdditionalMessages { get; private set; }
+
+		public bool PathIsAccessible(string directory)
+		{
+			try
+			{
+				var file = Path.Combine(directory, "Export.imdi");
+
+				if (File.Exists(file))
+					File.Delete(file);
+
+				File.WriteAllText(file, @"Export.imdi");
+
+				if (File.Exists(file))
+					File.Delete(file);
+			}
+			catch (Exception e)
+			{
+				DisplayMessage(e.Message, MessageType.Warning);
+				return false;
+			}
+
+			return true;
+		}
 	}
 
 	public interface ISupportMetadataOnly
--- a/SIL.Archiving/ArchivingDlgViewModel.cs
+++ b/SIL.Archiving/ArchivingDlgViewModel.cs
@@ -548,28 +548,26 @@ namespace SIL.Archiving
 		/// <remarks/>
 		public Dictionary<string, MessageType> AdditionalMessages { get; private set; }
 
-		public bool PathIsAccessible(string directory)
+		public bool IsPathWritable(string directory)
 		{
 			try
 			{
-				var isWritable = DirectoryUtilities.IsDirectoryWritable(directory);
-
-				if (isWritable)
+				if (DirectoryUtilities.IsDirectoryWritable(directory))
 					return true;
-
-				var msg = LocalizationManager.GetString(
-					"DialogBoxes.ArchivingDlg.DirectoryNotWritableMsg",
-					"The path is not accessible: {0}");
-
-				DisplayMessage(string.Format(msg, directory), MessageType.Normal);
-
-				return false;
 			}
 			catch (Exception e)
 			{
 				DisplayMessage(e.Message, MessageType.Warning);
 				return false;
 			}
+
+			var msg = LocalizationManager.GetString(
+				"DialogBoxes.ArchivingDlg.PathNotWritableMsg",
+				"The path is not writable: {0}");
+
+			DisplayMessage(string.Format(msg, directory), MessageType.Warning);
+
+			return false;
 		}
 	}
 
--- a/SIL.Archiving/ArchivingDlgViewModel.cs
+++ b/SIL.Archiving/ArchivingDlgViewModel.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Windows.Forms;
 using L10NSharp;
 using SIL.Archiving.Generic;
+using SIL.IO;
 
 namespace SIL.Archiving
 {
@@ -551,23 +552,24 @@ namespace SIL.Archiving
 		{
 			try
 			{
-				var file = Path.Combine(directory, "Export.imdi");
+				var isWritable = DirectoryUtilities.IsDirectoryWritable(directory);
 
-				if (File.Exists(file))
-					File.Delete(file);
+				if (isWritable)
+					return true;
 
-				File.WriteAllText(file, @"Export.imdi");
+				var msg = LocalizationManager.GetString(
+					"DialogBoxes.ArchivingDlg.DirectoryNotWritableMsg",
+					"The path is not accessible: {0}");
 
-				if (File.Exists(file))
-					File.Delete(file);
+				DisplayMessage(string.Format(msg, directory), MessageType.Normal);
+
+				return false;
 			}
 			catch (Exception e)
 			{
 				DisplayMessage(e.Message, MessageType.Warning);
 				return false;
 			}
-
-			return true;
 		}
 	}
 
--- a/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
+++ b/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
@@ -1,4 +1,3 @@
-﻿using System;
 using System.Collections.Generic;
 using System.Xml;
 using NUnit.Framework;
@@ -58,6 +57,23 @@ namespace SIL.Archiving.Tests
 
 			Assert.AreEqual(21, dirName.Length);
 			Assert.AreEqual("T_st_Title_", dirName.Substring(0, 11));
+		}
+
+		/// ------------------------------------------------------------------------------------
+		[Test]
+		public void PathIsAccessible_WritablePath_True()
+		{
+			var dir = _tmpFolder.Path;
+			var writable = _model.PathIsAccessible(dir);
+			Assert.True(writable);
+		}
+
+		[Test]
+		public void PathIsAccessible_NonexistentPath_False()
+		{
+			const string dir = "/one/two";
+			var writable = _model.PathIsAccessible(dir);
+			Assert.False(writable);
 		}
 
 		#endregion
--- a/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
+++ b/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
@@ -11,8 +11,21 @@ namespace SIL.Archiving.Tests
 	[Category("Archiving")]
 	internal class IMDIArchivingDlgViewModelTests
 	{
+		private class MessageData
+		{
+			public string MsgText;
+			public ArchivingDlgViewModel.MessageType MsgType;
+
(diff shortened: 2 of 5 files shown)


SIL.Archiving/ArchivingDlgViewModel.cs
```diff
@@ -546,6 +546,29 @@ public static bool IsMono
 
 		/// <remarks/>
 		public Dictionary<string, MessageType> AdditionalMessages { get; private set; }
+
```
tombogle: Let's put this in the DirectoryUtilities class and call it IsDirectoryWritable. Rather than creating/overwriting a file that could be present, generate a sub-folder name (ensure uniqueness by calling GetUniqueFolderPath) and just attempt to create (and then clean up) the subfolder. If we can create a subfolder, we should be assured of full write access.

phillip-hopper: `GetUniqueFolderPath(folderPath)` returns a directory name that is in the parent of `folderPath`, so I'm instead I'm using
```csharp
var tempPath = Path.Combine(folderPath, Path.GetRandomFileName());
```

phillip-hopper: Done.

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 564 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp7AmzwC916XY_ZE_F4:-Lp8ZqLS5gOnXTZrG_sY:b-3glmgh) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L564)):*

Displaying a message in the model seems a bit unfortunate to me. Wouldn't it be better to just return false and let the caller get the string and display the message?
</blockquote></details>

Agreed.

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 564 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp7AmzwC916XY_ZE_F4:-Lp8aG0rAaFuo3ygXoGJ:b8utod) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L564)):*

Agreed. 
</blockquote></details>

I just looked at this further, and despite its name, DisplayMessage actually just raises an event so the caller (in this case a dialog box) is responsible for setting up an appropriate event handler.

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 551 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8ayOm55JiQyPqTS1o:-Lp8ayOm55JiQyPqTS1p:bvqyh86) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L551)):*
> ```Smalltalk
> 		public Dictionary<string, MessageType> AdditionalMessages { get; private set; }
> 
> 		public bool PathIsAccessible(string directory)
> ```

Although we're not entirely consistent, the more common convention for properties and methods that return a bool is to put the word "Is" first in the name: IsPathAccessible

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 564 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp7AmzwC916XY_ZE_F4:-Lp8c-h527VM0f-WqnSm:b-k0uca8) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L564)):*

I just looked at this further, and despite its name, DisplayMessage actually just raises an event so the caller (in this case a dialog box) is responsible for setting up an appropriate event handler.
</blockquote></details>

Why is this message "Norma" but the one below a "Warning"?

tombogle:
___
*[SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs, line 75 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8cUcW5JEe0zsfWyFg:-Lp8cUcW5JEe0zsfWyFh:b-jqlyuu) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs#L75)):*
> ```Smalltalk
> 		{
> 			const string dir = "/one/two";
> 			var writable = _model.PathIsAccessible(dir);
> ```

These two tests (especially this one) should set up an event handler for OnDisplayMessage and also verify that the event is raised with the expected message (and not raised in the previous test).

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 568 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8cwfu032ohW-i78zP:-Lp8cwfu032ohW-i78zQ:bvhn9az) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L568)):*
> ```Smalltalk
> 				return false;
> 			}
> 			catch (Exception e)
> ```

Can we constrain the types of Exceptions we're catching to only include the ones we actually expect, so we don't accidentally catch a different problem? For example, if LocalizationManager.Get throws an exception, we don't want that to be caught here.

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 568 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8cwfu032ohW-i78zP:-Lp8dTmc8r48UoDrCwqG:b-bvwuye) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L568)):*

Can we constrain the types of Exceptions we're catching to only include the ones we actually expect, so we don't accidentally catch a different problem? For example, if LocalizationManager.Get throws an exception, we don't want that to be caught here.
</blockquote></details>

Can we constrain the types of Exceptions we're catching to only include the ones we actually expect, so we don't accidentally catch a different problem? For example, if LocalizationManager.GetString throws an exception, we don't want that to be caught here.

tombogle:
___
*[SIL.Core.Desktop/IO/DirectoryUtilities.cs, line 260 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8eO-4Az6FasErHO8U:-Lp8eO-4Az6FasErHO8V:b7fw71d) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Core.Desktop/IO/DirectoryUtilities.cs#L260)):*
> ```Smalltalk
> 
> 			// get a random name for a temporary directory in folderPath
> 			var tempPath = Path.Combine(folderPath, Path.GetRandomFileName());
> ```

The only problem with that (and it's extremely unlikely) is that there could already be a folder with that name in the directory and then we would be attempting to blow it away and re-create it. To be 100% safe, if it exists, we need to ask for a different random filename instead of blindly deleting it.

tombogle:
___
*[SIL.Core.Desktop/IO/DirectoryUtilities.cs, line 278 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8fidy14NUSFUnoaNn:-Lp8fidz4vEg52EZVYJV:b-881m6) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Core.Desktop/IO/DirectoryUtilities.cs#L278)):*
> ```Smalltalk
> 				try
> 				{
> 					if (Directory.Exists(tempPath))
> ```

I wonder if it would be better to say (inside the above try block):
if (!Directory.Exists(folderPath))
     return false;
Directory.Delete(folderPath);
return true;
I'm trying to think of a scenario where we would be unable to delete the temp directory we created but still be able to claim that the folder is writable. If something goes wrong in the attempt to clean up this temp folder, there's almost certainly something seriously wrong (probably some kind of hardware failure or a programming error) that we can't afford to ignore.

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 551 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8ayOm55JiQyPqTS1o:-Lp8hFhK5pgAzOJkVbRY:b-fo1xlo) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L551)):*

Although we're not entirely consistent, the more common convention for properties and methods that return a bool is to put the word "Is" first in the name: IsPathAccessible
</blockquote></details>

Also I think IsPathWritable would be more precise.

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 549 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp40Brj_sA4Nf2eJQqo-r1-549:-LphzQph0cl2IuqSGFU1:bb6ywmx) ([raw file](https://github.com/sillsdev/libpalaso/blob/be026af46b31c71eb6e0187da30b609e647c7908/SIL.Archiving/ArchivingDlgViewModel.cs#L549)):*
<details><summary><i>Previously, phillip-hopper (Phil Hopper) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

LGTM

tombogle:
___
*[SIL.Archiving/ArchivingDlgViewModel.cs, line 551 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8ayOm55JiQyPqTS1o:-Lphzexc8ZdBml4-F448:bb6ywmx) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Archiving/ArchivingDlgViewModel.cs#L551)):*
<details><summary><i>Previously, phillip-hopper (Phil Hopper) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

LGTM

tombogle:
___
*[SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs, line 93 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lpi-0pE336qTmxH10GS:-Lpi-0pE336qTmxH10GT:b-hmorgy) ([raw file](https://github.com/sillsdev/libpalaso/blob/b0b0440bc74cac724dad763b1512a45f9a2f9060/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs#L93)):*
> ```Smalltalk
> 			var writable = _model.IsPathWritable(dir);
> 			Assert.False(writable);
> 			Assert.IsNotEmpty(_messages);
> ```

Would be better to assert that there is exactly one message.

tombogle:
___
*[SIL.Core.Desktop/IO/DirectoryUtilities.cs, line 260 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/863#-Lp8eO-4Az6FasErHO8U:-Lpi-B6S5M6sOSDrMJfj:bb6ywmx) ([raw file](https://github.com/sillsdev/libpalaso/blob/e108db801006b742d70910f08f05186374c5b9a6/SIL.Core.Desktop/IO/DirectoryUtilities.cs#L260)):*
<details><summary><i>Previously, phillip-hopper (Phil Hopper) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

LGTM

PR #874, opened 2019-11-18
Moved several classes back to SIL.Core

andrew-polk:
So they can be used by .NetStandard application clients

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/874)


jasonleenaylor:
___
*[SIL.Core/Reporting/UsageReporter.cs, line 313 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/874#-Lu-LVxb8OWjJrGYqhww:-Lu-LVxcEvYXZl2DIlaX:bjfg922) ([raw file](https://github.com/sillsdev/libpalaso/blob/13967b58901bd8b69c3f8092823ef17e10cfd368/SIL.Core/Reporting/UsageReporter.cs#L313)):*
> ```Smalltalk
> 		{
> 			//ErrorReportSettings.Default.AnalyticsCookie = cookie; /*TODO: how to serialize the cookie?*/
> 			//ErrorReportSettings.Default.Save();
> ```

Hmm, if we were actually saving the analytics cookie (which would give us the ability to track usage accross sessions) this would be an issue. You can't access ErrorReportSettings.Default from SIL.Core (since it exists in SIL.Desktop and is derived from a windows Forms base class)

I guess it is safe, if unfortunate. It does prevent us from actually implementing the TODO here.

In the future we should probably be migrating the settings to something more modern and cross platform.

andrew-polk:
After merging, I realized this is, in fact, a breaking change because any project which has a dependency which uses one of these classes breaks.
I therefore reverted the change and resubmitted on the feature/nuget branch as #875.

PR #878, opened 2019-11-27
Improvements in ExceptionHandler

ermshiperete:
##  [SIL.Core] Deprecate ExceptionHandler.Init()

- Deprecate `ExceptionHandler.Init()`. Clients should use `ExceptionHandler.Init(ExceptionHandler)` 
  instead.
- Move `HandleUnhandledException()` to base class
- Add some unit tests

### Why deprecate `ExceptionHandler.Init()`?

In `ExeceptionHandler.Init()` so far we used reflection to load `WinFormsExceptionHandler` from `SIL.Windows.Forms.dll`. If we couldn't load that assembly we used `ConsoleExceptionHandler` instead. While this is convenient for the client it can hide problems when the client partially updates to use a newer version of the libpalaso assemblies. In particular the client app seems to work unless it gets an exception, in which case it still works but doesn't show the expected green screen for error reporting. Using `ExceptionHandler.Init(new WinFormsExceptionHandler())` instead will allow to catch the incompatibilities at compile time.

## [SIL.Core] Improve ConsoleExceptionHandler.DisplayError

Output at least some information in DisplayError.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/878)

### DIFF
--- a/SIL.Core/Reporting/ExceptionHandler.cs
+++ b/SIL.Core/Reporting/ExceptionHandler.cs
@@ -24,33 +24,33 @@ namespace SIL.Reporting
 		//ExceptionHandler from there through Reflection. Otherwise we will simply use a console
 		//exception handler
 		/// <summary>
-		/// Initialize the ExceptionHandler. By default, the exceptionhandler will be initialized with a ConsoleExceptionHandler
-		/// unless he entry assembly uses a dependency on SIL.Windows.Forms.dll. In that case we default to the WinFormsExceptionHandler
+		/// Initialize the ExceptionHandler. By default, the exception handler will be initialized with a ConsoleExceptionHandler
+		/// unless the entry assembly uses a dependency on SIL.Windows.Forms.dll. In that case we default to the WinFormsExceptionHandler
 		/// </summary>
+		[Obsolete("Use Init(ExceptionHandler) instead, e.g. Init(new ConsoleExceptionHandler()) or Init(new WinFormsExceptionHandler())")]
 		public static void Init()
 		{
-			if (_singleton == null)
-			{
-				//If we can't find the WinFormsExceptionHandler we'll use the Console
-				_singleton = GetObjectFromSilWindowsForms<ExceptionHandler>() ?? new ConsoleExceptionHandler();
-			}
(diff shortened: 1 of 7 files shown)


SIL.Core/Reporting/ExceptionHandler.cs
```diff
@@ -24,33 +24,46 @@ public abstract class ExceptionHandler
...
+			_singleton = GetObjectFromSilWindowsForms<ExceptionHandler>() ?? new ConsoleExceptionHandler();
 		}
 
 		/// <summary>
-		/// Use this method if you want to use an exeption handler besides the default.
+		/// Initialize the ExceptionHandler.
+		/// </summary>
+		/// <typeparam name="T">Type of exception handler</typeparam>
+		/// <exception cref="InvalidOperationException">If ExceptionHandler.Init has been called before</exception>
+		public static void Init<T>() where T: ExceptionHandler, new()
```
tombogle: I'm not entirely sure I see the benefit of having this version of Init. The client could just as easily construct the handler itself and pass it in, I think. I believe the only real point of having the Init method with no parameters was to prevent breaking existing apps, so why not just deprecate the plain Init() method and encourage clients to pass one in explicitly? In the particular case of the HearThis problem we're trying to address, none of the three Init methods would work anyway because the assemblies were incompatible. (Of course, either the template version or the one with an explicit handler passed as a parameter would actually throw an exception, so it would be easier for the developer to identify the problem.)

SIL.Core/Reporting/ExceptionHandler.cs
```diff
@@ -37,19 +37,6 @@ public static void Init()
...
-		public static void Init<T>() where T: ExceptionHandler, new()
-		{
-			if (_singleton != null)
-				throw new InvalidOperationException($"An ExceptionHandler (of type ${_singleton.GetType()}) has already been set.");
-
-			_singleton = new T();
-		}
-
 		/// <summary>
 		/// Use this method if you want to use an exception handler besides the default.
```
tombogle: This comment should probably be changed now because it's really the only non-deprecated version.

PR #881, opened 2019-12-05
Add ability to exclude language tags with scripts

StephenMcConnel:
For example, when asked to choose a language, users can be confused by
seeing "en-Brai", "en-Desr", "en-Dupl", and "en-Shaw" in addition to "en"
in the list.  See https://issues.bloomlibrary.org/youtrack/issue/BL-7641.
The dialog has a means to manually add a script to the tag if it is
needed.
This change does not involve any UI or the need for any project to change
its code unless it wants this new functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/881)


jasonleenaylor:
Also, there are quite a number of unit tests in this area. It would be good to see one for the new behavior so we don't accidentally break it in the future.

PR #886, opened 2020-01-16
Handle multiple unsupported keyboards

jasonleenaylor:
* Don't crash if two unsupported keyboards are
  found in an ldml file
* Support round tripping unsupported keyboards to ldml
* Refactor NullKeyboard functionality to allow detecting an
  unsupported keyboard and round tripping

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/886)


jasonleenaylor:
I wasn't super happy with it either. But there are two competing concerns here. Let's discuss it on slack and see if we can find a good solution.

PR #898, opened 2020-02-12
SP-2133: Crash when editing participant

phillip-hopper:
If a role is not selected for the participant, a "reentrant call" exception may be thrown when setting focus on the role cell.

This solution was found on S.O. - https://stackoverflow.com/questions/5114668/why-is-my-bound-datagridview-throwing-an-operation-not-valid-because-it-results

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/898)

### DIFF
--- a/SIL.Windows.Forms/ClearShare/WinFormsUI/ContributorsListControl.cs
+++ b/SIL.Windows.Forms/ClearShare/WinFormsUI/ContributorsListControl.cs
@@ -232,7 +232,9 @@ namespace SIL.Windows.Forms.ClearShare.WinFormsUI
 					var rc = _grid.GetCellDisplayRectangle(col, e.RowIndex, true);
 					var pt = new Point(rc.X + (rc.Width / 2), rc.Y + 4);
 					_msgWindow.Show(kvp.Value, _grid.PointToScreen(pt));
-					_grid.CurrentCell = _grid[col, e.RowIndex];
+
+					// Invoking here because of "reentrant call to the SetCurrentCellAddressCore" exception.
+					BeginInvoke((Action)(() =>_grid.CurrentCell = _grid[col, e.RowIndex]));
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
___
*[SIL.Windows.Forms/ClearShare/WinFormsUI/ContributorsListControl.cs, line 236 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/898#-M0KD0Sj9FzT5riHWzHB:-M0KD0Sj9FzT5riHWzHC:bt8n8gp) ([raw file](https://github.com/sillsdev/libpalaso/blob/e317698bbdcd64d32eebf9b093d68aa512895233/SIL.Windows.Forms/ClearShare/WinFormsUI/ContributorsListControl.cs#L236)):*
> ```Smalltalk
> 					_msgWindow.Show(kvp.Value, _grid.PointToScreen(pt));
> 
> 					// Invoking here because of "reentrant call to the SetCurrentCellAddressCore" exception.
> ```

So is this necessary because setting _grid.CurrentCell can trigger validation? If that is the case then the comment can be made more specific and clear. And even if that is true though this solution feels like a band-aid.

tombogle:
___
*[SIL.Windows.Forms/ClearShare/WinFormsUI/ContributorsListControl.cs, line 236 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/898#-M0KD0Sj9FzT5riHWzHB:-M0T8mnu0cCDbao6vFxh:b-zgth0q) ([raw file](https://github.com/sillsdev/libpalaso/blob/e317698bbdcd64d32eebf9b093d68aa512895233/SIL.Windows.Forms/ClearShare/WinFormsUI/ContributorsListControl.cs#L236)):*

So is this necessary because setting _grid.CurrentCell can trigger validation? If that is the case then the comment can be made more specific and clear. And even if that is true though this solution feels like a band-aid.
</blockquote></details>

I agree that the comment could be improved. Unfortunately the DataGridView is not one of Microsoft's finest feats. In Glyssen, we have some convoluted code to detect whether we are in EndEdit and we son't set the CurrentCell if we are. It might be an attempt to address this same weakness, but if it is, it's no less of a band aid. We looked at the option of just omitting this line, but I think it really is desirable to take the user to the column that has the problem. So unless you can suggest a better way, I'm okay with Phil's proposed fix.

PR #899, opened 2020-02-13
Changed unit tests to use NUnit 3.12

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/899)


SIL.Lift/PalasoDataObject.cs
```diff
@@ -439,7 +439,7 @@ public static string GetEmbeddedXmlNameForProperty(string name)
 		public override bool Equals(Object obj)
 		{
 			if (!(obj is PalasoDataObject)) return false;
-			return Equals((PalasoDataObject)obj);
+				return Equals((PalasoDataObject)obj);
```
ermshiperete: nit: wrong (and confusing) indentation

tombogle: Fixed

build/Palaso.proj
```diff
@@ -86,14 +87,24 @@
...
+		<NUnit3 Condition="'$(teamcity_version)' == ''"
+			Assemblies="@(TestAssemblies)"
+			ToolPath="$(RootDir)/packages/NUnit.ConsoleRunner.3.10.0/tools"
 			ExcludeCategory="$(ExtraExcludeCategories)$(excludedCategories)"
 			WorkingDirectory="$(RootDir)/output/$(Configuration)"
 			Force32Bit="$(useNUnit-x86)"
 			Verbose="true"
-			OutputXmlFile="$(RootDir)/output/$(Configuration)/TestResults.xml"/>
+			OutputXmlFile="$(RootDir)/output/$(Configuration)/TestResults.xml"
+      UseNUnit3Xml = "true"
```
ermshiperete: nit: indentation wrong

build/Palaso.proj
```diff
@@ -86,14 +87,24 @@
...
+			TeamCity="false"/>
+		<NUnit3 Condition="'$(teamcity_version)' != ''"
```
ermshiperete: It seems that it should be possible to combine the two TC and non-TC calls since they are basically identical.

tombogle: Maybe so. I copied this pretty much verbatim from HearThis. I'll have a look.

tombogle: I removed one of the three differences ("SkipOnTeamCity" was already in ExtraExcludeCategories, when run on TC). Concerning the other two differences: The TeamCity true/false should be straightforward, but I wasn't sure how to leave OutputXmlFile undefined for the TC build.

ermshiperete: Does it matter if we generate a TestResults.xml file on TC? We don't need it there, but it probably doesn't matter if it exists.

tombogle:
If I'm not mistaken, the point of those lines is to get into the parent folder of the build folder and then use that as the root dir. Since it will be a full path, it should work fine. I'm not sure how to run it to test it out. If I try to run it by double-clicking, it fails immediately because it can't find msbuild. If I try to run it from the VS dev prompt (the way I run the BAT files), it can't find Bash to run the script. I think I'll just leave it. If it really is wrong, then someone who knows how to run it should fix it and make sure it works.

build/Palaso.proj
```diff
@@ -73,27 +76,20 @@
...
+			Assemblies="@(TestAssemblies)"
+			ToolPath="$(RootDir)/packages/NUnit.ConsoleRunner.3.11.1/tools"
 			ExcludeCategory="$(ExtraExcludeCategories)$(excludedCategories)"
 			WorkingDirectory="$(RootDir)/output/$(Configuration)"
 			Force32Bit="$(useNUnit-x86)"
 			Verbose="true"
-			OutputXmlFile="$(RootDir)/output/$(Configuration)/TestResults.xml"/>
+			OutputXmlFile="$(TestOutputXmlFile)"
+			UseNUnit3Xml = "true"
+			TeamCity="$(TeamCity)"/>
```
ermshiperete: Another option would have been: `TeamCity="'$(teamcity_version)' != ''"`. But this works just as well.

ermshiperete:
> If I'm not mistaken, the point of those lines is to get into the parent folder of the build folder and then use that as the root dir. Since it will be a full path, it should work fine. I'm not sure how to run it to test it out. If I try to run it by double-clicking, it fails immediately because it can't find msbuild. If I try to run it from the VS dev prompt (the way I run the BAT files), it can't find Bash to run the script. I think I'll just leave it. If it really is wrong, then someone who knows how to run it should fix it and make sure it works.

No, it's not wrong, it's just unnecessary. It sets a default root dir similar to what the two .bat files did. This is a left-over from previously where the .proj file didn't set a fallback. But we can leave that cleanup for later.

tombogle:
___
*[build/Palaso.proj, line 98 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/899#-M0NplqPSbvydvD3goWR-r3-98:-M0YLi4rDb_Roxfr0H3e:b-khf4qz) ([raw file](https://github.com/sillsdev/libpalaso/blob/613c8f7575065b5ceacda09fb83b1ca4a9e45ff4/build/Palaso.proj#L98)):*

nit: indentation wrong
</blockquote></details>

Do you mean the spaces around the =, or the indentation of the line. I removed the spaces, but  all the attributes are indented the same.

tombogle:
___
*[build/Palaso.proj, line 100 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/899#-M0NplqPSbvydvD3goWR-r3-100:-M0YM2JI2wx8JvvGMXsf:bz7qsww) ([raw file](https://github.com/sillsdev/libpalaso/blob/613c8f7575065b5ceacda09fb83b1ca4a9e45ff4/build/Palaso.proj#L100)):*

Does it matter if we generate a TestResults.xml file on TC? We don't need it there, but it probably doesn't matter if it exists.
</blockquote></details>

I found a way to not generate it. Probably not a big deal either way, but no need to add stuff that will just take up space.

tombogle:
___
*[build/Palaso.proj, line 92 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/899#-M0NplqPSbvydvD3goWR-r6-92:-M0YOMvc-RM4A1gNZ_1Z:b4dgkfg) ([raw file](https://github.com/sillsdev/libpalaso/blob/e21ea73814c0a9fd6735427771abf955e58d9f6a/build/Palaso.proj#L92)):*

Another option would have been: `TeamCity="'$(teamcity_version)' != ''"`. But this works just as well.
</blockquote></details>

I tried that and it produced an error

tombogle:
___
*[build/Palaso.proj, line 92 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/899#-M0NplqPSbvydvD3goWR-r6-92:-M0YOWV92UrqqVuaJV3s:b7ha6os) ([raw file](https://github.com/sillsdev/libpalaso/blob/e21ea73814c0a9fd6735427771abf955e58d9f6a/build/Palaso.proj#L92)):*

I tried that and it produced an error
</blockquote></details>

The attribute seems to get passed verbatim (as text) to the build task, so it doesn't resolve to a Boolean.

PR #909, opened 2020-03-03
Fix reporting NUnit 3 tests on TC

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/909)

### DIFF
--- a/build/Palaso.proj
+++ b/build/Palaso.proj
@@ -88,10 +88,7 @@
 			Force32Bit="$(useNUnit-x86)"
 			Verbose="true"
 			OutputXmlFile="$(TestOutputXmlFile)"
-			UseNUnit3Xml="true"
 			TeamCity="$(TeamCity)"/>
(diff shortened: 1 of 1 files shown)


build/Palaso.proj
```diff
@@ -87,11 +87,7 @@
 			WorkingDirectory="$(RootDir)/output/$(Configuration)"
 			Force32Bit="$(useNUnit-x86)"
 			Verbose="true"
-			OutputXmlFile="$(TestOutputXmlFile)"
```
tombogle: If you're getting rid of TestOutputXmlFile, you can get rid of the lines above where it gets set.

ermshiperete: That's true. Unfortunately it doesn't work the way I had hoped it would - TC still doesn't see the individual tests. Need to do some more investigations when I find some time (and add this line back in).

ermshiperete: The thing to get it working was to install the two NUnit extension nuget packages.

PR #916, opened 2020-03-16
Fix bug where SimpleCustom collation could not be removed

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/916)


jasonleenaylor:
The flaky test on linux from the first build is not a result of this change, and can wait for a later fix.

PR #917, opened 2020-03-18
+semver:minor Added XmlSerializationHelper.Serialize<T> and XmlSerializationHelper.Deserialize<T>

tombogle:
This allows for serialization/deserialization of objects using TextWriter/TextReader.
Also added unit tests for these new methods (none existed previously for this class).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/917)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.DblBundle] TextBundle<TM, TL>.GetVersification (to replace deprecated CopyVersificationFile)
 - [SIL.DblBundle] TextBundle<TM, TL>.GetFonts (to replace deprecated CopyFontFiles)
 - [SIL.DblBundle] TextBundle<TM, TL>.GetLdml (to replace deprecated CopyLdmlFile)
+- [SIL.Scripture] ScrVers.Save overload to allow serialization to a TextWriter.
+- [SIL.Core] XmlSerializationHelper.Serialize<T> to allow serialization to a TextWriter.
+- [SIL.Core] XmlSerializationHelper.Deserialize<T> to allow deserialization from a TextReader.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


tombogle:
___
*[CHANGELOG.md, line 25 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/917#-M2k5uFdByuzYy-a5j9N:-M2mgJkjC7YjhGgAT7Lw:br57mdq) ([raw file](https://github.com/sillsdev/libpalaso/blob/2ff46d4d858c4f201f4944f156f97a71f061169f/CHANGELOG.md#L25)):*

> ```
> [SIL.Scripture] ScrVers.Save overload to allow serialization to a TextWriter.
> ```
I don't see this change.
</blockquote></details>

It's at line 185. This was from a previous PR (#914). Apparently, the change I made to this file at that time was not committed and so it didn't make it into that PR. Sorry for the confusion.

tombogle:
___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 525 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/917#-M2k6xuo9VRxDWVlEy-z:-M2mh-By5xdTi2dIb0pB:b-qnd5dc) ([raw file](https://github.com/sillsdev/libpalaso/blob/2ff46d4d858c4f201f4944f156f97a71f061169f/SIL.Core/Xml/XmlSerializationHelper.cs#L525)):*

My thoughts on dispose (applies here and to `Serialize()`):
- The comment is a little misleading. It implies that it won't disposed unless asked to. Rather the opposite is true.
- Other methods in this class do not do any disposing of XMLReaders or XMLWriters. That would be an argument in favor of not having a default for the `dispose` parameter.

</blockquote></details>

* The comment is not necessarily incorrect. You have to ask it to. But the default is to ask it to. :-) But, yeah, it could be worded: "..., unless requested not to."
* I agree that at first it seems a bit out of character, but actually these are the only two public methods on a public class in this file that have a reader/writer passed to them. @ermshiperete thought it was okay (assuming he looked at that carefully). It is admittedly a decision made out of convenience, but as we discussed when we were reviewing it back when these methods were statics in Project, it does feel like a reasonable default since in every current case we're using it, there is no reason why the caller would want to hold onto it. Because of the exception handling logic involved, it is cleaner (on the calling side) to let these methods handle it. (That said, I can live with the decision either way.)

PR #919, opened 2020-03-24
Update DialogAdapters package for mono5-sil compatibility

StephenMcConnel:
Note that this version of DialogAdapters limits Geckofx use to Geckofx45
and before.  (DialogAdapters 1.6.* limit Geckofx use to Geckofx60 and
later!)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/919)


ermshiperete:
I must be blind - I don't see the connection between DialogAdapters and Geckofx. I can see no dependency on Geckofx in DialogAdapters.

ermshiperete:
Oh, I see. DialogAdapters uses GTK, and Geckofx uses GTK, so the two have to use the GTK same version.

PR #923, opened 2020-04-02
LT-19930: Report correct .NET version (for 4)

papeh:
Environment.Version is supposed to return the version of the
common language runtime, not .NET. Also, for .NET Framework 4.5
and later, Microsoft does not recommend using Environment.Version
to determine the version, but instead recommends
"community-maintained" command-line tools that query the registry.
The registry query is tied to v4, so this will need to be rewritten for
.NET 5.0 at the latest (perhaps Microsoft will hide it again sooner).

* Label Environment.Version "CLR Version (deprecated)"
* Query the registry for the .NET (v4) Version
* Fix some warnings

This partially addresses https://jira.sil.org/browse/LT-19930

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/923)

### DIFF
--- a/ExtractCopyright/Program.cs
+++ b/ExtractCopyright/Program.cs
@@ -1,8 +1,7 @@
-﻿// Copyright (c) 2017 SIL International
+// Copyright (c) 2017-2020 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 using System;
 using System.IO;
-using System.Diagnostics;
 
 namespace SIL.ExtractCopyright
 {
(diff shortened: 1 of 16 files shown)


jasonleenaylor:
___
*[SIL.Windows.Forms.Keyboarding/Windows/WindowsKeyboardSwitchingAdapter.cs, line 77 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/923#-M3vaytU-Xrot0XR-kjy:-M3vaytU-Xrot0XR-kjz:b-sojt0h) ([raw file](https://github.com/sillsdev/libpalaso/blob/dcc19528d7910a9496da5c56b6490f871cbbbc3b/SIL.Windows.Forms.Keyboarding/Windows/WindowsKeyboardSwitchingAdapter.cs#L77)):*
> ```Smalltalk
> 				IsSwitchingKeyboards = false;
> 			}
> #if MONO
> ```

If I recall `#If MONO` doesn't work in newer versions of Mono (6?). We shouldn't be adding new uses of it.

PR #926, opened 2020-04-02
Update readmes

marksvc:
* Removing the External Dependencies section of README.Linux.txt.
  - It mentions libicu-cil, which doesn't seem to exist. Perhaps this
    was superceded by icu.net anyway.
  - The libicu dependency is listed on the wiki page pointed to by
    README.md.
* Moving the API Policy section of README.Linux.txt into README.md.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/926)


README.md
```diff
@@ -25,4 +25,14 @@ To get the source code, you'll need Git. Then from a command line, give this com
...
 For instructions on building and contributing, see <https://github.com/sillsdev/libpalaso/wiki/Development>.
+
+API Policy
+----------
+
+The palaso library is still under active development the current deprecation policy is:
+
+- public API's that become deprecated will be marked as obsolete, with an explanation in the default branch.
+- The API and its obsolete tag will remain through the beta.
+- The API will be removed when the library moves to stable.
```
ermshiperete, line 38: This seems fairly outdated and confusing given that we're at version 6. How about we replace it with something like:

The palaso library follows semantic versioning. For APIs this means:

- public APIs that become deprecated will be marked as obsolete.
- The API and its obsolete tag will remain through the beta versions and at least one more stable version.
- The API will be removed in one of the next stable versions.

PR #943, opened 2020-05-20
Use the SIL fork of NAudio for large file support

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/943)

### DIFF
--- a/SIL.Media/SIL.Media-Designer.csproj
+++ b/SIL.Media/SIL.Media-Designer.csproj
@@ -236,13 +236,6 @@
       <Project>{53cb36d9-2f66-4823-b1f0-4523e8565200}</Project>
       <Name>SIL.Core</Name>
     </ProjectReference>
-    <ProjectReference Include="SIL.Media.csproj">
-      <Project>{66B3504A-3B4E-45CB-903A-A9A75B22EF68}</Project>
-      <Name>SIL.Media</Name>
-    </ProjectReference>
-  </ItemGroup>
-  <ItemGroup>
-    <Content Include="Changes.xml" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it.
--- a/SIL.Media/SIL.Media-Designer.csproj
+++ b/SIL.Media/SIL.Media-Designer.csproj
@@ -66,15 +66,16 @@
     <Reference Include="irrKlang.NET4">
       <SpecificVersion>False</SpecificVersion>
     </Reference>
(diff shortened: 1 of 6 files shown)


SIL.Media/SIL.Media-Designer.csproj
```diff
@@ -235,6 +236,13 @@
       <Project>{53cb36d9-2f66-4823-b1f0-4523e8565200}</Project>
       <Name>SIL.Core</Name>
     </ProjectReference>
+    <ProjectReference Include="SIL.Media.csproj">
```
ermshiperete, line 239: Do we need this project reference?

SIL.Media/SIL.Media.csproj
```diff
@@ -34,7 +34,9 @@ See full changelog at https://github.com/sillsdev/libpalaso/blob/feature/nuget/C
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
-    <PackageReference Include="NAudio" Version="1.8.4" />
+    <PackageReference Include="SIL.NAudio" Version="1.9.0-beta0074">
+      <PrivateAssets>all</PrivateAssets>
```
ermshiperete, line 38: Do we really want `<PrivateAssets>all</PrivateAssets>` for `SIL.NAudio`? I'd think that we need the assembly at runtime, not just during the build.

ermshiperete (changes requested):
Would be nice to have a released version of SIL.NAudio before we release libpalaso 9.0

tombogle:
After talking with Zac Glenn, he thinks that before we defnitely commit to this, it would be good to take one more crack at seeing if we can get the official NAudio master to build and try to get our failing Nunit tasks into a branch so we can demonstrate that the bug still exists (or find out that it doesn't). if it does, we can create an issue in their guthub, and then he would be willing to negotiate with them if necessary to get it fixed (we could easily do the work I think).

imnasnainaec:
Can this branch be deleted?

tombogle:
Most likely. I *thought* there was still something that relied on our fork of NAudio, but I've searched and found nothing. On Nuget, we have only one package that is not marked as "pre-release" but even that one has "beta" in it's name, so that's highly suspect. And the project is archived on TeamCity.
It's a rather fuzzy memory, but I think the version(s) of 1.x that we use did incorporate our fixes for large audio files, but then maybe that got broken again in v. 2.
I do have a task on my to-do list to see if we could upgrade to NAudio v. 3, [which claims to restore full support for large audio files](https://github.com/naudio/NAudio/pull/437#event-25835182944). This seems to be a pretty major breaking change but could be really beneficial.

PR #948, opened 2020-06-17
Update NAudio to 1.10.0

phillip-hopper:


SIL.Media/SIL.Media-Designer.csproj
```diff
@@ -85,12 +85,12 @@
     </None>
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="AudioIrrKlangSession.cs" />
+    <Compile Include="IAudioPlayer.cs" />
```
phillip-hopper, line 88: It looks like some of the files in SIL.Media were moved around, but the Designer project was not updated.

PR #949, opened 2020-06-18
Fix keyboard switching on Ubuntu 18.04 (fixes #887)

### LINKED ISSUE #887: Keyboard switching doesn't seem to work in Ubuntu 18.04 (Bionic)
At least in test app, haven't tried in FW.

ermshiperete:
Ubuntu 18.04 changed the way how to switch keyboards. The previous
method of setting a dconf value is deprecated and no longer works.
Instead this is done by a DBus method call.

This change also adds a property to detect if the program runs in
a Gnome Shell, and fixes a bug where the wrong keyboard adaptor was
used in Ubuntu 18.04.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/949)

### DIFF
--- a/SIL.Windows.Forms.Keyboarding/Linux/GnomeShellIbusKeyboardSwitchingAdaptor.cs
+++ b/SIL.Windows.Forms.Keyboarding/Linux/GnomeShellIbusKeyboardSwitchingAdaptor.cs
@@ -1,0 +1,54 @@
+// Copyright (c) 2020 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+
+using System;
+using System.Diagnostics;
+using SIL.Reporting;
+
+namespace SIL.Windows.Forms.Keyboarding.Linux
+{
+	/// <summary>
+	/// Class for dealing with ibus keyboards on Gnome Shell (as found in Ubuntu Bionic >= 18.04)
+	/// </summary>
+	public class GnomeShellIbusKeyboardSwitchingAdaptor: UnityIbusKeyboardSwitchingAdaptor, IUnityKeyboardSwitchingAdaptor
+	{
+		public GnomeShellIbusKeyboardSwitchingAdaptor(IIbusCommunicator ibusCommunicator)
+			: base(ibusCommunicator)
+		{
+		}
+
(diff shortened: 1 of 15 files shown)


SIL.Windows.Forms.Keyboarding/Linux/GnomeShellIbusKeyboardSwitchingAdaptor.cs
```diff
@@ -0,0 +1,54 @@
...
+		{
+		}
+
+		void IUnityKeyboardSwitchingAdaptor.SelectKeyboard(uint index)
+		{
+			var okay = false;
+			// https://askubuntu.com/a/1039964
+			try
+			{
+				using (var proc = new Process {
```
marksvc, line 26: This is broken up very readably, thank you.

SIL.Windows.Forms.Keyboarding/Linux/UnityXkbKeyboardSwitchingAdaptor.cs
```diff
@@ -16,11 +16,13 @@ public UnityXkbKeyboardSwitchingAdaptor(IXklEngine engine): base(engine)
...
 			var xkbKeyboard = keyboard as XkbKeyboardDescription;
-			if (xkbKeyboard != null)
-			{
-				if (xkbKeyboard.GroupIndex >= 0)
-					UnityIbusKeyboardSwitchingAdaptor.SelectKeyboard((uint)xkbKeyboard.GroupIndex);
-			}
+			if (xkbKeyboard == null || xkbKeyboard.GroupIndex < 0)
+				return;
+
+			var switchingAdaptor = KeyboardController.Instance
```
marksvc, line 22: I'm actually a bit confused about how this is supposed to be working, and am having trouble figuring it out from just the github UI. I need to look at it in an IDE.

marksvc: So how do you know that `SwitchingAdaptor` will be an `IUnityKeyboardSwitchingAdaptor` and not result in a null reference exception on the next line?

I think I'm confused about why the `UnityXkbKeyboardSwitchingAdaptor` would be being used anyway if keyboard switching is being done thru ibus.

PR #951, opened 2020-06-19
LT-20211 fix 2 crashes related to getting/switching keyboards

jwickberg:


SIL.Windows.Forms.Keyboarding/Windows/WinKeyboardAdaptor.cs
```diff
@@ -191,10 +191,19 @@ private IEnumerable<LayoutName> GetAvailableKeyboardNames(InputLanguage inputLan
 				LayoutName layoutName;
 				if (profiles[0].Hkl == IntPtr.Zero)
 				{
+					try
+					{
+						layoutName = new LayoutName(inputLanguage.LayoutName,
+							ProcessorProfiles.GetLanguageProfileDescription(ref profiles[0].ClsId, profiles[0].LangId,
+								ref profiles[0].GuidProfile), profiles[0]);
+					}
+					catch
```
jasonleenaylor, line 200: Can we at least catch the specific exception type?

jasonleenaylor: Actually, we can make that improvement in a follow up commit. We want to look at a better user experience for this in any case.

jasonleenaylor:
LGTM

PR #952, opened 2020-06-25
Added missing mapping to eng.vrs to handle Hebrew subtitle in Psalm 11

tombogle:
Pending approval by Sarah Lind...


tombogle:
From Sarah Lind:
Hi Tom. Unfortunately, Paratext mapping is not able to handle partial verses. That is what's going on in Psalm 11. The heading is the first two words of v. 1, and then the psalm starts in the same verse. When the heading is in its own verse, there's no problem with mapping.
Sarah Lind
United Bible Societies
The Bible Translator managing editor
Paratext source language resources editor

PR #953, opened 2020-06-26
Improve keyboard switching on Ubuntu 18.04

ermshiperete:
This is a follow-up of PR #949. This change introduces a combined
keyboard for keyboards in GnomeShell. With that we're able to
undo some of the previous changes from #949. This change also adds
a bunch of unit tests for the GnomeShell*Adaptor classes.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/953)


SIL.Windows.Forms.Keyboarding/Linux/GnomeShellIbusKeyboardRetrievingAdaptor.cs
```diff
@@ -10,13 +17,134 @@ namespace SIL.Windows.Forms.Keyboarding.Linux
...
+			KeyboardRetrievingHelper.AddIbusVersionAsErrorReportProperty();
+			InitKeyboards();
+		}
+
+		protected override void InitKeyboards()
+		{
+			_helper.InitKeyboards(type => true, RegisterKeyboards);
+		}
+
+		private void RegisterKeyboards(IDictionary<string, uint> keyboards, (string type, string layout) firstKeyboard)
```
marksvc, line 60: I found I had to keep looking back and forth around this method to try to understand what it was doing, and what the difference is between `keyboards`  and  `curKeyboards`, or which was what meaning. 
I think probably the foreach loop over all ibus keyboards contributed to having trouble understanding at first.

Could we either give a brief method comment saying what the general activity of this method's algorithm is, or sprinkle in some comments inside the method to describe the purpose for sections or some of the contents of the loop. 

It would also help to rename `keyboards` to something like `requestedKeyboards`.

Maybe it could have a method comment that says something like  "Activate all ibus or xkb keyboards in KeyboardController that are requested in the `keyboards` list. Deactivate the rest. Add any keyboards to KeyboardController that were missing."

FWIW, I am leaving my analysis of what the foreach loop looks to be doing:

```
for each ibusKeyboard that ibus knows about: 

  skip processing if it is not in the request list (based on longname or xkb layout name)
  skip processing if we already processed a keyboard with this name/layout.

  if `curKeyboards` contains ibusKeyboard
    make sure it is set as available
    remove it from the `curKeyboards` list so we can later deactivate the rest
  if `curKeyboards` did not contain ibusKeyboard
    add the ibusKeyboard to KeyboardController (but maybe not necessarily to ibus's list of keyboards)
```

SIL.Windows.Forms.Keyboarding/Linux/GnomeShellIbusKeyboardSwitchingAdaptor.cs
```diff
@@ -3,22 +3,29 @@
...
+
 		public GnomeShellIbusKeyboardSwitchingAdaptor(IIbusCommunicator ibusCommunicator)
 			: base(ibusCommunicator)
 		{
 		}
 
-		void IUnityKeyboardSwitchingAdaptor.SelectKeyboard(uint index)
+		protected override void SelectKeyboard(KeyboardDescription keyboard)
 		{
+			var ibusKeyboard = (IbusKeyboardDescription) keyboard;
```
marksvc, line 26: This cast looks like it could fail. Do we know it will always be an object of class IbusKeyboardDescription?

SIL.Windows.Forms.Keyboarding/Linux/GnomeXkbInfo.cs
```diff
@@ -0,0 +1,55 @@
...
+			return array;
+		}
+
+		public bool GetLayoutInfo(string id, out string displayName, out string shortName,
+			out string xkbLayout, out string xkbVariant)
+		{
+			// var displayNamePtr = IntPtr.Zero;
+			// var shortNamePtr = IntPtr.Zero;
+			// var xkbLayoutPtr = IntPtr.Zero;
+			// var xkbVariantPtr = IntPtr.Zero;
```
marksvc, line 34: (Delete these 4 lines?)

PR #974, opened 2020-10-02
Added overload for SerializeToString to specify encoding

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/974)

### DIFF
--- a/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
+++ b/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
@@ -417,7 +417,11 @@ namespace SIL.Tests.Xml
 		public void SerializeToString_NoEncodingSpecified_XmlHeaderHasDefaultUtf16Encoding()
 		{
 			var result = XmlSerializationHelper.SerializeToString(new TestObject("Fred"));
-			Assert.That(result.Contains("encoding=\"utf-16\""));
+			var lines = result.Split(new[] {'\r', '\n'}, StringSplitOptions.RemoveEmptyEntries).ToList();
+			Assert.IsTrue(lines[0].StartsWith("<?xml"));
+			Assert.That(lines[0].Contains("encoding=\"utf-16\""));
+			Assert.IsTrue(lines[1].StartsWith("<MyRoot"));
+			Assert.IsTrue(lines.Last().EndsWith("MyRoot>"));
 		}
 
 		[Test]
@@ -425,7 +429,22 @@ namespace SIL.Tests.Xml
 		{
 			var result = XmlSerializationHelper.SerializeToString(new TestObject("Fred"),
 				Encoding.UTF8);
-			Assert.That(result.Contains("encoding=\"utf-8\""));
+			var lines = result.Split(new[] {'\r', '\n'}, StringSplitOptions.RemoveEmptyEntries).ToList();
+			Assert.IsTrue(lines[0].StartsWith("<?xml"));
+			Assert.That(lines[0].Contains("encoding=\"utf-8\""));
+			Assert.IsTrue(lines[1].StartsWith("<MyRoot"));
+			Assert.IsTrue(lines.Last().EndsWith("MyRoot>"));
+		}
+
+		[Test]
+		public void SerializeToString_OmitXmlHeading_XmlHeaderHasExpectedEncoding()
+		{
+			var result = XmlSerializationHelper.SerializeToString(new TestObject("Fred"),
+				true);
+			Assert.IsFalse(result.StartsWith("<?xml"));
+			var lines = result.Split(new[] {'\r', '\n'}, StringSplitOptions.RemoveEmptyEntries).ToList();
+			Assert.IsTrue(lines.First().StartsWith("<MyRoot"));
+			Assert.IsTrue(lines.Last().EndsWith("MyRoot>"));
 		}
 	}
 }
--- a/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
+++ b/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
@@ -427,5 +427,17 @@ namespace SIL.Tests.Xml
 				Encoding.UTF8);
 			Assert.That(result.Contains("encoding=\"utf-8\""));
 		}
+
+		/// <summary>
+		/// This tests the (probably somewhat nonsensical) case of asking for a UTF-8
+		/// XML string, but suppressing the header.
+		/// </summary>
+		[Test]
+		public void SerializeToString_EncodingSpecifiedWithXmlHeaderOmitted_XmlHeaderOmitted()
+		{
+			var data = new TestObject("Fred");
+			Assert.AreEqual(XmlSerializationHelper.SerializeToString(data, true, Encoding.UTF8),
+				XmlSerializationHelper.SerializeToString(data, true));
+		}
 	}
 }
--- a/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
+++ b/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
@@ -427,17 +427,5 @@ namespace SIL.Tests.Xml
 				Encoding.UTF8);
 			Assert.That(result.Contains("encoding=\"utf-8\""));
 		}
-
-		/// <summary>
-		/// This tests the (probably somewhat nonsensical) case of asking for a UTF-8
-		/// XML string, but suppressing the header.
-		/// </summary>
-		[Test]
-		public void SerializeToString_EncodingSpecifiedWithXmlHeaderOmitted_XmlHeaderOmitted()
-		{
-			var data = new TestObject("Fred");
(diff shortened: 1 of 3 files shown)


SIL.Core/Xml/XmlSerializationHelper.cs
```diff
@@ -139,27 +139,35 @@ public static byte[] SerializeToByteArray<T>(T data, bool omitXmlDeclaration)
 		/// Serializes an object to an XML string.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		public static string SerializeToString<T>(T data)
+		public static string SerializeToString<T>(T data, Encoding encoding = null)
```
ermshiperete, line 142: This is a breaking change, i.e. a client needs to be recompiled in order to work with this API. This means that the commit message should have a `+semver:major` added.

Or we make it a real overload `public static string SerializeToString<T>(T data, Encoding encoding)` and add `+semver:minor` to the commit message. Same for the other method below.

Depending on what you decide to do the line in the changelog might have to be moved to the `Changed` section.

ermshiperete: I think having a default value is the more elegant solution and since we have an unreleased major version change I'd go with `+semver:major`

ermshiperete (changes requested):

SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs
```diff
@@ -412,5 +412,15 @@ public void DeserializeAndSerialize_Normal_RoundTripObject(bool dispose)
...
 			}
 		}
+
+		[Test]
+		public void SerializeToString_EncodingSpecified_XmlHeaderHasExpectedEncoding()
+		{
+			var testObj = new TestObject("Fred");
+			var result = XmlSerializationHelper.SerializeToString(testObj);
+			Assert.That(result.Contains("encoding=\"utf-16\""), "Baseline test failed");
+			result = XmlSerializationHelper.SerializeToString(testObj, Encoding.UTF8);
```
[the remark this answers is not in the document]
tombogle: I wanted to use two test cases, but the encodings are static objects,not enums, so they can't be passed as values to the parameter of the test method using TestCase. The first one is really just a baseline/regression to show that utf-8 is not the default. In any case, I was trying to throw something together quick just to see what's going on, since I can no longer run unit tests locally :-(

tombogle:
___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 142 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBCA8h05n0AxM5h-r1-142:-MIdOppN9chrAPdLnPxl:b-km3lnf) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core/Xml/XmlSerializationHelper.cs#L142)):*

I think having a default value is the more elegant solution and since we have an unreleased major version change I'd go with `+semver:major`
</blockquote></details>

Good to know. I (probably should have but) didn't know that adding an optional parameter would require it to be major. I started with a separate version but then thought an optional param would be better (except in that it allows the kind of nonsensical combination of asking for a UTF-8 string but no XML header - in that sense maybe a separate method is better).

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/974)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @ermshiperete and @jasonleenaylor)

___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 155 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBCA8h05n0AxM5h-r1-155:-MIdPP5wA0sptY1m1D-H:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core/Xml/XmlSerializationHelper.cs#L155)):*

What happens if `omitXmlDeclaration == true` and an encoding is specified? This shouts for a unit test.
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs, line 422 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBKeX768qHcB0m5-r1-422:-MIdPQR9DhTKMgLPlPHF:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs#L422)):*

I wanted to use two test cases, but the encodings are static objects,not enums, so they can't be passed as values to the parameter of the test method using TestCase. The first one is really just a baseline/regression to show that utf-8 is not the default. In any case, I was trying to throw something together quick just to see what's going on, since I can no longer run unit tests locally :-(
</blockquote></details>

Done.

ermshiperete (changes requested):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/974)* status: all files reviewed, 2 unresolved discussions (waiting on @jasonleenaylor and @tombogle)

___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 142 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBCA8h05n0AxM5h-r1-142:-MIsDx3K3FZgWQgT3s2t:bi2y11n) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core/Xml/XmlSerializationHelper.cs#L142)):*

Good to know. I (probably should have but) didn't know that adding an optional parameter would require it to be major. I started with a separate version but then thought an optional param would be better (except in that it allows the kind of nonsensical combination of asking for a UTF-8 string but no XML header - in that sense maybe a separate method is better).
</blockquote></details>

I didn't know either, but wondered when reviewing this change, so I wrote a little test app and tried it: a library project and a console app. At first things worked until I realized that I shouldn't try to run the console app from within VS because that recompiles the executable and then of course it works :smile: 
___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 155 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBCA8h05n0AxM5h-r1-155:-MIsEtsRDJ46GTdHnJd3:b-6ubm2g) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core/Xml/XmlSerializationHelper.cs#L155)):*

Done.
</blockquote></details>

I don't see a unit test for `omitXmlDeclaration == true` :confused:

tombogle:
___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 155 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBCA8h05n0AxM5h-r1-155:-MIskex03Lk_i2tQ-gaz:bet07za) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core/Xml/XmlSerializationHelper.cs#L155)):*

I don't see a unit test for `omitXmlDeclaration == true` :confused: 
</blockquote></details>

It was preexisting, but I've gone ahead and written a test for it and added a few more sanity-checking lines to the other tests as well.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/974)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @ermshiperete and @jasonleenaylor)

___
*[SIL.Core/Xml/XmlSerializationHelper.cs, line 142 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/974#-MIayjBCA8h05n0AxM5h-r1-142:-MIske23Cjk1AyHqmtER:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/0940ee1dfcd91bd2ca1053a1b904d9b0bf234c92/SIL.Core/Xml/XmlSerializationHelper.cs#L142)):*

I didn't know either, but wondered when reviewing this change, so I wrote a little test app and tried it: a library project and a console app. At first things worked until I realized that I shouldn't try to run the console app from within VS because that recompiles the executable and then of course it works :smile: 
</blockquote></details>

Done.

ermshiperete (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/974)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

ermshiperete (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/974#-:-MIsud2F8HQr-8f9UcqI:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

PR #982, opened 2020-10-28
Added unit tests (intended to include with previous commit on Ocotber 5)

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/982)


papeh:
The test itself looks good.

The other classes etc. seem more complicated than they need to be. Could you help me understand the need for SomethingWeird?

PR #983, opened 2020-10-29
Improve filename checking

tombogle:
This is mostly taken from the implementation in Transcelerator. Needed by Glyssen for PG-1412.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)

### DIFF
--- a/SIL.Core.Tests/Extensions/StringExtensionTests.cs
+++ b/SIL.Core.Tests/Extensions/StringExtensionTests.cs
@@ -171,7 +171,7 @@ namespace SIL.Tests.Extensions
 			Assert.AreEqual(expectedResult, str.RemoveDiacritics());
 		}
 
-		[Test, TestCaseSource(nameof(GetInvalidFilenameCharacters))]
+		[TestCaseSource(nameof(GetInvalidFilenameCharacters))]
 		public void SanitizeFilename_InvalidCharacter_Replaced(char invalidChar)
 		{
 			Assert.AreEqual("File name.txt", $"File{invalidChar}name.txt".SanitizeFilename(' '));
@@ -361,6 +361,8 @@ namespace SIL.Tests.Extensions
 		[TestCase(@"c:\root")]
 		[TestCase(@"c:root")] // This is valid, though some apps might not allow with it.
 		[TestCase(@".git")]
+		[TestCase(@"c:")]
+		[TestCase(@"Z:")]
 		public void SanitizePath_StringContainsCharactersThatAreNotValidFileCharactersButAreValidPathCharacters_NoChange(string orig)
 		{
 			Assert.AreEqual(orig, orig.SanitizePath('_'));
@@ -371,13 +373,13 @@ namespace SIL.Tests.Extensions
 		/// Tests that on Windows improperly placed colons are replaced.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
+		[Platform(Include = "win", Reason = "Rules about colons and slashes only apply to Windows.")]
 		[TestCase(@"\:c", ExpectedResult = @"\_c")]
 		[TestCase(@"c::\", ExpectedResult = @"c:_\")]
 		[TestCase(@"c:\:", ExpectedResult = @"c:\_")]
 		[TestCase(@"%:\yeah", ExpectedResult = @"%_\yeah")]
 		[TestCase(@"abc:", ExpectedResult = "abc_")]
-		[TestCase(@"abc:", ExpectedResult = "abc_")]
-		[Platform(Include = "win", Reason = "Rules about colons and slashes only apply to Windows.")]
+		[TestCase(@"c :", ExpectedResult = "c _")]
 		public string SanitizePath_StringContainsBogusColonsOrSlashes_Sanitized(string orig)
 		{
 			return orig.SanitizePath('_');
@@ -390,6 +392,7 @@ namespace SIL.Tests.Extensions
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[TestCase(@"blah\blah\.", ExpectedResult = @"blah\blah\")] // This would also be acceptable: blah\blah\.
+		[TestCase(@"blah\blah/.", ExpectedResult = @"blah\blah/")] // This would also be acceptable: blah\blah/.
 		[TestCase(@"c:\root\..", ExpectedResult = @"c:\root\..")]
 		[TestCase(@"My directory is awesome.", ExpectedResult = @"My directory is awesome")]
 		[TestCase(".", ExpectedResult = ".")]
@@ -397,7 +400,9 @@ namespace SIL.Tests.Extensions
 		[TestCase(@". .", ExpectedResult = @"_")]
 		[TestCase(@"\.", ExpectedResult = @"\")] // This would also be acceptable: \.
 		[TestCase(@".\..", ExpectedResult = @".\..")]
+		[TestCase(@"./..", ExpectedResult = @"./..")]
 		[TestCase(@".\...", ExpectedResult = @".\")] // REVIEW: maybe we want: .\_
+		[TestCase(@"....", ExpectedResult = @"_")]
 		public string SanitizePath_StringHasTrailingDots_TrailingDotsTrimmed(string orig)
 		{
 			return orig.SanitizePath('_');
--- a/SIL.Core.Tests/Extensions/StringExtensionTests.cs
+++ b/SIL.Core.Tests/Extensions/StringExtensionTests.cs
@@ -239,7 +239,7 @@ namespace SIL.Tests.Extensions
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Tests that final periods and spaces are stripped.
-		/// (See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming_conventions)
+		/// (See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[TestCase(".")]
@@ -260,7 +260,7 @@ namespace SIL.Tests.Extensions
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Tests that final periods and spaces are stripped.
-		/// (See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming_conventions)
+		/// (See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[TestCase(".\u00A0.")]
@@ -273,7 +273,7 @@ namespace SIL.Tests.Extensions
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Tests that leading spaces and formatting characters are stripped.
-		/// (See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming_conventions)
+		/// (See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[TestCase(" ")]
@@ -313,12 +313,13 @@ namespace SIL.Tests.Extensions
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Tests that valid path charcters (: and \) that are invalid in filenames are not
-		/// replaced.
+		/// Tests that valid path characters (: and \ and leading dots) that are invalid in
+		/// filenames are not replaced.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[TestCase(@"..\relative\..\path")]
 		[TestCase(@"c:\root")]
+		[TestCase(@".git")]
 		public void SanitizePath_StringContainsCharactersThatAreNotValidFileCharactersButAreValidPathCharacters_NoChange(string orig)
 		{
 			Assert.AreEqual(orig, orig.SanitizePath('_'));
--- a/SIL.Core.Tests/Extensions/StringExtensionTests.cs
+++ b/SIL.Core.Tests/Extensions/StringExtensionTests.cs
@@ -1,6 +1,9 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework;
 using SIL.Extensions;
+using static System.String;
 
 namespace SIL.Tests.Extensions
 {
@@ -16,7 +19,7 @@ namespace SIL.Tests.Extensions
 		[Test]
 		public void SplitTrimmed_StringIsEmpty_GivesEmptyList()
 		{
-			Assert.AreEqual(0, string.Empty.SplitTrimmed(',').Count);
+			Assert.AreEqual(0, Empty.SplitTrimmed(',').Count);
 		}
 
 		[Test]
@@ -129,7 +132,7 @@ namespace SIL.Tests.Extensions
 		[Test]
 		public void ToIntArray_EmptyString_ReturnsEmptyArray()
 		{
-			Assert.AreEqual(0, string.Empty.ToIntArray().Length);
+			Assert.AreEqual(0, Empty.ToIntArray().Length);
 		}
 
 		[Test]
@@ -168,6 +171,15 @@ namespace SIL.Tests.Extensions
 			Assert.AreEqual(expectedResult, str.RemoveDiacritics());
 		}
 
+		[Test, TestCaseSource(nameof(GetInvalidFilenameCharacters))]
+		public void SanitizeFilename_InvalidCharacter_Replaced(char invalidChar)
+		{
+			Assert.AreEqual("File name.txt", $"File{invalidChar}name.txt".SanitizeFilename(' '));
+		}
+
+		private static IEnumerable<char> GetInvalidFilenameCharacters() =>
+			System.IO.Path.GetInvalidFileNameChars();
+		
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Tests that we get a valid filename when the filename contains invalid characters.
@@ -177,7 +189,29 @@ namespace SIL.Tests.Extensions
 		[TestCase('w', ExpectedResult = "MywwFilewwDudewwwww.'[];funny()www")]
 		public string SanitizeFilename_ManyInvalidCharacters_InvalidCharactersReplacedWithSpecifiedCharacter(char errorChar)
 		{
+			var invalidFilename = @"My?|File<>Dude\?*:/.'[];funny()" + "\u000a\t" +
+				GetInvalidFilenameCharacters().First();
+			var validFileName = invalidFilename.SanitizeFilename(errorChar);
+			Assert.IsFalse(validFileName.Any(c => GetInvalidFilenameCharacters().Contains(c)));
+			Assert.IsTrue(validFileName.Contains(errorChar));
 			return (@"My?|File<>Dude\?*:/.'[];funny()" + "\u000a\t" + '"').SanitizeFilename(errorChar);
+		}
+		
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Tests that we get a valid filename that does not end with spaces when the filename
+		/// ends with invalid characters and the error replacement character is a space.
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[TestCase(' ')]
+		[TestCase('\u200C')]
+		public void SanitizeFilename_LastCharIsInvalidAndReplacementCharIsSpace_SanitizedStringDoesNotEndWithSpace(char errorChar)
+		{
+			var invalidFilename = @"My?|File<>Dude\?*:/.'[];funny" +
+				GetInvalidFilenameCharacters().First();
+			var validFileName = invalidFilename.SanitizeFilename(errorChar);
+			Assert.IsFalse(validFileName.Any(c => GetInvalidFilenameCharacters().Contains(c)));
+			Assert.IsTrue(validFileName.EndsWith("funny"));
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -189,7 +223,10 @@ namespace SIL.Tests.Extensions
 		[TestCase('.', ExpectedResult = ".My..File.Dude..'[];funny()")]
 		public string SanitizeFilename_ReplacingInvalidCharactersWithSpaceOrDot_InvalidCharactersAtStartOrEndOfStringTrimmedIfInvalid(char errorChar)
 		{
-			return (@"<My?|File>Dude\.'[];funny()?*:/").SanitizeFilename(errorChar);
+			var someInvalidCharacters = Join("", GetInvalidFilenameCharacters().Take(5));
+			var invalidFilename = someInvalidCharacters.First() + @"My?|File>Dude\.'[];funny()" +
+				someInvalidCharacters;
+			return invalidFilename.SanitizeFilename(errorChar);
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -207,8 +244,12 @@ namespace SIL.Tests.Extensions
 		public void SanitizeFilename_LeadingAndTrailingSpacesAndDotsPlusInvalidCharacters_InvalidCharactersReplacedAndLeadingAndTrailingJunkTrimmed(
 			string leading, string trailing)
 		{
-			Assert.AreEqual("My__File__Dude_____.'[];funny()____",
-				(leading + @"My?|File<>Dude\?*:/.'[];funny()" + "\u000a\t\uFFFC" + '"' + trailing).SanitizeFilename('_'));
+			var filenameWithLeadingAndTrailingJunk = leading +
+				@"My?|File<>Dude\?*:/.'[];funny()" + GetInvalidFilenameCharacters().Last() + trailing;
+			var sanitizedFilename = filenameWithLeadingAndTrailingJunk.SanitizeFilename('_');
+			Assert.That(sanitizedFilename.StartsWith("My"));
+			Assert.That(sanitizedFilename.EndsWith("funny()_"));
+			Assert.IsFalse(sanitizedFilename.Any(c => GetInvalidFilenameCharacters().Contains(c)));
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -216,17 +257,15 @@ namespace SIL.Tests.Extensions
 		/// Tests that we throw an exception if caller passes a bogus errorChar.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		[TestCase('<')]
-		[TestCase('?')]
+		[Test, TestCaseSource(nameof(GetInvalidFilenameCharacters))]
 		public void SanitizeFilename_ErrorCharInvalid_Throws(char errorChar)
 		{
-			Assert.Throws<ArgumentException>(() =>
-				(@"My?|File<>Dude\?*:/.'[];funny()" + "\u000a\t" + '"').SanitizeFilename(errorChar));
+			Assert.Throws<ArgumentException>(() => @"My?|Dude\?funny".SanitizeFilename(errorChar));
 		}
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Tests that we get a filename with a normal space when the given filename contain a
+		/// Tests that we get a filename with a normal space when the given filename contains a
 		/// non-breaking space.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tombogle:
___
*[SIL.Core/Extensions/StringExtensions.cs, line 167 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuNxCgCE_XW7n6dglK:-MKurBVT7dhNlqz_P2a-:b7gfznq) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core/Extensions/StringExtensions.cs#L167)):*

Why is doing it this way (adding an override) better than an optional parameter?
</blockquote></details>

I think @ermshiperete said that adding an optional parameter actually causes any existing clients to break if they run against the new version (so basically instead of a minor version change, it becomes a major one). Probably not significant until we get this out of beta, but that's why I did it. If I misunderstood or we decide it really doesn't matter, I'd prefer an optional parameter.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @ermshiperete)

___
*[SIL.Core/Extensions/StringExtensions.cs, line 174 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuNnOpCTv6Afsh8pWl:-MKvSVa-D9XdhWqfdOsJ:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core/Extensions/StringExtensions.cs#L174)):*

Something went wrong with this param comment
</blockquote></details>

Done.
___
*[SIL.Core/Extensions/StringExtensions.cs, line 195 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuPGndCGIJ8U0NCmJm:-MKvSy5h57CONl3mX0bN:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core/Extensions/StringExtensions.cs#L195)):*

Should you add a helpful message here?
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 229 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuSrKR4QpEhaDIBLZK:-MKvVfVO3PRfID7bsedJ:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L229)):*

> ```
> Tests that we get a valid filename when the filename contains invalid characters.
> ```
Not exactly accurate is it? 
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 320 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuTp4I9bUPUPAJ6eGX:-MKvVf36DhSb0VFO4hgK:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L320)):*

Incorrect summary
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 338 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuTzFVBIQ_u81z0KgX:-MKvVeUh2EWQHg-Y7xA_:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L338)):*

Incorrect summary
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 343 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuU787CQ12JxPxOYFA:-MKvVhHv7FUOjMfOlpmz:bb74x83) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L343)):*

I don't think I agree with this expected result. What's wrong with the path `..`?
And we should have a test for `.`.
</blockquote></details>

True

andrew-polk (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @tombogle)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 343 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuU787CQ12JxPxOYFA:-ML8RnDMB5Xni7FjPA1K:br5uor8) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L343)):*

True
</blockquote></details>

Sorry, I meant we should have `[TestCase(@".", ExpectedResult = @".")]`

ermshiperete (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: all files reviewed, 6 unresolved discussions (waiting on @tombogle)

___
*[SIL.Core/Extensions/StringExtensions.cs, line 167 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuNxCgCE_XW7n6dglK:-ML8Z6yYDiGFD5AA4dZH:b-blapsk) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core/Extensions/StringExtensions.cs#L167)):*

I think @ermshiperete said that adding an optional parameter actually causes any existing clients to break if they run against the new version (so basically instead of a minor version change, it becomes a major one). Probably not significant until we get this out of beta, but that's why I did it. If I misunderstood or we decide it really doesn't matter, I'd prefer an optional parameter.
</blockquote></details>

Yes, that's correct. I wrote a test app and test lib and found that adding an optional parameter requires a recompile of the test app (if the test app uses the method, IIRC), whereas if we add an overload instead we can replace the binary lib without having to recompile the app.
___
*[SIL.Core/Extensions/StringExtensions.cs, line 224 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8bqIK8HxX3ti5oHzi:-ML8bqIK8HxX3ti5oHzj:bhfy6up) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core/Extensions/StringExtensions.cs#L224)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
> 					(lastCharPos < 2 ||
> 					result[lastCharPos - 1] != '.' ||
> 					(result[lastCharPos - 2] != Path.DirectorySeparatorChar &&
> 					result[lastCharPos - 2] != Path.AltDirectorySeparatorChar)) &&
> 					!(lastCharPos == 1 && result[0] == '.'))
> ```
</details>

I have a hard time wrapping my brain around these nested expressions and understand what we test. I wonder if we could use `result.EndsWith("/.")` or something like that - maybe that would make it easier to understand. Or extract to a variable or method with an appropriate name... ?
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 171 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8dNg30-GtKOWmCXv3:-ML8dNg45eeaX8gt5CRO:b-k1cv67) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L171)):*
> ```Smalltalk
> 		}
> 
> ```

Several of these tests fail on TC and Linux because `Path.GetInvalidFileNameChars()` / `Path.GetInvalidPathChars()` return different values on Linux. Might be easiest to create different versions of theses tests for Linux and Windows with a `Platform(Exclude...)` attribute. Or change the illegal characters used in the tests to only consist of the illegal characters that are the same on both platforms.

(When checking which chars are reported as invalid on Linux it would be good to check with our custom Mono 5 as well as the upstream Mono 6. It's possible that Mono 6 is closer to Windows, in which case we should patch our Mono 5. You might also check the other libpalaso classes - do we have a method somewhere that returns invalid filename/path chars?)
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 242 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8fPC4-egrAqZv2HYE:-ML8fOA95JE1U7c9eU8N:b-6dfs4t) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L242)):*
> ```
> https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming_conventions
> ```

The link could be simplified to https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions. Same below.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 313 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8i2WX5G_CBupXhoGv:-ML8i2WX5G_CBupXhoGw:bnbw32a) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L313)):*
> ```Smalltalk
> 			Assert.AreEqual("_", orig.SanitizeFilename('_'));
> 		}
> ```

We should have a test for `.foo` (which is valid according to the naming conventions)
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 316 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8hOlm-vn4zhfic9nF:-ML8hOlm-vn4zhfic9nG:b-908dxu) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L316)):*
> ```
> charcters
> ```
typo

tombogle:
___
*[SIL.Core/Extensions/StringExtensions.cs, line 224 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8bqIK8HxX3ti5oHzi:-ML9GHiHEVvdIoVSO_D4:b-5rlb5y) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core/Extensions/StringExtensions.cs#L224)):*

> ```
> 					(lastCharPos < 2 ||
> 					result[lastCharPos - 1] != '.' ||
> 					(result[lastCharPos - 2] != Path.DirectorySeparatorChar &&
> 					result[lastCharPos - 2] != Path.AltDirectorySeparatorChar)) &&
> 					!(lastCharPos == 1 && result[0] == '.'))
> ```
I have a hard time wrapping my brain around these nested expressions and understand what we test. I wonder if we could use `result.EndsWith("/.")` or something like that - maybe that would make it easier to understand. Or extract to a variable or method with an appropriate name... ?
</blockquote></details>

Yeah, I was wrestling over that. I almost broke out a method. That's probably the thing to do. I was/am a little worried because the original version made a point of saying that the lookup for invalid characters should be lightning fast, so it made me wonder if maybe this method is going to get called in some context where it's speed-critical. The nuit tests themselves do (sort of) document the required behavior, but it's definitely not the most readable code I've ever written.

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 171 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8dNg30-GtKOWmCXv3:-ML9IEAdBE7nKyvbylUD:b4ht9zd) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L171)):*

Several of these tests fail on TC and Linux because `Path.GetInvalidFileNameChars()` / `Path.GetInvalidPathChars()` return different values on Linux. Might be easiest to create different versions of theses tests for Linux and Windows with a `Platform(Exclude...)` attribute. Or change the illegal characters used in the tests to only consist of the illegal characters that are the same on both platforms.

(When checking which chars are reported as invalid on Linux it would be good to check with our custom Mono 5 as well as the upstream Mono 6. It's possible that Mono 6 is closer to Windows, in which case we should patch our Mono 5. You might also check the other libpalaso classes - do we have a method somewhere that returns invalid filename/path chars?)
</blockquote></details>

I'll try to use this strategy: https://stackoverflow.com/questions/866816/dynamically-create-tests-in-nunit

tombogle:
___
*[SIL.Core/Extensions/StringExtensions.cs, line 224 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8bqIK8HxX3ti5oHzi:-ML9PV8h2pUFAvVqTM2C:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core/Extensions/StringExtensions.cs#L224)):*

Yeah, I was wrestling over that. I almost broke out a method. That's probably the thing to do. I was/am a little worried because the original version made a point of saying that the lookup for invalid characters should be lightning fast, so it made me wonder if maybe this method is going to get called in some context where it's speed-critical. The nuit tests themselves do (sort of) document the required behavior, but it's definitely not the most readable code I've ever written.
</blockquote></details>

Done.

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 343 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuU787CQ12JxPxOYFA:-ML9UpAU1um9reo7-5xE:b-bl2mwf) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L343)):*

Sorry, I meant we should have `[TestCase(@".", ExpectedResult = @".")]`
</blockquote></details>

There is a test case for that (different test), but the expected result is _. That said, I'm ambivalent about that expectation. Technically a single . is legal, just not very likely to be useful. Now that I've refactored that code into a separate method, it would be pretty trivial to change it to keep a single dot.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @ermshiperete)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 242 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8fPC4-egrAqZv2HYE:-ML9U-8K5Tz0UgNLL174:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L242)):*

> ```
> https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming_conventions
> ```

The link could be simplified to https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions. Same below.
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 313 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8i2WX5G_CBupXhoGv:-ML9Uiny9RjIN6FZagN6:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L313)):*

We should have a test for `.foo` (which is valid according to the naming conventions)
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 316 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8hOlm-vn4zhfic9nF:-ML9Uji96ozut3iKrx2Y:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L316)):*

> ```
> charcters
> ```
typo
</blockquote></details>

Done.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @tombogle)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 343 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuU787CQ12JxPxOYFA:-ML9oo026tiIspUS_n6G:b-ut6wk3) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L343)):*

There is a test case for that (different test), but the expected result is _. That said, I'm ambivalent about that expectation. Technically a single . is legal, just not very likely to be useful. Now that I've refactored that code into a separate method, it would be pretty trivial to change it to keep a single dot.
</blockquote></details>

I'm pretty sure the old code would have kept "." as ".". And in this case, "." vs. "_" are two very different things, much differenter than "a" vs. "_". If there is some ambiguity in what the expected result should be, it seems like we need to maintain the prior behavior.

andrew-polk (changes requested):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: all files reviewed, 7 unresolved discussions (waiting on @ermshiperete and @tombogle)

___
*[SIL.Core/Extensions/StringExtensions.cs, line 233 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML9pdGO6zcBf6MyA7Md:-ML9pdGO6zcBf6MyA7Me:b5ikgmz) ([raw file](https://github.com/sillsdev/libpalaso/blob/0d5a49781ef1b8349e5e19cfb27d15127181444b/SIL.Core/Extensions/StringExtensions.cs#L233)):*
> ```Smalltalk
> 		}
> 
> 		private static bool TrailingDotIsValidHierarchySpecifier(StringBuilder result, int lastCharPos)
> ```

You're calling this with a char but using it as a position.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: all files reviewed, 7 unresolved discussions (waiting on @ermshiperete and @tombogle)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 343 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuU787CQ12JxPxOYFA:-ML9q2DMEvXqQnzKU80X:b-m6jebw) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L343)):*

I'm pretty sure the old code would have kept "." as ".". And in this case, "." vs. "_" are two very different things, much differenter than "a" vs. "_". If there is some ambiguity in what the expected result should be, it seems like we need to maintain the prior behavior.
</blockquote></details>

Thanks for the formatting, Reviewable...
Try again.
I'm pretty sure the old code would have kept "." as ".". And in this case, "." vs. "\_" are two very different things, much differenter than "a" vs. "\_". If there is some ambiguity in what the expected result should be, it seems like we need to maintain the prior behavior.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @tombogle)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 171 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8dNg30-GtKOWmCXv3:-MLAnXkh0WOhoyQxF4Mr:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L171)):*

I'll try to use this strategy: https://stackoverflow.com/questions/866816/dynamically-create-tests-in-nunit
</blockquote></details>

Done.

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 343 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MKuU787CQ12JxPxOYFA:-MLAnq0o-2FPGW6VPdVc:b-l534v1) ([raw file](https://github.com/sillsdev/libpalaso/blob/73f8df04ae7cfdb64d260f8069bfa397470e3bd9/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L343)):*

Thanks for the formatting, Reviewable...
Try again.
I'm pretty sure the old code would have kept "." as ".". And in this case, "." vs. "\_" are two very different things, much differenter than "a" vs. "\_". If there is some ambiguity in what the expected result should be, it seems like we need to maintain the prior behavior.

</blockquote></details>

I see. I need to have two different test cases. For SanitizeFilename, "." needs to become "_" because a single dot is not a useful/legal filename. But for SanitizePath, "." should stay "."

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @andrew-polk and @ermshiperete)

___
*[SIL.Core/Extensions/StringExtensions.cs, line 233 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML9pdGO6zcBf6MyA7Md:-MLAosRZDkKyxrNCsmQB:bfhayip) ([raw file](https://github.com/sillsdev/libpalaso/blob/0d5a49781ef1b8349e5e19cfb27d15127181444b/SIL.Core/Extensions/StringExtensions.cs#L233)):*

You're calling this with a char but using it as a position.
</blockquote></details>

Yeah, I changed horses midstream.

ermshiperete (changes requested):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: all files reviewed, 8 unresolved discussions (waiting on @andrew-polk and @tombogle)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 171 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/983#-ML8dNg30-GtKOWmCXv3:-MLC0pL31q9myfq0kLqQ:b4toox8) ([raw file](https://github.com/sillsdev/libpalaso/blob/9cbbf80560616b2054035f4ff94fc82cf696c418/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L171)):*

Done.
</blockquote></details>

Nice :+1:  :smile: 
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 174 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC0sjy4oWzhEhPMg7N:-MLC0sjz-fb4dqmX1Ei-:blc3ye2) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L174)):*
> ```Smalltalk
> 		}
> 
> 		[Test, TestCaseSource(nameof(GetInvalidFilenameCharacters))]
> ```

I don't think we need the `Test` attribute if we use `TestCaseSource` (https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html)
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 379 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC2BYM8ouk-e4BpYIz:-MLC2BYM8ouk-e4BpYJ-:b-hqaq95) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L379)):*
> ```Smalltalk
> 		[TestCase(@"%:\yeah", ExpectedResult = @"%_\yeah")]
> 		[TestCase(@"abc:", ExpectedResult = "abc_")]
> 		[TestCase(@"abc:", ExpectedResult = "abc_")]
> ```

Is this last line the same test case as the line above or do I overlook something?
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 380 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC3Knx4Ha6_mGrvnIs:-MLC3Kny17LmeChqfeKF:b9n9my4) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L380)):*
> ```Smalltalk
> 		[TestCase(@"abc:", ExpectedResult = "abc_")]
> 		[TestCase(@"abc:", ExpectedResult = "abc_")]
> 		[Platform(Include = "win", Reason = "Rules about colons and slashes only apply to Windows.")]
> ```

Can you add the Platform attribute above the TestCase attributes? IMO that would make it more visible.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 384 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC2RwQ3oTOSpDwBO0Y:-MLC2RwQ3oTOSpDwBO0Z:b-gir6pq) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L384)):*
> ```Smalltalk
> 		{
> 			return orig.SanitizePath('_');
> 		}
> ```

What about "c:" ? Might be good to add a test case for that for documentation purposes.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 401 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC37M-3EGEPb-j075q:-MLC37M-3EGEPb-j075r:b-q6sx1h) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L401)):*
> ```Smalltalk
> 		[TestCase(@".\..", ExpectedResult = @".\..")]
> 		[TestCase(@".\...", ExpectedResult = @".\")] // REVIEW: maybe we want: .\_
> 		public string SanitizePath_StringHasTrailingDots_TrailingDotsTrimmed(string orig)
> ```

Maybe add "..\.." as test case?

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 174 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC0sjy4oWzhEhPMg7N:-MLFYTpLDcutPCZF6YTQ:bnta8iz) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L174)):*

I don't think we need the `Test` attribute if we use `TestCaseSource` (https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html)
</blockquote></details>

You are correct, and I guess that makes sense. The example I was following had it.

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 379 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC2BYM8ouk-e4BpYIz:-MLFYjsR6uHZqX_m389X:b1uz2fi) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L379)):*

Is this last line the same test case as the line above or do I overlook something?
</blockquote></details>

Identical except for the line number :-)

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 384 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC2RwQ3oTOSpDwBO0Y:-MLFZg39D5vQcZ-IGf9T:blw3vjs) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L384)):*

What about "c:" ? Might be good to add a test case for that for documentation purposes.
</blockquote></details>

Added that test case to SanitizePath_StringContainsCharactersThatAreNotValidFileCharactersButAreValidPathCharacters_NoChange

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: 1 of 2 files reviewed, 8 unresolved discussions (waiting on @andrew-polk and @ermshiperete)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 380 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC3Knx4Ha6_mGrvnIs:-MLFZ9ng7FHuyPll5Bww:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L380)):*

Can you add the Platform attribute above the TestCase attributes? IMO that would make it more visible.
</blockquote></details>

Done.
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 401 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC37M-3EGEPb-j075q:-MLF_u4C58RtG5K-Xx1h:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L401)):*

Maybe add "..\.." as test case?
</blockquote></details>

Done.

ermshiperete (approved):
:lgtm:

> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983#-:-MLHcwcz5Y21kobPCvmM:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 401 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC37M-3EGEPb-j075q:-MLHccVn7ykyUcb27is7:b-5a7k5w) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L401)):*

Done.
</blockquote></details>

Oops, I meant "..\.." - where did that backslash got lost? Anyways, probably not that important...

andrew-polk (approved):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/983)* status: :shipit: complete! all files reviewed, all discussions resolved

tombogle:
___
*[SIL.Core.Tests/Extensions/StringExtensionTests.cs, line 401 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/983#-MLC37M-3EGEPb-j075q:-MLJiGJaEC3Bdw2wPecK:b-8g4ssg) ([raw file](https://github.com/sillsdev/libpalaso/blob/ed8f67520ddb9cb833e016ec31ab6f2ca5a97c90/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L401)):*

Oops, I meant "..\.." - where did that backslash got lost? Anyways, probably not that important...
</blockquote></details>

I think Reviewable (or github) is clobbering it because it's still not showing. I guessed you meant ..slash.. so I tried that and it works. I also tried \.. which is non-sensical but allowed by the OS, and it works too. Didn't bother checking them in for fear it would require another review.

PR #988, opened 2020-11-05
Replace MarkdownDeep with Markdig

ArmorBearer:
In order to replace MarkdownDeep, which doesn't seem to be maintained and hasn't had an update in a very long time, Todd Hoatson has revised the ShowReleaseNotesDialog form by using the signed version of Markdig.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/988)


ermshiperete:
This PR should go against `feature/nuget` - @ArmorBearer can you please change the target branch. Thanks!

PR #997, opened 2020-11-20
Update VerseRef.TryGetVerseNum() to handle non-latin numerals

SamDelaney:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/997)

### DIFF
--- a/SIL.Scripture.Tests/VerseRefTests.cs
+++ b/SIL.Scripture.Tests/VerseRefTests.cs
@@ -2067,6 +2067,38 @@ namespace SIL.Scripture.Tests
 			Assert.AreEqual(new VerseRef("EXO 6:4"), new VerseRef("EXO 6:4-10").UnBridge());
 			Assert.AreEqual(new VerseRef("EXO 6:150monkeys"), new VerseRef("EXO 6:150monkeys").UnBridge());
 		}
+
+		/// <summary>
+		/// Tests the Verse property's set method with numerals from various Unicode-supported scripts
+		/// </summary>
+		[Test]
+		public void SetVerse_InterpretUnicodeNumerals()
+		{
+			VerseRef vref = new VerseRef("EXO 6:5");
+			Assert.AreEqual(5, vref.VerseNum);
+
+			vref.Verse = "५"; // Devanagari numeral
+			Assert.AreEqual(5, vref.VerseNum);
+
+			vref.Verse = "૧૬"; // Gujarati numeral
+			Assert.AreEqual(16, vref.VerseNum);
+
+			vref.Verse = "൬"; // Malayalam numeral
+			Assert.AreEqual(6, vref.VerseNum);
+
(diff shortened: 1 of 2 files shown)


jwickberg (approved):

SIL.Scripture.Tests/VerseRefTests.cs
```diff
@@ -2067,6 +2067,38 @@ public void UnBridge()
...
 			Assert.AreEqual(new VerseRef("EXO 6:150monkeys"), new VerseRef("EXO 6:150monkeys").UnBridge());
 		}
+
+		/// <summary>
+		/// Tests the Verse property's set method with numerals from various Unicode-supported scripts
+		/// </summary>
+		[Test]
+		public void SetVerse_InterpretUnicodeNumerals()
+		{
+			VerseRef vref = new VerseRef("EXO 6:5");
```
ermshiperete, line 2077: IMO it would be better to write this as multiple tests with one test case per test, or use the TestCase attribute. Something like:

```
[TestCase("५", TestName="Devanagari numeral")]
[TestCase("૧૬", TestName="Gujarati numeral")]
...
public void SetVerse_InterpretUnicodeNumerals(string verseNumber)
{
var vref = new VerseRef("EXO 6:5");
vref.Verse = verseNumber;
Assert.That(vref.VerseNum, Is.EqualTo(5));
}
```

SamDelaney: I agree. Done.

SIL.Scripture.Tests/VerseRefTests.cs
```diff
@@ -2067,6 +2067,24 @@ public void UnBridge()
...
+		[TestCase("᠔", 4, TestName = "Mongolian numeral")]
+		[TestCase("A", -1, TestName = "Latin non-numeral")]
+		[TestCase("ะ", -1, TestName = "Thai non-numeral")]
+		public void SetVerse_InterpretUnicodeNumerals(string verseStr, int parsedVerse)
+		{
+			VerseRef vref = new VerseRef("EXO 6:1");
+
+			vref.Verse = verseStr;
+			Assert.AreEqual(parsedVerse, vref.VerseNum);
+		}
```
ermshiperete, line 2086: :+1: 

Ahh, sorry, didn't notice that we test with different values.

This is fine as it is, but if you want you could use the ExpectedResult parameter:

```
[TestCase("ะ", ExpectedResult = -1, TestName = "Thai non-numeral")]
public int SetVerse_InterpretUnicodeNumerals(string verseStr)
{
	VerseRef vref = new VerseRef("EXO 6:1");

	vref.Verse = verseStr;
	return vref.VerseNum;
}
```

SamDelaney: I do like that slightly better. Thanks.

ermshiperete (approved):
lgtm

PR #998, opened 2020-11-22
Handle transient languages with invalid .NET data

mcdurdin:
In some circumstances, it appears that `inputLanguage.culture.KeyboardLayoutId` property has a mismatch with the `inputLanguage.Handle` property, and is returning the wrong value (e.g. `0x04090409` instead of `0x04092000`), and we should instead be enumerating on the basis of the input value from `inputLanguage`.

This was spotted in testing by June, who found that Keyman keyboards associated with transient languages were not listed as available for use in PT.

About transient languages: https://docs.microsoft.com/en-us/globalization/locale/locale-names

Using the `inputLanguage.Handle` property removes this lookup and should return the same results.

It would be good to know if this is a bug in .NET or something else that's environmental. I have tested on Win 10 1909 and Win 10 2004 and the values are correct in both cases on my test machines (in either case), and this was only going wrong on June's machine.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/998)

PR #999, opened 2020-11-24
Separate non-Roman verse setting to restore USX compliance in VerseRef

SamDelaney:
Done in response to issues with PR #997

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/999)

### DIFF
--- a/SIL.Scripture.Tests/VerseRefTests.cs
+++ b/SIL.Scripture.Tests/VerseRefTests.cs
@@ -2067,6 +2067,38 @@ namespace SIL.Scripture.Tests
 			Assert.AreEqual(new VerseRef("EXO 6:4"), new VerseRef("EXO 6:4-10").UnBridge());
 			Assert.AreEqual(new VerseRef("EXO 6:150monkeys"), new VerseRef("EXO 6:150monkeys").UnBridge());
 		}
+
+		/// <summary>
+		/// Tests the Verse property's set method with numerals from various Unicode-supported scripts
+		/// </summary>
+		[Test]
+		public void SetVerse_InterpretUnicodeNumerals()
+		{
+			VerseRef vref = new VerseRef("EXO 6:5");
+			Assert.AreEqual(5, vref.VerseNum);
+
+			vref.Verse = "५"; // Devanagari numeral
+			Assert.AreEqual(5, vref.VerseNum);
+
+			vref.Verse = "૧૬"; // Gujarati numeral
+			Assert.AreEqual(16, vref.VerseNum);
+
+			vref.Verse = "൬"; // Malayalam numeral
+			Assert.AreEqual(6, vref.VerseNum);
+
+			vref.Verse = "᠔"; // Mongolian numeral
+			Assert.AreEqual(4, vref.VerseNum);
+
+			vref.Verse = "๙"; // Thai numeral
+			Assert.AreEqual(9, vref.VerseNum);
+
+			vref.Verse = "A"; // Latin non-numeral
+			Assert.AreEqual(-1, vref.VerseNum);
+
+			vref.Verse = "ะ"; // Thai non-numeral
+			Assert.AreEqual(-1, vref.VerseNum);
+		}
+
 		#endregion
 
 		#region InRange
--- a/SIL.Scripture.Tests/VerseRefTests.cs
+++ b/SIL.Scripture.Tests/VerseRefTests.cs
@@ -2077,11 +2077,11 @@ namespace SIL.Scripture.Tests
 		[TestCase("᠔", ExpectedResult = 4, TestName = "Mongolian numeral")]
 		[TestCase("A", ExpectedResult = -1, TestName = "Latin non-numeral")]
 		[TestCase("ะ", ExpectedResult = -1, TestName = "Thai non-numeral")]
-		public int SetVerse_InterpretUnicodeNumerals(string verseStr)
+		public int SetVerseUnicode_InterpretNumerals(string verseStr)
 		{
 			VerseRef vref = new VerseRef("EXO 6:1");
 
-			vref.Verse = verseStr;
+			vref.SetVerseUnicode(verseStr);
 			return vref.VerseNum;
 		}
 
--- a/SIL.Scripture.Tests/VerseRefTests.cs
+++ b/SIL.Scripture.Tests/VerseRefTests.cs
@@ -2071,18 +2071,18 @@ namespace SIL.Scripture.Tests
 		/// <summary>
 		/// Tests the Verse property's set method with numerals from various Unicode-supported scripts
 		/// </summary>
-		[TestCase("५", 5, TestName = "Devanagari numeral")]
-		[TestCase("૧૬", 16, TestName = "Gujarati numeral")]
-		[TestCase("5",5, TestName = "Latin numeral")]
-		[TestCase("᠔", 4, TestName = "Mongolian numeral")]
-		[TestCase("A", -1, TestName = "Latin non-numeral")]
-		[TestCase("ะ", -1, TestName = "Thai non-numeral")]
-		public void SetVerse_InterpretUnicodeNumerals(string verseStr, int parsedVerse)
+		[TestCase("५", ExpectedResult = 5, TestName = "Devanagari numeral")]
+		[TestCase("૧૬", ExpectedResult = 16, TestName = "Gujarati numeral")]
+		[TestCase("5", ExpectedResult = 5, TestName = "Latin numeral")]
+		[TestCase("᠔", ExpectedResult = 4, TestName = "Mongolian numeral")]
+		[TestCase("A", ExpectedResult = -1, TestName = "Latin non-numeral")]
+		[TestCase("ะ", ExpectedResult = -1, TestName = "Thai non-numeral")]
+		public int SetVerse_InterpretUnicodeNumerals(string verseStr)
 		{
 			VerseRef vref = new VerseRef("EXO 6:1");
 
 			vref.Verse = verseStr;
-			Assert.AreEqual(parsedVerse, vref.VerseNum);
+			return vref.VerseNum;
 		}
 
 		#endregion
--- a/SIL.Scripture.Tests/VerseRefTests.cs
+++ b/SIL.Scripture.Tests/VerseRefTests.cs
@@ -2069,7 +2069,7 @@ namespace SIL.Scripture.Tests
 		}
 
 		/// <summary>
-		/// Tests the Verse property's set method with numerals from various Unicode-supported scripts
+		/// Tests the TrySetVerseUnicode method with numerals from various Unicode-supported scripts
 		/// </summary>
 		[TestCase("५", ExpectedResult = 5, TestName = "Devanagari numeral")]
 		[TestCase("૧૬", ExpectedResult = 16, TestName = "Gujarati numeral")]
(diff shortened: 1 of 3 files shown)


SIL.Scripture.Tests/VerseRefTests.cs
```diff
@@ -2077,11 +2077,11 @@ public void UnBridge()
...
 		[TestCase("ะ", ExpectedResult = -1, TestName = "Thai non-numeral")]
-		public int SetVerse_InterpretUnicodeNumerals(string verseStr)
+		public int SetVerseUnicode_InterpretNumerals(string verseStr)
 		{
 			VerseRef vref = new VerseRef("EXO 6:1");
 
-			vref.Verse = verseStr;
+			vref.SetVerseUnicode(verseStr);
 			return vref.VerseNum;
 		}
```
ermshiperete, line 2086: We should add a test for `vRef.Verse = verseStr`

SIL.Scripture/VerseRef.cs
```diff
@@ -302,14 +302,14 @@ static bool TryGetVerseNum(string verseStr, out short vNum)
 			for (int i = 0; i < verseStr.Length; i++)
 			{
 				char ch = verseStr[i];
-				if (!char.IsDigit(ch))
+				if (romanOnly ? (ch < '0' || ch > '9') : !char.IsDigit(ch))
```
ermshiperete, line 305: Looking at the documentation for [`char.IsDigit/char.IsNumber`](https://docs.microsoft.com/en-us/dotnet/api/system.char.isnumber?view=net-5.0#System_Char_IsNumber_System_Char_) it seems our method here won't work for surrogate pairs (e.g. Chakma digits), and it won't work for numbers that don't have the DecimalDigitNumber category (e.g. Aegean - probably not very relevant for us, but I'm sure there are other similar numbering systems). Cf. <http://unicode.org/pending/properties.html>, [UnicodeData](http://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt)

I'm not sure we have to fix this now, but it would be good to document it by creating a GitHub issue and possibly (ignored) failing tests cases.
If you don't want to

ermshiperete (changes requested):
Since this adds a new method the commit message should have a line:

`+semver:minor`

Also please add a line to `CHANGELOG.md` with the new method.

SIL.Scripture/VerseRef.cs
SamDelaney: @ermshiperete Writing some tests for this, but I'm having trouble finding an example of Chakma (or any other script) using a surrogate pair numeral to test against. Could you provide one?

ermshiperete: I'd think "\U0001113A\U00011138" would be one (42; CHAKMA DIGIT FOUR; CHAKMA DIGIT TWO)

SIL.Scripture.Tests/VerseRefTests.cs
SamDelaney: Done

SIL.Scripture/VerseRef.cs
```diff
@@ -1338,6 +1338,27 @@ static void ParseVerseNumber(string vNum, out int number, out string segment)
 			segment = vNum.Substring(j);
 		}
 
+		/// <summary>
+		/// Parses a verse string and gets the leading numeric portion as a number.
+		/// Functionally identical to Verse.Set for Roman numbers, made distinct to preserve USX standard.
+		/// </summary>
+		/// <returns><c>true</c> if the entire string could be parsed as a single,
+		/// simple verse number in any supported script; <c>false</c> if the verse string represented
+		/// a verse bridge, contained segment letters, or was invalid</returns>
```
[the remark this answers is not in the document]
SamDelaney: Implemented the comment. Thanks.

SIL.Scripture/VerseRef.cs
```diff
@@ -1338,6 +1338,27 @@ static void ParseVerseNumber(string vNum, out int number, out string segment)
...
+		public void SetVerseUnicode(string value)
+		{
+			{
+				short vNum;
+				verse = !TryGetVerseNum(value, out vNum, false) ? value.Replace(rtlMark, "") : null;
```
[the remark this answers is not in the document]
SamDelaney: Changed here and in Verse.Set for consistency. Thanks.

SIL.Scripture/VerseRef.cs
SamDelaney: Thanks. Wrote the tests and created #1000.

SIL.Scripture/VerseRef.cs
```diff
@@ -1338,6 +1337,27 @@ static void ParseVerseNumber(string vNum, out int number, out string segment)
...
+		/// Parses a verse string and gets the leading numeric portion as a number.
+		/// Functionally identical to Verse.Set for Roman numbers, made distinct to preserve USX standard.
+		/// </summary>
+		/// <returns><c>true</c> if the entire string could be parsed as a single,
+		/// simple verse number in any supported script; <c>false</c> if the verse string represented
+		/// a verse bridge, contained segment letters, or was invalid</returns>
+		public bool TrySetVerseUnicode(string value)
+		{
+			{
+				verse = !TryGetVerseNum(value, out short vNum, false) ? value.Replace(rtlMark, "") : null;
```
ermshiperete, line 1350: nit: why don't you directly use `out short verseNum`? Then the assignment in the next line wouldn't be necessary... (sorry, I overlooked that the first time...)

SIL.Scripture/VerseRef.cs
```diff
@@ -1338,6 +1337,27 @@ static void ParseVerseNumber(string vNum, out int number, out string segment)
...
+		/// <summary>
+		/// Parses a verse string and gets the leading numeric portion as a number.
+		/// Functionally identical to Verse.Set for Roman numbers, made distinct to preserve USX standard.
+		/// </summary>
+		/// <returns><c>true</c> if the entire string could be parsed as a single,
+		/// simple verse number in any supported script; <c>false</c> if the verse string represented
+		/// a verse bridge, contained segment letters, or was invalid</returns>
+		public bool TrySetVerseUnicode(string value)
+		{
+			{
```
ermshiperete, line 1349: Since this is not a property we don't need the extra block

SIL.Scripture/VerseRef.cs
```diff
@@ -1338,6 +1337,27 @@ static void ParseVerseNumber(string vNum, out int number, out string segment)
...
+				verse = !TryGetVerseNum(value, out short vNum, false) ? value.Replace(rtlMark, "") : null;
+				verseNum = vNum;
+				if (verseNum >= 0)
+					return true;
+
+				Trace.TraceWarning("Just failed to parse a verse number: " + value);
+				TryGetVerseNum(verse, out verseNum, false);
+				return false;
+			}
+		}
```
ermshiperete, line 1359: This method is almost identical to the code in `Verse.set`. We should refactor this to a private method `private bool SetVerse(string value, bool romanOnly)` to remove the code duplication.

SIL.Scripture/VerseRef.cs
```diff
@@ -1338,6 +1337,27 @@ static void ParseVerseNumber(string vNum, out int number, out string segment)
 			segment = vNum.Substring(j);
 		}
 
+		/// <summary>
+		/// Parses a verse string and gets the leading numeric portion as a number.
```
ermshiperete, line 1341: We should have test cases for "1-", "2.3", "5-7", and "6a"

SIL.Scripture.Tests/VerseRefTests.cs
```diff
@@ -2069,19 +2069,40 @@ public void UnBridge()
...
+		/// Tests the TrySetVerseUnicode method with numerals from various Unicode-supported scripts
 		/// </summary>
 		[TestCase("५", ExpectedResult = 5, TestName = "Devanagari numeral")]
 		[TestCase("૧૬", ExpectedResult = 16, TestName = "Gujarati numeral")]
 		[TestCase("5", ExpectedResult = 5, TestName = "Latin numeral")]
 		[TestCase("᠔", ExpectedResult = 4, TestName = "Mongolian numeral")]
 		[TestCase("A", ExpectedResult = -1, TestName = "Latin non-numeral")]
 		[TestCase("ะ", ExpectedResult = -1, TestName = "Thai non-numeral")]
-		public int SetVerse_InterpretUnicodeNumerals(string verseStr)
+		[TestCase("二十", ExpectedResult = 20, TestName = "Japanese numeral", IgnoreReason = "Non-decimal numeral systems not yet implemented.")]
```
ermshiperete, line 2080: You could mention #1000 in the ignore reason

ermshiperete (changes requested):
Thanks for adding the ignored tests!

I noticed a few more things...

SIL.Scripture/VerseRef.cs
SamDelaney: Updated to use `out verseNum`. Using `short` would have declared a new variable, which would have conflicted with the member variable.

SIL.Scripture/VerseRef.cs
SamDelaney: Good point. Changed.

SIL.Scripture/VerseRef.cs
SamDelaney: Done

ermshiperete (approved):
:lgtm: now :smile: 

> Reviewed 1 of 3 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/999#-:-MNSenPmEdhDsmsJwaR3:bgx7kq5)* status: all files reviewed, 1 unresolved discussion (waiting on @SamDelaney)

___
*[SIL.Scripture/VerseRef.cs, line 282 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/999#-MNSeGf-3jTjArWYOFDz:-MNSeGf-3jTjArWYOFE-:by6r5vy) ([raw file](https://github.com/sillsdev/libpalaso/blob/ce21b22f53b9cc4bb260fe75819eed4805aa53d2/SIL.Scripture/VerseRef.cs#L282)):*
> ```Smalltalk
> 		/// </summary>
> 		/// <returns><c>true</c> if the verse was set successfully </returns>
> 		bool TrySetVerse(string value, bool romanOnly = true)
> ```

nit: for a new method I think it would be clearer not to use a default value, especially since it's used only twice. But that might be a matter of personal preference...

SamDelaney (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/999)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @ermshiperete)

___
*[SIL.Scripture/VerseRef.cs, line 282 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/999#-MNSeGf-3jTjArWYOFDz:-MNUJ2dy04WJcjCCcS-o:b-8l85mp) ([raw file](https://github.com/sillsdev/libpalaso/blob/ce21b22f53b9cc4bb260fe75819eed4805aa53d2/SIL.Scripture/VerseRef.cs#L282)):*

nit: for a new method I think it would be clearer not to use a default value, especially since it's used only twice. But that might be a matter of personal preference...
</blockquote></details>

My thought process was that the default value helps indicate the preference of the USX standard for Latin numerals. I removed the default value and instead left a comment to that effect.

ermshiperete (approved):
:lgtm:

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/999#-:-MNUYPAm4w4sZ7NofT8V:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

FoolRunning:
@ermshiperete, It looks like this is waiting for checks to pass, but they look like they passed in TC. Are you able to override that and merge this PR? Could you then do a merge from master to the NuGet branch?

ermshiperete:
Retriggering the builds on TC did the trick. Unfortunately I won't be able to do a merge from `master` into `feature/nuget` until next week.

FoolRunning:
@ermshiperete, The merge from master to feature/nuget is now holding up the Paratext 9.1 release (we're waiting for the changes in #998). Are you deliberately holding off on doing the merge for some reason or is it just having the time to do it? If it's just time, I can find someone else that can do it.

ermshiperete:
@FoolRunning I just don't have time this week, so if somebody else can do it - great!

FoolRunning:
@ermshiperete, Tom Bogle created a PR and we did it that way. Thanks for making it so easy to create the NuGet packages. :)

PR #1024, opened 2020-12-07
Change http to https in comments and UI strings

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1024)


tombogle:
Started with wrong base, and Appveyor can't see that I( fixed it, so it it doesn't kick off the required build.

PR #1054, opened 2021-01-22
Fix AcknowledgementsProvider

ermshiperete:
For some reason (probably newer mono version) we get a different exception message when the AcknowledgmentsProvider can't load a file. I can't see a reason why we'd have to check the exception message, so now we always ignore a BadImageException.

This change also fixes the SWF test app so that it properly works with Gecko on Linux.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1054)


jasonleenaylor (approved):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1054)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1060, opened 2021-02-04
SP-2193: Added handler for DataError in ContributorsListControl grid

tombogle:
This prevents a null Role from causing an error to be displayed when formatting that cell in a new row.
Added to the TestApp so it would be easy to see how ContributorsListControl is to be used and to enable debugging it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1060)

### DIFF
--- a/TestApps/SIL.Windows.Forms.TestApp/ContributorsForm.cs
+++ b/TestApps/SIL.Windows.Forms.TestApp/ContributorsForm.cs
@@ -161,13 +161,15 @@ namespace SIL.Windows.Forms.TestApp
 		/// ------------------------------------------------------------------------------------
 		private void UpdateNames(object sender, EventArgs e)
 		{
+			// The following illustrates how to avoid an InvalidOperationException when the
+			// grid is not in a state where it passes validation.
+			if (!_contributorsControl.Validate())
+				return;
+
 			var contribs = _model.Contributions;
-			for (var i = 0; i < _contributorNames.RowCount; i++)
+			for (var i = 0; i < _contributorNames.RowCount && i < contribs.Count; i++)
 			{
-				if (i < contribs.Count)
-				{
-					contribs[i].ContributorName = (string)_contributorNames.Rows[i].Cells[0].Value;
-				}
+				contribs[i].ContributorName = (string)_contributorNames.Rows[i].Cells[0].Value;
 			}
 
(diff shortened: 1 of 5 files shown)


TestApps/SIL.Windows.Forms.TestApp/ContributorsForm.cs
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			return new KeyValuePair<string, string>();
+		}
+
+		/// ------------------------------------------------------------------------------------
+		private void UpdateNames(object sender, EventArgs e)
+		{
+			var contribs = _model.Contributions;
+			for (var i = 0; i < _contributorNames.RowCount; i++)
```
papeh, line 165: i < _contributorNames.RowCount && i < contribs.Count

papeh (approved):

TestApps/SIL.Windows.Forms.TestApp/ContributorsForm.cs
```diff
@@ -0,0 +1,176 @@
...
+
+			public IEnumerable<string> GetValuesForKey(string key)
+			{
+				if (key == "person")
+				{
+					foreach (DataGridViewRow row in Source.Rows)
+					{
+						yield return row.Cells[0].Value as string;
+					}
+				}
```
ermshiperete, line 31: Don't we have to return something if `key != "person"` ?

ermshiperete (approved):

TestApps/SIL.Windows.Forms.TestApp/ContributorsForm.cs
tombogle: Because it's an IEnumerable, not returning anything (or not hitting the yield return) is okay. That will just result in an empty enumeration. In this particular instance, if you look at the caller, the key will currently always be "person," but I put that check into the implementation because if it ever changes, it would presumably not make sense to return the same values.

papeh (approved):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1060)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @hatton)

PR #1071, opened 2021-04-15
Save non-CC licenses properly in images

StephenMcConnel:
The change allows users to go from a CC license to a non-CC license for
images.  The reverse was already possible.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1071)

### DIFF
--- a/SIL.Windows.Forms.Tests/ClearShare/MetadataTests.cs
+++ b/SIL.Windows.Forms.Tests/ClearShare/MetadataTests.cs
@@ -6,6 +6,7 @@ using NUnit.Framework;
 using SIL.IO;
 using SIL.TestUtilities;
 using SIL.Windows.Forms.ClearShare;
+using TagLib.Xmp;
 
 namespace SIL.Windows.Forms.Tests.ClearShare
 {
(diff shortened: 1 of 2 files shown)


hatton:
The problem here is that the original code didn't explain what the thinking was in a comment, so I'm finding it difficult to know that this doesn't fix the case at hand but break some other case?

ermshiperete:
Maybe it would it be possible to add a unit test?

StephenMcConnel (commented):
The code was originally written by @johnthomson  in commit 8570f8837 on July 8, 2014.  The commit title was "Replace exiftool with taglib#".  There was more detail following, but nothing specific to this function or these lines in the code.
A unit test has been added.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1071#-:-MYQS8ABD0QJ_-XN7o4_:bs3ukwo)* status: 0 of 2 files reviewed, all discussions resolved

ermshiperete (approved):

hatton:
We have an unrelated test failing sometimes. I've created a card to fix it [here](https://issues.bloomlibrary.org/youtrack/issue/BL-9825), and will merge this.

PR #1081, opened 2021-06-17
Prevent second exception when handling parser exception (PTX-15835)

jwickberg:
We have had exception reports when Paratext is saving language data that has passed the Paratext checks, but causes an error in the SimpleRulesParser. This change prevents a second exception from being created because the exception from the Spart parser must not contain the expected data.

Hopefully with this change, we will get a better message and be able to figure out what the problem data is.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1081)

### DIFF
--- a/SIL.WritingSystems/SimpleRulesParser.cs
+++ b/SIL.WritingSystems/SimpleRulesParser.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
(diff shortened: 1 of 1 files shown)


FoolRunning (approved):
:lgtm: 

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1081#-:-McU56vvEmJibg1zdVPy:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jwickberg)

Issue #797, opened 2019-03-01
SelectProjectDlgBase should not access abstract properties in the constructor.


tombogle:
There's no actual need to initialize m_fileDialog (etc.) until we actually get ready to show the dialog for the first time. By accessing abstract properties in the constructor, it prevents Subclasses from having arguments passed in to their constructors that might be needed to implement the abstract members. In this case, there are times when subclasses might need to have a more meaningful Title based on something that is not static.

ermshiperete:
Suggestion: new method `protected virtual OpenFileDialog CreateFileDialog()` that gets called at the beginning of `ShowDialog`.

PR #1085, opened 2021-07-27
add protected method CreateFileDialog for abstract properties

elisunger:
Resolves: #797

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1085)


SIL.Windows.Forms.DblBundle/SelectProjectDlgBase.cs
```diff
@@ -18,6 +18,11 @@ public abstract class SelectProjectDlgBase : IDisposable
 		protected SelectProjectDlgBase(bool allowProjectFiles = true, string defaultFile = null)
 		{
 			FileName = File.Exists(defaultFile) ? Path.GetFileName(defaultFile) : null;
+			
+		}
+
+		protected virtual OpenFileDialog CreateFileDialog()
+		{
```
ermshiperete, line 25: This fails to compile because `defaultFile` is unknown. We could store `defaultFile` (or `defaultDir`) in a private field.

ermshiperete (changes requested):

ermshiperete (changes requested):
Almost... Right now `m_allowProjectFiles` will always have the default value (`false`). We need to assign `allowProjectFiles` to `m_allowProjectFiles` in the c'tor.

SIL.Windows.Forms.DblBundle/SelectProjectDlgBase.cs
```diff
@@ -8,30 +8,37 @@ namespace SIL.Windows.Forms.DblBundle
...
+		private string m_defaultDir;
+		private bool m_allowProjectFiles;
 
 		protected abstract string DefaultBundleDirectory { get; set; }
 		protected abstract string ProjectFileExtension { get; }
 		protected abstract string Title { get; }
 		protected abstract string ProductName { get; }
 
-		protected SelectProjectDlgBase(bool allowProjectFiles = true, string defaultFile = null)
+		protected SelectProjectDlgBase(bool allowProjectFiles, string defaultFile = null)
```
ermshiperete, line 20: We shouldn't change the signature of the c'tor!

```suggestion
		protected SelectProjectDlgBase(bool allowProjectFiles = true, string defaultFile = null)
```

ermshiperete (changes requested):
Please add a line to the `Added` section under `Unreleased` of `CHANGELOG.md`:

```
- [SIL.Windows.Forms.DblBundle] virtual method `SelectProjectDlgBase.CreateFileDialog()` to allow
  customization in derived class (#797)
```

And one comment in the c'tor. Otherwise :lgtm:

ermshiperete (approved):

PR #1096, opened 2021-09-16
Added extension method overloads

tombogle:
IEnumerable<T>.FirstOrDefault that takes a different default to use
string.IndexOf that takes an IComparer<string>

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1096)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms] `CssLinkHref` property to `ShowReleaseNotesDialog` to allow linking to CSS file for
     displaying Markdown output.
 - [SIL.Scripture] IScrVerseRef interface (largely extracted from VerseRef)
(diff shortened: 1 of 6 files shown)


tombogle:
Looking at my unit tests now I think they are "unfair" tests. If "th" should be interpreted as a diphthong for the purpose of comparing strings, then is "his" a substring of "this"? If you first convert the whole string, then clearly it is not. But I believe there is no way you could hope to write an implementation of IndexOf that could know that, based only on what IComparer<string> can tell it. Basically I think the problem presents a logical contradiction.
I just read something online that corrected my misunderstanding of the equivalence between ss and ß. Although different countries have different conventions, assuming a consistent standard orthography is applied, it might never be necessary to treat them as equal, at least for the purpose of finding a whole-word substring (though a sort order could do so with presumably no ill effects). So I think I'm probably back to the place of not needing this more "advanced" IndexOf functionality.

tombogle:
At best, I think you'd have to implement this to say that if an ordinal IndexOf found the substring, that match would always win. If not then maybe you could try the custom comparer to see if you could get a less exact match.

PR #1098, opened 2021-09-27
Fixed stack overflow caused by recursive definition of AreEqual

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1098)


FoolRunning:
I approved the changes and then realized that it would be best if there were some tests for that method/operator. I can't figure out how to rescind my approval.

PR #1103, opened 2021-10-04
Ensure settings use CrossPlatformSettingsProvider

papeh:


ermshiperete:
@papeh Do we still need the branch?

papeh:
Implementing this would be nice, but it's complicated, and there will always be more exciting things to do.

PR #1106, opened 2021-10-06
Enhance PortableClipboard and remove dependency on gtk-sharp

ermshiperete:
- extract Windows and Linux clipboarding code into separate files
- implement Linux clipboarding methods with unmanaged gtk
  functionality instead of gtk-sharp
- add GraphicsManager to allow selecting desired GTK version.
  Default: GTK2
- dynamically load unmanaged gtk libraries
- remove gtk-sharp/gdk-sharp dependencies

Fixes LT-20773.

Note that SIL.Windows.Forms still has a nuget dependency on DialogAdapters.Gtk2. However, that should be OK since the assembly name is simply `DialogAdapters`, and so the target project can simply reference DialogAdapters.Gtk3 instead and it should work with GTK3.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1106)

### DIFF
--- a/SIL.Windows.Forms/Clipboarding/WindowsClipboard.cs
+++ b/SIL.Windows.Forms/Clipboarding/WindowsClipboard.cs
@@ -15,40 +15,13 @@ namespace SIL.Windows.Forms.Clipboarding
 {
 	internal class WindowsClipboard: IClipboard
 	{
-		public bool ContainsText()
-		{
-			return Clipboard.ContainsText();
-		}
-
-		public string GetText()
-		{
-			return Clipboard.GetText();
-		}
-
-		public string GetText(TextDataFormat format)
-		{
-			return Clipboard.GetText(format);
-		}
-
-		public void SetText(string text)
-		{
-			Clipboard.SetText(text);
-		}
-
-		public void SetText(string text, TextDataFormat format)
-		{
-			Clipboard.SetText(text, format);
-		}
-
-		public bool ContainsImage()
-		{
-			return Clipboard.ContainsImage();
-		}
-
-		public Image GetImage()
-		{
-			return Clipboard.GetImage();
-		}
+		public bool ContainsText() => Clipboard.ContainsText();
+		public string GetText() => Clipboard.GetText();
+		public string GetText(TextDataFormat format) => Clipboard.GetText(format);
+		public void SetText(string text) => Clipboard.SetText(text);
+		public void SetText(string text, TextDataFormat format) => Clipboard.SetText(text, format);
+		public bool ContainsImage() => Clipboard.ContainsImage();
+		public Image GetImage() => Clipboard.GetImage();
 
 		public void CopyImageToClipboard(PalasoImage image)
 		{
@@ -81,7 +54,8 @@ namespace SIL.Windows.Forms.Clipboarding
 		// Try to get an image from the System.Windows.Forms.Clipboard. If there simply isn't anything on the System.Windows.Forms.Clipboard that
 		// can reasonably be interpreted as an image, return null. If there is something that makes sense
 		// as an image, but trying to load it causes an exception, let the exception propagate.
-		public PalasoImage GetImageFromClipboard() {
+		public PalasoImage GetImageFromClipboard()
+		{
 			// N.B.: PalasoImage does not handle .svg files
 			var dataObject = Clipboard.GetDataObject();
 			if (dataObject == null)
@@ -103,7 +77,7 @@ namespace SIL.Windows.Forms.Clipboarding
 					if (haveFileUrl)
 					{
 						var imageWithPathAndMaybeMetadata = PalasoImage.FromFileRobustly(textData);
-						plainImage.Dispose();//important: don't do this until we've successfully created the imageWithPathAndMaybeMetadata
+						plainImage.Dispose(); // important: don't do this until we've successfully created the imageWithPathAndMaybeMetadata
 						return imageWithPathAndMaybeMetadata;
 					}
 
@@ -134,23 +108,24 @@ namespace SIL.Windows.Forms.Clipboarding
 				}
 			}
 
-			//People can do a "copy" from the Windows Photo Viewer but what it puts on the System.Windows.Forms.Clipboard is a path, not an image
+			// People can do a "copy" from the Windows Photo Viewer but what it puts on the System.Windows.Forms.Clipboard is a path, not an image
 			if (dataObject.GetDataPresent(DataFormats.FileDrop))
 			{
-				//This line gets all the file paths that were selected in explorer
+				// This line gets all the file paths that were selected in explorer
 				string[] files = dataObject.GetData(DataFormats.FileDrop) as string[];
-				if (files == null) return null;
+				if (files == null)
+					return null;
 
 				foreach (var file in files.Where(f => RobustFile.Exists(f)))
 				{
-						return PalasoImage.FromFileRobustly(file);
+					return PalasoImage.FromFileRobustly(file);
 				}
 
-				return null; //not an image
+				return null; // not an image
 			}
 
 			if (Clipboard.ContainsText() && RobustFile.Exists(Clipboard.GetText()))
-				return PalasoImage.FromImage( Image.FromStream(new FileStream(Clipboard.GetText(), FileMode.Open)));
+				return PalasoImage.FromImage(Image.FromStream(new FileStream(Clipboard.GetText(), FileMode.Open)));
 
 			if (ex != null)
 			{
--- a/SIL.Windows.Forms/Miscellaneous/GraphicsManager.cs
+++ b/SIL.Windows.Forms/Miscellaneous/GraphicsManager.cs
@@ -17,11 +17,17 @@ namespace SIL.Windows.Forms.Miscellaneous
 		public const GtkVersion GTK2 = GtkVersion.Gtk2;
 		public const GtkVersion GTK3 = GtkVersion.Gtk3;
 
-		// NOTE: just selecting GTK3 usually isn't enough to make things work. System.Windows.Forms
-		// still uses GTK2 and runs a GTK2 message loop. To do anything with GTK3 (e.g.
-		// copy to/from the clipboard) you'll have to make sure that a GTK3 message loop is
-		// running.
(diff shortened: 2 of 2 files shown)


StephenMcConnel (commented):
This fixes the clipboard issues in Bloom if the one line is added to Bloom to set the GtkVersionInUse to Gtk3.
I have one comment that doesn't necessarily need to be addressed if it would hold up merging this.

> Reviewed 5 of 18 files at r1, 5 of 6 files at r2, 1 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-:-MlLAXcE9bUg0E7pOCw1:b-8of1a7)* status: 20 of 24 files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @jasonleenaylor)

___
*[TestApps/ClipboardTestApp/Program.cs, line 17 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLCJ4Y6zhO9eO1kcKv:-MlLCJ4ZATrqMDSDcuKB:b-4iqy5y) ([raw file](https://github.com/sillsdev/libpalaso/blob/123cee8d21bbfcde60a6ef6d1e43ad932c07e7b7/TestApps/ClipboardTestApp/Program.cs#L17)):*
> ```Smalltalk
> 		static void Main()
> 		{
> 			Application.EnableVisualStyles();
> ```

For Linux testing, it would be useful to have a command line argument to select between Gtk2 and Gtk3.

StephenMcConnel (approved):
:lgtm: albeit with the one comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-:-MlLDagK0LrYjYLVbpTv:b-6e9cee)* status: 20 of 24 files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @jasonleenaylor)

ermshiperete (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1106)* status: 20 of 24 files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor and @StephenMcConnel)

___
*[TestApps/ClipboardTestApp/Program.cs, line 17 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLCJ4Y6zhO9eO1kcKv:-MlLHGhP4o0bRcZKduYw:b-o68rl) ([raw file](https://github.com/sillsdev/libpalaso/blob/123cee8d21bbfcde60a6ef6d1e43ad932c07e7b7/TestApps/ClipboardTestApp/Program.cs#L17)):*

For Linux testing, it would be useful to have a command line argument to select between Gtk2 and Gtk3.
</blockquote></details>

But that in itself won't help - it will also have to start a GTK main loop - which is more than I want to get into with this PR.

ermshiperete (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1106)* status: 22 of 24 files reviewed, 5 unresolved discussions (waiting on @jasonleenaylor, @marksvc, and @StephenMcConnel)

___
*[SIL.Windows.Forms/app.config, line 10 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLMuq04-ePlSwWpFtk:-MlLT4qj57EA-qTGTLYz:b-bgml2x) ([raw file](https://github.com/sillsdev/libpalaso/blob/e03a28571c5fda0e4928b3d4c9a8a38ce7a6e1cf/SIL.Windows.Forms/app.config#L10)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

> ```
> "libxklavier"
> ```

I was surprised to see libxklavier in the list. But OK if it should go. 
</blockquote></details>

I guess that was a left-over from earlier. libxklavier is not used in SIL.Windows.Forms and so it does no good in having it in here. It's used in SIL.Windows.Forms.Keyboarding and is included in the app.config there.
___
*[SIL.Windows.Forms/SIL.Windows.Forms.csproj, line 44 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLNFeVEetKBcBiNyXb:-MlLTge5EPN-IbYw4-rk:b5obd3h) ([raw file](https://github.com/sillsdev/libpalaso/blob/e03a28571c5fda0e4928b3d4c9a8a38ce7a6e1cf/SIL.Windows.Forms/SIL.Windows.Forms.csproj#L44)):*
> It seems like.. the wrong direction to remove the reference to gtk-sharp and instead use the C libraries. My understanding is that the reason for doing this is because then we can choose gtk2 or gtk3 at run-time instead of at compile time with a reference to an assembly. But I would think we could do that with the assemblies too: at run time choosing between gtk-sharp v2 or gtk-sharp v3 assemblies using reflection. But maybe the run-time differentiation and usage of gtk-sharp 2 or 3 is much more complex than doing it with the C .so libraries.

That might also have been possible. But it feels like using libgtk dynamically is easier than using gtk-sharp via reflection. The former needs just a few method definitions. For the latter I think you'll have to redefine much more. I don't know what differences it would make in performance. I'd guess reflection to be slower since gtk-sharp has to call the libgtk methods as well, so there's an additional layer.

> Also, what impact does this have for client software that might want to access and use gtk features via C#? For example, if FieldWorks wants to use gtk-sharp to open a File Open dialog, or a Print dialog. I'm guessing that FieldWorks should go on referencing whatever gtk-sharp it wants to and using it from C#. But I also remember problems with different versions of gtk loaded at the same time. So if FieldWorks uses gtk-sharp v2 to open a File Open dialog, is there going to be a problem if FieldWorks also asks libpalaso to use gtk v3? I suspect so, and so then FieldWorks (or other client software) would just need to choose whether to use gtk-sharp v2 or v3 and match that with the gtk version it asks libpalaso to use. Is that right?

yes, it's the client applications responsibility to make sure that it's using consistent gtk versions everywhere, or to find ways of using both in parallel.
___
*[SIL.Windows.Forms/Clipboarding/LinuxClipboard.NativeMethods.cs, line 78 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLIvJC7kBJOkHUp0pY:-MlLVQMr42Ope2z46Dlr:b-426f1n) ([raw file](https://github.com/sillsdev/libpalaso/blob/21d3669b6244f69ca311e5888fbe9fe57fb3ae45/SIL.Windows.Forms/Clipboarding/LinuxClipboard.NativeMethods.cs#L78)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

> ```
> 				GraphicsManager.GtkVersionInUse == GraphicsManager.GtkVersion.Gtk2
> 					? gdk2lib
> 					: gdk3lib);
> ```

My understanding is that gtk4 is stable (as of the end of 2020). And while we may not need to do anything with it any time soon, it will be good to keep in mind, and write code in light of, that there isn't a library situation (at least more broadly than libpalaso) of "either gtk2 or gtk3", but moreso a situation of "gtk2 or gtk3 or gtk4". I don't see anything yet that sets us up poorly for this, but it will be good to keep in mind. The ternary made me think of that. 

(To anyone reading this with dismay: My sense is that we aren't the only ones who are moving to gtk3 around this time.)
</blockquote></details>

yep, I was thinking about gtk4 as well. Therefore GraphicsManager takes an enum instead of a bool. The ternary here is easily changed once we add gtk4.
___
*[SIL.Windows.Forms/Clipboarding/LinuxClipboard.NativeMethods.cs, line 100 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLHKHT2GGcbtbqnpy7:-MlLWC9U2Fpygv43wyh8:b4l2nu8) ([raw file](https://github.com/sillsdev/libpalaso/blob/21d3669b6244f69ca311e5888fbe9fe57fb3ae45/SIL.Windows.Forms/Clipboarding/LinuxClipboard.NativeMethods.cs#L100)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

> ```
> RTLD_NOW
> ```


What's the purpose of specifying this for our usage? Is it to fail faster? I see the manual says 

> RTLD_NOW
>     If this value is specified, or the environment variable LD_BIND_NOW is set to a nonempty string, all undefined symbols in the library are resolved before dlopen() returns. If this cannot be done, an error is returned. 


</blockquote></details>

I copied this code from icu-dotnet, and there it got copied from somewhere else on the internet, so I don't really know why RTLD_NOW or if it would work with something different :smile: 
___
*[SIL.Windows.Forms/Miscellaneous/GraphicsManager.cs, line 25 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLM5aU3POjH5zSNCJm:-MlLWtU2ElVA5u2m3mGn:bpedii4) ([raw file](https://github.com/sillsdev/libpalaso/blob/21d3669b6244f69ca311e5888fbe9fe57fb3ae45/SIL.Windows.Forms/Miscellaneous/GraphicsManager.cs#L25)):*

Well, we might want to change the default in the future, but right now this seems the safer change since that is what actually worked before.
</blockquote></details>

No, GTK3 won't work unless you explicitly start a GTK message loop. Winforms uses GTK2 and so the GTK2 message loop is already running. Another reason is backwards compatibility - I would be a little nervous that things would break if we make GTK3 the default without further testing.
___
*[TestApps/ClipboardTestApp/Program.cs, line 17 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1106#-MlLCJ4Y6zhO9eO1kcKv:-MlLSzpo4aLIc2iv_g6z:b2nign6) ([raw file](https://github.com/sillsdev/libpalaso/blob/123cee8d21bbfcde60a6ef6d1e43ad932c07e7b7/TestApps/ClipboardTestApp/Program.cs#L17)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

So it looks like this test program is not specifying what gtk version and so therefore it is going to be using gtk2 as per the default, is that right?
</blockquote></details>

yes

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1106)* status: 22 of 24 files reviewed, 4 unresolved discussions (waiting on @jasonleenaylor, @marksvc, and @StephenMcConnel)

PR #1111, opened 2021-10-27
Allow overriding WS root path

marksvc:
In flatpak,
`Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData)`
returns an application-specific path, like
"/home/user/.var/app/my-app/data". Allow overriding the beginning of
the basepath to continue using "/home/user/.local/share/..." to share
writing systems among applications.

+semver:minor

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1111)

### DIFF
--- a/SIL.WritingSystems/GlobalWritingSystemRepository.cs
+++ b/SIL.WritingSystems/GlobalWritingSystemRepository.cs
@@ -147,7 +147,9 @@ namespace SIL.WritingSystems
 		///<summary>
 		/// The DefaultBasePath is %CommonApplicationData%\SIL\WritingSystemRepository
 		/// On Windows 7 this is \ProgramData\SIL\WritingSystemRepository\
-		/// On Linux this must be in ~/.local/share so that it may be edited
+		/// On Linux this must be in ~/.local/share so that it may be edited.
+		/// The root may be overridden by environment variable to help in flatpak where
+		/// LocalApplicationData is application-specific.
 		///</summary>
 		public static string DefaultBasePath
 		{
@@ -156,9 +158,9 @@ namespace SIL.WritingSystems
 				// This allows unit tests to set the _defaultBasePath (through reflection)
 				if (string.IsNullOrEmpty(_defaultBasePath))
 				{
-					string basePath = Platform.IsLinux
-						? Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData)
-						: Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
+					string basePath = Environment.GetEnvironmentVariable("WSR_ROOT_PATH") ??
+						(Platform.IsLinux ? Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) :
+						Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData));
 					_defaultBasePath = Path.Combine(basePath, "SIL", "WritingSystemRepository");
 				}
 


marksvc:
The freedesktop spec at https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html says 
> $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored.

On Ubuntu 20.04, `XDG_DATA_HOME` is not set (but we know the default that freedesktop says to use, `$HOME/.local/share`), tho a user could conceivably set it to something if they desired.
In a flatpak environment,  the value is set to a location such as `$HOME/.var/app/org.sil.FieldWorks/data` -- a different "base directory relative to which user-specific data files should be stored" than we might have expected, but a valid one.

In a recent code review, @papeh  asked some questions about the use of `WSR_ROOT_PATH`. We talked about whether there is any use case where libpalaso _shouldn't_ store its writing system information in the shareable `$HOME/.local/share` location. I thought about what I would do if we found that other parts of libpalaso were becoming candidates for also storing their data in the shareable `$HOME/.local/share` location. 

If there are no use cases for libpalaso to keep its WS information in the private location, then it may be okay for libpalaso to keep its WS information in the shareable location (like, without the client application asking for it). 
If we found that libpalaso started having other pieces that we wanted to keep in a shareable location, then rather than a client application start to have to set `WSR_ROOT_PATH`, `SOMETHINGELSE_ROOT_PATH`, `BLAH_ROOT_PATH`, it would be nicer to consolidate it down to a single setting, like `USER_DATA_HOME` or similar. My inclination would be to stay with the `WSR_ROOT_PATH` for now, and at a later time if we find that another part of libpalaso needs something similar, to move to a `USR_DATA_HOME` usage. Hasso preferred for using the consolidated setting ahead of time before waiting to find a next thing in libpalaso that would need it. 

And so I'm thinking about making a change to libpalaso GlobalWritingSystemRepository.cs DefaultBasePath in this regard. Such as:  
```
if (string.IsNullOrEmpty(_defaultBasePath))
{
-  string basePath = Environment.GetEnvironmentVariable("WSR_ROOT_PATH") ??
+  string basePath = Environment.GetEnvironmentVariable("USER_DATA_HOME") ??
    (Platform.IsLinux ? Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) :
```
And of course an option would be to use the `HOST_XDG_DATA_HOME` and we would ensure that is set in a flatpak application (since it's not set if the host `XDG_DATA_HOME` is not set):
```
if (string.IsNullOrEmpty(_defaultBasePath))
{
-  string basePath = Environment.GetEnvironmentVariable("WSR_ROOT_PATH") ??
+  string basePath = Environment.GetEnvironmentVariable("HOST_XDG_DATA_HOME") ??
    (Platform.IsLinux ? Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) :
```
In either example, the libpalaso library would be doing: Put this in an alleged more findable/sharable location, if defined, or go ahead and use the location we know about. 

Changing from `WSR_ROOT_PATH` to `USER_DATA_HOME` retains some of the same pros and cons that `WSR_ROOT_PATH` had, with notable changes being
- More generic variable name that would be more sensible to re-use in other parts of libpalaso some day, if we find we have multiple pieces that should always be in the more shareable area.
- libpalaso would kind of be _trying_ to store its WS data in the more shareable `$HOME/.local/share` location, rather than more passively being _told_ to.

@ermshiperete : Do you have an opinion on what to do here? I guess this is somewhat  "much ado about nothing", but little changes can lead to long-lived annoyances, so it can be good to get them right at the beginning.

ermshiperete:
I like the more generic variable name `USER_DATA_HOME`. Makes sense. And yes, it's easier done right at the beginning than later having to migrate data to the new location.

papeh:
Should we be checking XDG_DATA_HOME before or after Platform.IsLinux?

ermshiperete:
The mono implementation of `Environment.GetFolderPath` checks `XDG_DATA_HOME`.

marksvc:
@papeh ,

> Should we be checking XDG_DATA_HOME before or after Platform.IsLinux?

(Assuming you mean USER_DATA_HOME in #1113. )
Right; good question. It could be either. Checking before would mean another OS could make use of it, tho that might never happen. 

I'll leave #1113 open for a bit in case you want it to be changed.

PR #1116, opened 2021-11-17
Fixed code for positioning modal child window

tombogle:
when parent is on screen whose coordinates are negative (wrt primary screen).
Note: Technically, this code should perhaps determine what screen the parent is on and then use that screen's upper left corner as the minimum location, but typically a child window will be smaller than the parent, so it will seldom matter.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1116)

### DIFF
--- a/SIL.Windows.Forms/ParentFormBase.cs
+++ b/SIL.Windows.Forms/ParentFormBase.cs
@@ -135,8 +135,8 @@ namespace SIL.Windows.Forms
 			if (childForm.StartPosition == FormStartPosition.CenterParent)
 			{
 				childForm.StartPosition = FormStartPosition.Manual;
-				childForm.Location = new Point(Math.Max(0, Location.X + (Width - childForm.Width) / 2),
-					Math.Max(0, Location.Y + (Height - childForm.Height) / 2));
+				childForm.Location = new Point(Math.Max(Location.X, Location.X + (Width - childForm.Width) / 2),
+					Math.Max(Location.Y, Location.Y + (Height - childForm.Height) / 2));
 			}
 			childForm.Show(this);
 			OnModalFormShown?.Invoke();


andrew-polk (approved):

FoolRunning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1116)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[SIL.Windows.Forms/ParentFormBase.cs, line 139 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1116#-Moj7N7hA46bl2v7ilfr:-Moj7N7hA46bl2v7ilfs:b-a0rn6f) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ca3a303ad6d730d2e934f4466b6749c4b216f8/SIL.Windows.Forms/ParentFormBase.cs#L139)):*
> ```Smalltalk
> 				childForm.StartPosition = FormStartPosition.Manual;
> 				childForm.Location = new Point(Math.Max(Location.X, Location.X + (Width - childForm.Width) / 2),
> 					Math.Max(Location.Y, Location.Y + (Height - childForm.Height) / 2));
> ```

With your changes, I don't think it will work well if the child window is larger (width/height) than this window.

FoolRunning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1116)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[SIL.Windows.Forms/ParentFormBase.cs, line 139 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1116#-Moj7N7hA46bl2v7ilfr:-Moj7mPw6QxnqDMrTbZe:b-9cd8zs) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ca3a303ad6d730d2e934f4466b6749c4b216f8/SIL.Windows.Forms/ParentFormBase.cs#L139)):*

With your changes, I don't think it will work well if the child window is larger (width/height) than this window.
</blockquote></details>

Unfortunately, to do this properly, you probably need to look at the bounds of the screen that owns this form and use those bounds as the minimum values.

tombogle:
___
*[SIL.Windows.Forms/ParentFormBase.cs, line 139 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1116#-Moj7N7hA46bl2v7ilfr:-MonLsL_A6i6twx4tg-M:b5imozp) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ca3a303ad6d730d2e934f4466b6749c4b216f8/SIL.Windows.Forms/ParentFormBase.cs#L139)):*

Unfortunately, to do this properly, you probably need to look at the bounds of the screen that owns this form and use those bounds as the minimum values.
</blockquote></details>

Yeah, that's what I said in the PR details. I think it it generally fairly rare for a dialog box to be larger than the parent window. Combined with the (somewhat) rarity of having a screen with negative coordinates, I wasn't necessarily sure it was worth the additional effort. Worse case (I think) is that instead of being centered over the parent, it will just cover it from the upper left corner. In some weird edge cases, that obviously could mean that the dialog box ends up being partially off-screen, but it should generally mean that enough of the title bar is visible to be able to grab it and move it to a better position. I'm willing to do the additional work to truly center it as much as possible if you think it's worth it.

FoolRunning (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1116#-:-MoncabP2fnYarmdE5qt:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[SIL.Windows.Forms/ParentFormBase.cs, line 139 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1116#-Moj7N7hA46bl2v7ilfr:-MoncZaU2zXm2e1hapl7:b-4kln3d) ([raw file](https://github.com/sillsdev/libpalaso/blob/d5ca3a303ad6d730d2e934f4466b6749c4b216f8/SIL.Windows.Forms/ParentFormBase.cs#L139)):*

Yeah, that's what I said in the PR details. I think it it generally fairly rare for a dialog box to be larger than the parent window. Combined with the (somewhat) rarity of having a screen with negative coordinates, I wasn't necessarily sure it was worth the additional effort. Worse case (I think) is that instead of being centered over the parent, it will just cover it from the upper left corner. In some weird edge cases, that obviously could mean that the dialog box ends up being partially off-screen, but it should generally mean that enough of the title bar is visible to be able to grab it and move it to a better position. I'm willing to do the additional work to truly center it as much as possible if you think it's worth it.
</blockquote></details>

Yeah, I guess it's not worth the effort.

PR #1117, opened 2021-11-30
Added SettingsProtectionHelper.RemoveManagedComponent

tombogle:
Added a way to get settings protection text to make sense when the UI to access it is via menu item instead of a button.
Fixed a typo and other cosmetic stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @hatton, and @tombogle)

___
*[SIL.Windows.Forms/SettingProtection/SettingProtectionDialog.cs, line 60 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-MpqmGrI2lovLrH71ych:-MpqmGrI2lovLrH71yci:bja8y0k) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingProtectionDialog.cs#L60)):*
> ```Smalltalk
> 			}
> 
> 			_normallyHiddenCheckbox.Tag = _normallyHiddenCheckbox.Text;
> ```

Needs a comment like "Save the original text for possible restoration when setting SettingsProtectionIsViaMenuItem back to false."
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 170 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-MpqlPhC6DkKv7vt_RWi:-MpqlPhC6DkKv7vt_RWj:bux2phi) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L170)):*
> ```Smalltalk
> 			if (_controlIsUnderSettingsProtection.ContainsKey(controlOrToolStripItem))
> 				_controlIsUnderSettingsProtection.Remove(controlOrToolStripItem);
> 			_controlIsUnderSettingsProtection.Add(controlOrToolStripItem, true);
> ```

Aren't these three lines equivalent to _controlIsUnderSettingsProtection[controlOrToolStripItem] = true?
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-Mpqmsim0RaJujdJfQHA:blprtsf) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*
> ```Smalltalk
> 		/// for a component that is not a control.
> 		/// </remarks>
> 		public void RemoveManagedComponent(Component controlOrToolStripItem)
> ```

Since it gets put in the dictionary by a method ManageComponent (rather than, say, AddManagedComponent), I wonder whether a better name would be StopManagingComponent? Or UnmanageComponent?
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 182 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-MpqllcP8EKcIfKmSeIT:-MpqllcP8EKcIfKmSeIU:boggbg5) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L182)):*
> ```Smalltalk
> 		public void RemoveManagedComponent(Component controlOrToolStripItem)
> 		{
> 			if (_controlIsUnderSettingsProtection.ContainsKey(controlOrToolStripItem))
> ```

Unnecessary. Remove doesn't care if it's not there.

JohnThomson (commented):
Basically looks good, a few minor points.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-:-MpqoLwZ4e1EQgh-ZSRR:bplomd9)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @hatton, and @tombogle)

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MprU6H40QQ9fB4TJ33O:b-gsldiw) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

Since it gets put in the dictionary by a method ManageComponent (rather than, say, AddManagedComponent), I wonder whether a better name would be StopManagingComponent? Or UnmanageComponent?
</blockquote></details>

Or Mismanage! :-) I'm torn between StopManagingComponent and RemoveSettingsProtection. The latter feels like a good counter to SetSettingsProtection, which is presumably what ManageComponent wishes it could have been called.

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MprWpQPBNdp1n9k3NkU:bkcx091) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

Or Mismanage! :-) I'm torn between StopManagingComponent and RemoveSettingsProtection. The latter feels like a good counter to SetSettingsProtection, which is presumably what ManageComponent wishes it could have been called.
</blockquote></details>

Here's an alternative: public void SetSettingsProtection(Component c, bool isProtected)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @hatton, @JohnThomson, and @tombogle)

___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpryPf4Dvr64Jpy4JRG:b-ritnz7) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

Here's an alternative: public void SetSettingsProtection(Component c, bool isProtected)
</blockquote></details>

We already have such a method. However, where the new method removes the key from the dictionary, SetSettingsProtection just changes the boolean value in the dictionary (well, it will add the key if not already present). I don't remember exactly, but I _think_ there's a difference between "not in the dictionary" and "in with value false"....if not, we probably ought to be using a set which a component either is or is not a member of, and may not need a new method at all.

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpsDd8y2sdvKSUC8uzN:b-h50mr0) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

We already have such a method. However, where the new method removes the key from the dictionary, SetSettingsProtection just changes the boolean value in the dictionary (well, it will add the key if not already present). I don't remember exactly, but I _think_ there's a difference between "not in the dictionary" and "in with value false"....if not, we probably ought to be using a set which a component either is or is not a member of, and may not need a new method at all.
</blockquote></details>

I implemented the new version to also leave it in the dictionary, so the behavior is the same. There appears to be a very slight (though not necessarily intentional) difference between a control that is not in the dictionary and one that is present with a value of false. In UpdateDisplay, a control that is present with a value of false will always be made visible. So if the client code (or in Designer), it was set to be invisible but it is under settings protection, it with be forced to be visible. That's not a problem for HearThis, but neither is it a benefit. I'm not sure if any existing software other than HearThis is using any of this SettingsProtection stuff.

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpsGSyTBgZYmZYE40N6:b-lhcvzv) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

I implemented the new version to also leave it in the dictionary, so the behavior is the same. There appears to be a very slight (though not necessarily intentional) difference between a control that is not in the dictionary and one that is present with a value of false. In UpdateDisplay, a control that is present with a value of false will always be made visible. So if the client code (or in Designer), it was set to be invisible but it is under settings protection, it with be forced to be visible. That's not a problem for HearThis, but neither is it a benefit. I'm not sure if any existing software other than HearThis is using any of this SettingsProtection stuff.
</blockquote></details>

Ah, looks like BloomDesktop uses it too.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @hatton, @JohnThomson, and @tombogle)

___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpsMdS2EwagLD27dscT:b-2rj8d5) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

Ah, looks like BloomDesktop uses it too.
</blockquote></details>

I don't think Bloom has any controls that are hidden by default unless SettingsProtection makes them visible. I suppose we can't be absolutely sure no one is depending on that behavior, but without comments calling out the difference it seems very unlikely that anyone is depending on the difference in behavior, so if we wanted to simplify things to a set of controls that are visible when settings protection is off and hidden when it is on, I'd think it an improvement. @ermshiperete might have stronger feelings about even ever-so-subtle breaking changes.
You talk about a new version of something...I would have thought SetSettingsProtection...but I don't see a new version of it.

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-Mpt0aVECJk97SOUoRrx:b4xbttx) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

I don't think Bloom has any controls that are hidden by default unless SettingsProtection makes them visible. I suppose we can't be absolutely sure no one is depending on that behavior, but without comments calling out the difference it seems very unlikely that anyone is depending on the difference in behavior, so if we wanted to simplify things to a set of controls that are visible when settings protection is off and hidden when it is on, I'd think it an improvement. @ermshiperete might have stronger feelings about even ever-so-subtle breaking changes.
You talk about a new version of something...I would have thought SetSettingsProtection...but I don't see a new version of it.
</blockquote></details>

By "new version" I just meant that when I changed the name I also changed the implementation (of the new overload of SetSettingsProtection) to: _controlIsUnderSettingsProtection[c] = isProtected;
So, like the existing version that takes a Control, this one leaves the component in the dictionary. Of course, all this becomes moot if we change from a dictoinary to a HashSet. We'll see if @ermshiperete weighs in. I'm sort of inclined not to do this as part of this PR, partly because it will be easier to revert if it turns out we actually broke something and partly because I think it will be easier to see the distinct intention of the two changes if they are merged separately. BUt I don't feel super strongly about it.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @hatton, and @JohnThomson)

___
*[SIL.Windows.Forms/SettingProtection/SettingProtectionDialog.cs, line 60 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-MpqmGrI2lovLrH71ych:-MprWuj3C9_eEIDiQNL_:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingProtectionDialog.cs#L60)):*

Needs a comment like "Save the original text for possible restoration when setting SettingsProtectionIsViaMenuItem back to false."
</blockquote></details>

Done.
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 170 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-MpqlPhC6DkKv7vt_RWi:-MprSSkR3jsJE4iMyT7M:b-rmv1y5) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L170)):*
> ```
> if (_controlIsUnderSettingsProtection.ContainsKey(controlOrToolStripItem))
> ```
Yes. I'm not very sure why this wasn't just implemented as a HashSet. I don't think there's any meaningful difference between a value of false and not having the control/component in the dictionary.
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 182 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-MpqllcP8EKcIfKmSeIT:-MprWtv8BgbMqhWtb5sC:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L182)):*

Unnecessary. Remove doesn't care if it's not there.
</blockquote></details>

Done.

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpvwZ2Y5_xgFPDLV76N:bklcdeo) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

By "new version" I just meant that when I changed the name I also changed the implementation (of the new overload of SetSettingsProtection) to: _controlIsUnderSettingsProtection[c] = isProtected;
So, like the existing version that takes a Control, this one leaves the component in the dictionary. Of course, all this becomes moot if we change from a dictoinary to a HashSet. We'll see if @ermshiperete weighs in. I'm sort of inclined not to do this as part of this PR, partly because it will be easier to revert if it turns out we actually broke something and partly because I think it will be easier to see the distinct intention of the two changes if they are merged separately. BUt I don't feel super strongly about it. 
</blockquote></details>

I decided that it was worth beefing up the test dialog to demonstrate the new functionality. Then as an experiment, I tried making the change form a Dictionary to a Hashset. Thus I discovered the subtle rationale for keeping the false entries in the dictionary. If a component is simply removed, there is no code to make it visible again, so if the caller doesn't do it explicitly, it will remain hidden until the form that contains it is re-constructed. (If this is the main form of the application -- as it is in HearThis -- it means it won't reappear until the application is restarted.) An easy fix is just to make it visible when removing it from the HashSet. That at least allows for the application to hide it for some other reason if need be.

tombogle:
___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpvzFYIAxcpdrhTGBAu:bshtdam) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

I decided that it was worth beefing up the test dialog to demonstrate the new functionality. Then as an experiment, I tried making the change form a Dictionary to a Hashset. Thus I discovered the subtle rationale for keeping the false entries in the dictionary. If a component is simply removed, there is no code to make it visible again, so if the caller doesn't do it explicitly, it will remain hidden until the form that contains it is re-constructed. (If this is the main form of the application -- as it is in HearThis -- it means it won't reappear until the application is restarted.) An easy fix is just to make it visible when removing it from the HashSet. That at least allows for the application to hide it for some other reason if need be. 
</blockquote></details>

I have added a new PR (https://github.com/sillsdev/libpalaso/pull/1118) against this branch with those changes.  We can merge it into this branch, wait and merge it into master after this branch is merged, or decide it isn't worth it and close it without merging.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-:-MpwP7qV8iCfCVcb_loR:bnfp4nl)* status: 1 of 11 files reviewed, all discussions resolved (waiting on @andrew-polk, @ermshiperete, @hatton, and @JohnThomson)

___
*[SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs, line 180 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1117#-Mpqmsim0RaJujdJfQH9:-MpwNjjB7EqwNOr4Prlb:bs3d162) ([raw file](https://github.com/sillsdev/libpalaso/blob/411448fce1bad51e23b1ef9154353e60a5a1c343/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L180)):*

I have added a new PR (https://github.com/sillsdev/libpalaso/pull/1118) against this branch with those changes.  We can merge it into this branch, wait and merge it into master after this branch is merged, or decide it isn't worth it and close it without merging.
</blockquote></details>

I had missed the fact that the two overloads take a different class of Control to set the protection of. And for some reason Reviewable was not showing me the latest changes. I like what you now have, though I like it even better changed to a Hashset as in your other PR.

PR #1118, opened 2021-12-02
Changed underlying implementation in SettingsProtectionHelper to use a HashSet…

tombogle:
…instead of a Dictionary to track managed controls. This is simpler and slightly more efficient. It does cause a very subtle change in an edge case, but the previous behavior was almost certainly unintended: previously a component that was added to the dictionary with a value of false (meaning that it was not supposed to be managed) would always be made explicitly visible every time UpdateDisplay was called. This meant that an application could not explicitly hide it (for some other reason) and keep it hidden.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1118)

### DIFF
--- a/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.Designer.cs
+++ b/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.Designer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace SIL.Windows.Forms.SettingProtection
 {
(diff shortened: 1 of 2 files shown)


JohnThomson (approved):
LGTM. Much less room for confusion.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1118#-:-MpwNJ0H5mboKyElPJ03:bri8gcl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

PR #1119, opened 2021-12-02
Ensure components are made visible when no longer being managed by SettingsProtectionHelper

tombogle:
Note: this code was supposed to be part of previous merge, but was inadvertently omitted.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1119)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms] Options for `FlexibleMessageBox` to show in the taskbar and to show on top of other windows
 - [SIL.Windows.Forms.DblBundle] virtual method `SelectProjectDlgBase.CreateFileDialog()` to allow
   customization in derived class (#797)
+- [SIL.Windows.Forms.SettingProtection] overload of SetSettingsProtection method that takes a ToolStripItem
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


JohnThomson (approved):
LGTM. Definitely an improvement as-is.
However, some thoughts on further improvements:
1. Consider deprecating ManageComponent, as it is redundant with SetSettingsProtection (especially if you made the second argument to that default true).
2. Consider a private overload of SetSettingsProtection with argument of type Component, which the two current overloads would trivially call, so the common implementation would only be in one place.
3. Consider making the current ManageComponent private, and creating two public overloads which call it but take the two actually possible types. (This would technically be a breaking change, but only code that would previously have thrown would now fail to compile.)
4. Do you definitely not want SetSettingsProtection to call ShowOrHideComponent when setting true? It feels like a lack of parallelism that you need to call UpdateDisplay to get the effect when setting true but not when setting false. If there is a good reason for it, documenting it would be helpful.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1119#-:-MqKFxilCFej6v-h3TU3:br8ynaw)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk, @ermshiperete, and @FoolRunning)

JohnThomson (approved):
Actually, with a private component version of SetSettingsProtection, both overloads of ManageComponent would be trivial calls to it, too.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1119#-:-MqKHwvBA7fbdbaNZErk:b-9x6l05)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk, @ermshiperete, and @FoolRunning)

ermshiperete (approved):

Issue #763, opened 2018-10-17
Region list of "Script and Variant" dialog shows deprecated subtags


ermshiperete:
https://www.w3.org/International/questions/qa-choosing-language-tags#regionsubtag mentions that deprecated subtags should be avoided when creating language tags. Yet they appear in the Region list of the "Script and Variant" dialog (e.g. German Democratic Republic).

glasseyes:
libpalaso knows about regions and subtags that are deprecated. tag.IsDeprecated. Whatever it is that is listing them needs to filter out the deprecated ones.

PR #1126, opened 2022-01-25
ignore deprecated region subtags (#763)

elisunger:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1126)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   gtk-sharp
 - [SIL.Windows.Forms.SettingProtection] Deprecated ManageComponent method
 - [SIL.Scripture] VerseRef.TrySetVerseUnicode: Improve handling of non-decimal numerals and surrogate pair numerals (#1000)
+- [SIL.WritingSystems] Ignore deprecated region subtags (#763)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   gtk-sharp
 - [SIL.Windows.Forms.SettingProtection] Deprecated ManageComponent method
 - [SIL.Scripture] VerseRef.TrySetVerseUnicode: Improve handling of non-decimal numerals and surrogate pair numerals (#1000)
-- [SIL.WritingSystems] Ignore deprecated region subtags in `ScriptRegionVariantView`(#763)
+- [SIL.Windows.Forms.WritingSystems] Ignore deprecated region subtags in `ScriptRegionVariantView`(#763)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   gtk-sharp
 - [SIL.Windows.Forms.SettingProtection] Deprecated ManageComponent method
 - [SIL.Scripture] VerseRef.TrySetVerseUnicode: Improve handling of non-decimal numerals and surrogate pair numerals (#1000)
-- [SIL.WritingSystems] Ignore deprecated region subtags (#763)
+- [SIL.WritingSystems] Ignore deprecated region subtags in `ScriptRegionVariantView`(#763)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms] `PortableClipboard` uses unmanaged libgtk/libgdk methods instead of using
   gtk-sharp
 - [SIL.Windows.Forms.SettingProtection] Deprecated ManageComponent method
+- [SIL.Scripture] VerseRef.TrySetVerseUnicode: Improve handling of non-decimal numerals and surrogate pair numerals (#1000)
 
 ### Fixed
 
--- a/SIL.WritingSystems.Tests/StandardSubtagsTests.cs
+++ b/SIL.WritingSystems.Tests/StandardSubtagsTests.cs
@@ -130,13 +130,6 @@ namespace SIL.WritingSystems.Tests
 		}
 
 		[Test]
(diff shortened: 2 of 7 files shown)


jasonleenaylor (commented):
I think this is good based on my memory of how and where the StandardSubtags are used. But I just wanted to verify, tags with deprecated regions will still validate so existing projects will not have trouble, and in FieldWorks the user can still use them if the enter the advanced view and type them manually?

> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1126#-:-MuLikLKCdN5fIMovrgN:b-lh4j4m)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

ermshiperete:
Thanks! Yes, you're right. Validation of existing tags with deprecated regions wouldn't be possible anymore. So it sounds like it would be best to leave `RegisteredRegions` as it is, containing deprecated regions, and filter out the deprecated regions in the "Script and Variant" dialog.

@elisunger Could you please make the changes? That would be in [ScriptRevionVariantView.cs](https://github.com/sillsdev/libpalaso/blob/a971ea676fe26e2a7b4b6bc97d7c4cd02fe02df4/SIL.Windows.Forms.WritingSystems/WSIdentifiers/ScriptRegionVariantView.cs#L27)

CHANGELOG.md
```diff
@@ -45,7 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   gtk-sharp
 - [SIL.Windows.Forms.SettingProtection] Deprecated ManageComponent method
 - [SIL.Scripture] VerseRef.TrySetVerseUnicode: Improve handling of non-decimal numerals and surrogate pair numerals (#1000)
-- [SIL.WritingSystems] Ignore deprecated region subtags (#763)
+- [SIL.WritingSystems] Ignore deprecated region subtags in `ScriptRegionVariantView`(#763)
```
ermshiperete, line 48:
```suggestion
- [SIL.Windows.Forms.WritingSystems] Ignore deprecated region subtags in `ScriptRegionVariantView`(#763)
```

There should also be an entry under Added that mentions the additional `StandardSubtags.FilteredRegions` property.

And since we're adding a new property, the commit message should have a `+semver:minor` line.

ermshiperete (changes requested):
@elisunger This is a solution that would work, but it increases the memory footprint since we store almost the same information twice now.

I think a better approach would be to filter out the deprecated tags in `ScriptRegionVariantView` before adding to the combo.

SIL.WritingSystems/StandardSubtags.cs
```diff
@@ -149,7 +149,7 @@ internal static void InitialiseIanaSubtags(string twotothreecodes, string subtag
 			RegisteredLanguages = new KeyedList<string, LanguageSubtag>(sortedLanguages, l => l.Code, StringComparer.InvariantCultureIgnoreCase);
 			RegisteredScripts = new KeyedList<string, ScriptSubtag>(scripts.OrderBy(s => s.Name), s => s.Code, StringComparer.InvariantCultureIgnoreCase);
 			RegisteredRegions = new ReadOnlyKeyedCollection<string, RegionSubtag>(new KeyedList<string, RegionSubtag>(regions.OrderBy(r => r.Name), r => r.Code, StringComparer.InvariantCultureIgnoreCase));
-			FilteredRegions = new ReadOnlyKeyedCollection<string, RegionSubtag>(new KeyedList<string, RegionSubtag>(regions.FindAll(f => !f.IsDeprecated).OrderBy(f => f.Name), f => f.Code, StringComparer.InvariantCultureIgnoreCase));
+			//FilteredRegions = new ReadOnlyKeyedCollection<string, RegionSubtag>(new KeyedList<string, RegionSubtag>(regions.FindAll(f => !f.IsDeprecated).OrderBy(f => f.Name), f => f.Code, StringComparer.InvariantCultureIgnoreCase));
```
ermshiperete, line 152: Please remove the commented code

SIL.WritingSystems.Tests/StandardSubtagsTests.cs
```diff
@@ -130,9 +130,9 @@ public void RegisteredRegions_PrivateUse([Values("AA", "ZZ")] string code)
 		}
 
 		[Test]
-		public void FilteredRegions_DoesNotHaveDeprecatedRegion()
+		public void RegisteredRegions_DoesNotHaveDeprecatedRegion()
 		{
-			Assert.That(StandardSubtags.FilteredRegions.Contains("DD"), Is.False);
+			Assert.That(StandardSubtags.RegisteredRegions.Contains("DD"), Is.False);
```
ermshiperete, line 135: That won't work...

ermshiperete (changes requested):

SIL.WritingSystems/StandardSubtags.cs
```diff
@@ -166,6 +166,8 @@ internal static void InitialiseIanaSubtags(string twotothreecodes, string subtag
 
 		public static IReadOnlyKeyedCollection<string, RegionSubtag> RegisteredRegions { get; private set; }
 
+		//public static IReadOnlyKeyedCollection<string, RegionSubtag> FilteredRegions { get; private set; }
```
ermshiperete, line 169: Sorry, here's another line of commented code...

ermshiperete (changes requested):

ermshiperete:
@jasonleenaylor, @FoolRunning Could you take another look, please?

jasonleenaylor:
This looks good, I have one user scenario question though. What is the user experience like if existing data has a deprecated region?

ermshiperete:
> This looks good, I have one user scenario question though. What is the user experience like if existing data has a deprecated region?

Good question. I tested that in the test app:

If the user has an existing writing system that uses a deprecated region, and he goes to the Writing System dialog to modify the WS, the "Region" field will be empty. However, the dialog still remembers the region (and displays it in the tag at the top right), so unless the user selects a different region the region won't change, even if he changes other parts of the WS. However, if he selects a different region there is no way to get back to the deprecated region. 

I haven't tested this in FW which might have a slightly different WS dialog.

ermshiperete (approved):

PR #1138, opened 2022-03-09
LT-20651: Prevent creation of null files

mark-sil:
These changes are an attempt to prevent the creation of null
files. The theory is that FieldWorks is closing and power is lost
before the operating system cache is written to disk. These changes
cause the cache to be flushed immediately after writing instead of
relying on the cache manager lazy writing mechanism.
https://docs.microsoft.com/en-us/windows/win32/fileio/file-caching

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1138)

### DIFF
--- a/SIL.Core/IO/FileHelper.cs
+++ b/SIL.Core/IO/FileHelper.cs
@@ -1,8 +1,10 @@
 // Copyright (c) 2018 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 
+using System;
 using System.IO;
 using System.Text.RegularExpressions;
+using System.Xml.Serialization;
 
 namespace SIL.IO
 {
@@ -39,6 +41,44 @@ namespace SIL.IO
 				}
 			}
 			return false;
+		}
+
+		/// <summary>
+		/// Serialize an object directly to disk, avoiding the operating system cache.
+		/// 
+		/// This method is intended to prevent the problem with null files being generated
+		/// as reported here: LT-20651, LT-20333, LTB-3915, LTB-3916, and LTB-3917
+		/// The theory is that FieldWorks is closing and power is lost before the operating
+		/// system cache is written to disk. This method is intended to prevent that problem
+		/// by avoiding the cache and writing directly to disk.
+		///
+		/// The disadvantage of this method is that it is slow; so it should only be used to
+		/// write relatively small files that are not frequently written.
+		/// </summary>
+		/// <param name="objToSerialize">Object to be serialized to the file.</param>
+		/// <param name="path">The full path (containing the file name and extension).</param>
+		public static void WriteXmlFileDirectlyToDisk(object objToSerialize, string path)
+		{
+			try
+			{
+				// Note: Using FileOptions.WriteThrough causes the data to still be written to the
+				//       operating system cache but it is immediately flushed. If this doesn't address
+				//       the problem then a more thorough solution that completely bypasses the cache is
+				//       to use the c++ CreateFile() api and pass both FILE_FLAG_NO_BUFFERING and
+				//       FILE_FLAG_WRITE_THROUGH.
+				//       https://docs.microsoft.com/en-us/windows/win32/fileio/file-caching
+				using (var writer = new FileStream(path, FileMode.Create,
+					FileAccess.Write, FileShare.None,
+					4096, FileOptions.WriteThrough))
+				{
+					XmlSerializer xmlSerializer = new XmlSerializer(objToSerialize.GetType());
+					xmlSerializer.Serialize(writer, objToSerialize);
+				}
+			}
+			catch (Exception err)
+			{
+				throw new ApplicationException("There was a problem saving your xml file.", err);
+			}
 		}
 	}
 }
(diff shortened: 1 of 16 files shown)


SIL.Core/IO/FileHelper.cs
```diff
@@ -40,5 +42,43 @@ public static bool Grep(string inputPath, string pattern)
...
+		/// The theory is that FieldWorks is closing and power is lost before the operating
+		/// system cache is written to disk. This method is intended to prevent that problem
+		/// by avoiding the cache and writing directly to disk.
+		///
+		/// The disadvantage of this method is that it is slow; so it should only be used to
+		/// write relatively small files that are not frequently written.
+		/// </summary>
+		/// <param name="objToSerialize">Object to be serialized to the file.</param>
+		/// <param name="path">The full path (containing the file name and extension).</param>
+		public static void WriteXmlFileDirectlyToDisk(object objToSerialize, string path)
```
tombogle, line 60: RobustFile already has a Create method that creates a FileStream with WriteThrough. I wonder if you can just use that or make this method use it. In any case, it kind of seems that this logic maybe belongs in that class rather than in FileHelper.

tombogle: Alternatively, could this be put as another overload on XmlSerializationHelper.SerializeToFile? That would probably make it more easily discoverable.

mark-sil: Thanks Tom, I wasn't aware of these classes.  I'll take a look and see where it seems to fit best.

mark-sil: @tombogle I moved the utility to XmlSerializationHelper and changed it to call RobustFile.Create().

jasonleenaylor (commented):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1138)* status: all files reviewed, 1 unresolved discussion (waiting on @papeh and @tombogle)

___
*[SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs, line 460 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1138#-MxoBVSZ-u1mtLdIUhWn:-MxoBVSZ-u1mtLdIUhWo:b-d8r31u) ([raw file](https://github.com/sillsdev/libpalaso/blob/f7b0c78dd79fee496b1d1e3e230719912d56651d/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs#L460)):*
> ```Smalltalk
> 			try
> 			{
> 				XElement element = new XElement("test"); ;
> ```

A few extra characters slipped in here.



_Code quote:_
```Smalltalk
 ;
```
___
*[SIL.Lexicon/FileSettingsStore.cs, line 35 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1138#-MxoC9Ls1dSnv9CIngxQ:-MxoC9Ls1dSnv9CIngxR:buu8mus) ([raw file](https://github.com/sillsdev/libpalaso/blob/f7b0c78dd79fee496b1d1e3e230719912d56651d/SIL.Lexicon/FileSettingsStore.cs#L35)):*
> ```Smalltalk
> 			XmlSerializationHelper.SerializeToFile_AvoidCache(_settingsFilePath,
> 				_settingsDoc.Root, out ex);
> 			if (ex != null)
> ```

Unless we have a use case planned (maybe in another user of the API) where we check the return value and don't rethrow the exception I'd simplify the api by letting the method throw the exception.



_Code quote:_
```Smalltalk
			if (ex != null)
				throw ex;
```

SIL.Core/Xml/XmlSerializationHelper.cs
```diff
@@ -273,6 +274,48 @@ public static bool SerializeToFile<T>(string filename, T data, string rootElemen
...
+		/// system cache is written to disk. This method is intended to prevent that problem
+		/// by avoiding the cache by writing through the cache directly to disk.
+		///
+		/// The disadvantage of this method is that it is slow; so it should only be used to
+		/// write relatively small files that are not frequently written.
+		/// </summary>
+		/// <param name="data">Object to be serialized to the file.</param>
+		/// <param name="path">The full path (containing the file name and extension).</param>
+		/// ------------------------------------------------------------------------------------
+		public static bool SerializeToFile_AvoidCache<T>(string path, T data, out Exception e)
```
tombogle, line 293: I think this is better. I do kind of wonder about the advisability of using a two-part name as opposed to adding another overload to SerializeToFile, with a Boolean parameter, But perhaps the value of that parameter would always be hard-coded, and the "false" case would always end up calling a simpler overload. Searching all of libpalaso, I do not see any precedence for a two-part name for a public API method. So I would suggest instead naming it SerializeToFileWithWriteThrough, or something like that.

tombogle (changes requested):
Change name to not have an underscore

mark-sil:
Hi @jasonleenaylor and @tombogle I think I have finished the changes you requested.  Is there anything else you want changed in this PR?

mark-sil:
Hi @jasonleenaylor and @tombogle I think I have finished the changes you requested.  Is there anything else you want changed in this PR?

jasonleenaylor (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1138)* status: all files reviewed, 2 unresolved discussions (waiting on @papeh and @tombogle)

tombogle (approved):

PR #1139, opened 2022-03-14
HT-402: Moved lock to cover check for initial _recordingState in BeginRecording

tombogle:
I'm looking into HT-402. I don't have any good reason to believe this could be the cause of that problem, but it does seem like this lock should be done earlier lest the value change between the time we check it and the time we set it. I don't think this could inadvertently result in a new deadlock. If it does, then it probably shows there is a real problem, but because none of the unit tests probably are capable of testing this well even if they tried, it is worth careful scrutiny.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1139)

### DIFF
--- a/SIL.Media/Naudio/AudioRecorder.cs
+++ b/SIL.Media/Naudio/AudioRecorder.cs
@@ -126,8 +126,7 @@ namespace SIL.Media.Naudio
 			RecordedTime = TimeSpan.Zero;
 			RecordingState = RecordingState.Monitoring;
 			_fileWriterThread.Abort();
-			if (Stopped != null)
-				Stopped(this, new ErrorEventArgs(new Exception("NAudio recording stopped unexpectedly")));
+			Stopped?.Invoke(this, new ErrorEventArgs(new Exception("NAudio recording stopped unexpectedly")));
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -135,12 +134,26 @@ namespace SIL.Media.Naudio
 		{
 			if (_waveIn != null)
 			{
-				try { _waveIn.StopRecording(); }
-				catch { /* It's amazing all the bizarre things that can go wrong */ }
+				try
+				{
+					_waveIn.StopRecording();
+				}
+				catch (Exception e)
+				{
+					// It's amazing all the bizarre things that can go wrong.
+					Logger.WriteError(e);
+				}
 				_waveIn.DataAvailable -= waveIn_DataAvailable;
 				_waveIn.RecordingStopped -= OnRecordingStopped;
-				try { _waveIn.Dispose(); }
-				catch { /* It's amazing all the bizarre things that can go wrong */ }
+				try
+				{
+					_waveIn.Dispose();
+				}
+				catch (Exception e)
+				{
+					// It's amazing all the bizarre things that can go wrong.
+					Logger.WriteError(e);
+				}
 				_waveIn = null;
 				RecordingState = RecordingState.Stopped;
 			}
@@ -149,7 +162,7 @@ namespace SIL.Media.Naudio
 		/// ------------------------------------------------------------------------------------
 		public virtual WaveFormat RecordingFormat
 		{
-			get { return _recordingFormat; }
+			get => _recordingFormat;
 			set
 			{
 				_recordingFormat = value;
@@ -158,12 +171,46 @@ namespace SIL.Media.Naudio
 		}
 
 		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Begins monitoring incoming data from the selected device but not
+		/// processing it for the purpose of recording to a file.
+		/// </summary>
+		/// <exception cref="InvalidOperationException">Called when in an invalid state. This
+		/// should only be called once for a new WaveIn device. Note that
+		/// <see cref="BeginRecording(string)"/> also begins monitoring, so if that is called
+		/// directly, this method should not be called.</exception>
+		/// <remarks>Other than the documented exceptions, any otherwise unhandled exception
+		/// thrown as a result of this call will be reported via
+		/// <see cref="ErrorReport.NotifyUserOfProblem(string,object[])"/>.
+		/// Methods in this class should not call this method. Instead use the private
+		/// internal version.</remarks>
+		/// ------------------------------------------------------------------------------------
 		public virtual void BeginMonitoring()
+		{
+			try
+			{
+				if (!BeginMonitoringIfNeeded())
+					throw new InvalidOperationException("only call this once for a new WaveIn device");
+			}
+			catch (Exception e)
+			{
+				ErrorReport.NotifyUserOfProblem(new ShowOncePerSessionBasedOnExactMessagePolicy(),
+					e, "There was a problem starting up volume monitoring.");
+			}
+		}
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>Begins monitoring if in the correct state to do so</summary>
+		/// <returns><c>true</c> if monitoring was started; <c>false</c> if not in a state where
+		/// monitoring could be started (e.g., already monitoring or recording, or stopped
+		/// with a waveIn that is still connected to the stopped/disconnected device).</returns>
+		/// ------------------------------------------------------------------------------------
+		private bool BeginMonitoringIfNeeded()
 		{
 			lock (this)
 			{
 				if (_waveIn != null || (_recordingState != RecordingState.NotYetStarted && _recordingState != RecordingState.Stopped))
-					throw new InvalidOperationException("only call this once for a new WaveIn device");
+					return false;
 				try
 				{
 					Debug.Assert(_waveIn == null);
@@ -180,18 +227,18 @@ namespace SIL.Media.Naudio
 					catch (MmException error)
 					{
 						if (error.Result != MmResult.AlreadyAllocated)
-							//TODO: I get this most of the time, but I don't know how to prevent it... maybe it's a hold over from previous runs? In which case, we need to make this disposable and stop the recording
+							//REVIEW: I get this most of the time, but I don't know how to prevent it... maybe it's a hold over from previous runs? In which case, we need to stop the recording
 							throw;
 					}
 
 					ConnectWithVolumeControl();
 					RecordingState = RecordingState.Monitoring;
+					return true;
 				}
-				catch (Exception e)
+				catch (Exception)
 				{
 					CloseWaveIn();
-					ErrorReport.NotifyUserOfProblem(new ShowOncePerSessionBasedOnExactMessagePolicy(), e,
-						 "There was a problem starting up volume monitoring.");
+					throw;
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1139)* status: all files reviewed, 1 unresolved discussion (waiting on @FoolRunning and @tombogle)

*[a discussion](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-My9AKX0Ev9kLOBpGSXx:-My9AKX0Ev9kLOBpGSXy:b-bxg5pz) (no related file):*
There's a lot happening in BeginMonitoring. I see what you're trying to do, but it's really hard to be sure no deadlock is possible. A whole lot of methods in this class lock(this) so if any of them gets called on another thread while we're trying to begin monitoring, and if whatever calls it needs to finish before we release the lock, we're in trouble.
Another thought is that it would be a good thing (even apart from this change) if BeginMonitoring released the lock before reporting failure to the user. That will involve switching to the UI thread if we're not already on it, and it will also involve holding the lock until the user closes the dialog. These sort of sync locks should not be held for such long periods if it can possibly be avoided. In this case, we know the BeginMonitoring failed, we're not going to do anything else to protected data structures, so it should be safe to release the lock, then report the problem.
(For example: suppose the user looking at the dialog decides to quit the program. Some cleanup code comes trying to StopMonitoring. But the thread that was trying to StartMonitoring has the lock. Now the poor user can't quit!)
Releasing the lock in BeginMonitoring before doing the report doesn't help if a caller also has the lock. So if we want to support that, maybe the exception needs to be reported at a higher level? But then you're changing the API in a possibly breaking way...grrrr...
I wonder if it is safer, and sufficient, to move the lock before the test for _recordingState != RecordingState.Monitoring? Seems like any race condition would then be exactly parallel to the other thread getting in before the start of the method. For example, if two threads are both trying to StartRecording at more-or-less the same time, with the lock right at the start, one of them will BeginMonitoring and then move on to the recording state, and the other one will skip BeginMonitoring and throw when it finds that the initial state is neither NotYetStarted nor Monitoring.
If the lock comes after the BeginMonitoring test, the second thread is likely to end up waiting to BeginMonitoring, and then reporting that Recording is the wrong state to call BeginMonitoring. That's not as helpful as reporting that we're in the wrong state to try BeginRecording.
What if we make a new method, BeginMonitoringIfNecessary? This would have the guts of BeginMonitoring, but not the error reporting. Instead, it returns an enumeration:
- OK, if it started monitoring successfully, or things are already in the Monitoring state
- BadState, if things are not in either the monitoring state or a state where we can start monitoring
- Failed, if it tried to start and couldn't. (We'd want an out param to return the exception.)
The new BeginMonitoring would call this and then, outside the lock, throw or notify the user as at present.
Meanwhile, without a lock, BeginRecording calls the new method (unconditionally). Then if the result was Failed it reports just like the new BeginMonitoring. Then it gets the lock. At this point, if the result was BadState, OR if the state has subsequently been changed (e.g., some other thread got in and started recording), it reports we're in the wrong state to start recording.
If I'm thinking straight, this would
- definitely not introduce any new possibilities for deadlock
- remove some existing possibilities for deadlock
- report any state problems in BeginRecording just as reliably as putting the lock around everything.
What you have is probably OK. Presumably, BeginMonitoring will seldom fail, and I can't see any clear path to a deadlock. But I do feel a bit nervous about approving it. I'd be happier with something like this suggestion. But feel free to seek other opinions...I don't consider myself expert on threading.

---

JohnThomson:
It would certainly make sense for BeginMonitoringIfNecessary to just rethrow any exception after calling CloseWaveIn. Just means both callers have to do a suitable try/catch and decide whether to NotifyUserOfProblem or rethrow (if it was InvalidOperationException).

SIL.Media/Naudio/AudioRecorder.cs
```diff
@@ -279,44 +326,82 @@ public virtual void BeginRecording(string waveFileName)
...
-				_recordingStartTime = DateTime.Now;
-				_prevRecordedTime = 0d;
-				RecordingState = RecordingState.Recording;
+				if (RecordingStarted != null)
+					RecordingStarted(this, EventArgs.Empty);
+			}
+			catch (InvalidRecordingStateException e)
+			{
+				var error = new InvalidOperationException(
+					$"Can't begin recording while we are in this state: {_recordingState}", e);
```
JohnThomson, line 396: Technically it would be better to capture _recordingState in a local variable inside the lock (or as part of the InvalidRecordingStateException). A race condition could conceivably change _recordingState between when you threw the InvalidRecordingStateException and when you now try to report the reason for it, resulting in a nonsensical message.
It would also be OK for this first catch to be inside the lock. It's the NotifyUserOfProblem that wants to be outside it.

SIL.Media/Naudio/AudioRecorder.cs
```diff
@@ -279,44 +326,82 @@ public virtual void BeginRecording(string waveFileName)
...
+				// while another tries to start recording). If they are both trying to start
+				// recording, we will end up throwing an InvalidOperationException on one of
+				// them, but we don't care which one. If one is merely trying to start
+				// monitoring and it happens to get there first, the other will find it is no
+				// longer in a valid state to begin monitoring but will happily find that it
+				// is now in a valid state to begin recording. If the one that is trying to
+				// begin recording gets there first, the other one will get the exception.
+				// In any case, handling an InvalidOperationException is the caller's
+				// responsibility.
+				BeginMonitoringIfNeeded();
```
JohnThomson, line 350: The comment above is very helpful. But it still worried me that this code is ignoring a return result that is generally used to report a problem. What you've done is valid...the false return does not necessarily mean a problem, just that monitoring was not begun. But it might be worth adding, "The return result is intentionally ignored. It doesn't matter whether BeginMonitoringIfNeeded started monitoring, found we were already monitoring, or found us in some other state. What matters is the test (below, inside the lock) that we end up in the right state to start recording."

JohnThomson (approved):
A few minor suggestions, but basically looks good. Feel free to ignore them if not helpful.

> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-:-MySIhB77eDPbzEAcqUg:bcq6otd)* status: all files reviewed, 3 unresolved discussions (waiting on @FoolRunning and @tombogle)

___
*[SIL.Media/Naudio/AudioRecorder.cs, line 379 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MyP6PQu8TC5pA_LsCw9:-MyP6PQu8TC5pA_LsCwA:bb9blws) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media/Naudio/AudioRecorder.cs#L379)):*
> ```Smalltalk
> 					if (_waveInBuffersChanged)
> 					{
> 						CloseWaveIn();
> ```

There seems to be some inconsistency in commenting (at least) in that the comments below indicate that CloseWaveIn() puts things in a state where BeginMonitoring can be called, while the comment on BeginMonitoring indicates that it should only ever be called once. (Of course here you're calling BeginMonitoringIfNeeded, but still...)
___
*[SIL.Media.Tests/AudioRecorderTests.cs, line 68 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MySE9AV3sIgO9WgxxNG:-MySE9AV3sIgO9WgxxNH:b6nmgn0) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media.Tests/AudioRecorderTests.cs#L68)):*
> ```Smalltalk
> 					}
> 
> 					Assert.That(recordingStopped);
> ```

I think there's a syntax for this, something like Assert.That(recordingStopped, Is.True.After(maxTime,interval). Not a necessary change, but an option to consider.
___
*[SIL.Media.Tests/AudioRecorderTests.cs, line 164 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MySHTpeDwsch7mXZ1g5:-MySHTpeDwsch7mXZ1g6:b7ipctw) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media.Tests/AudioRecorderTests.cs#L164)):*
> ```Smalltalk
> 					{
> 						Application.DoEvents();
> 						Thread.Sleep(200);
> ```

Polling more often might make this run faster. Any idea how long it really takes to BeginMonitoring?

tombogle:
___
*[SIL.Media/Naudio/AudioRecorder.cs, line 379 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MyP6PQu8TC5pA_LsCw9:-MySLse0EcYoUueYm0Ye:bh8erl6) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media/Naudio/AudioRecorder.cs#L379)):*

There seems to be some inconsistency in commenting (at least) in that the comments below indicate that CloseWaveIn() puts things in a state where BeginMonitoring can be called, while the comment on BeginMonitoring indicates that it should only ever be called once. (Of course here you're calling BeginMonitoringIfNeeded, but still...)
</blockquote></details>

The comment actually says "should only be called once for a new WaveIn device." CloseWaveIn disposes of the existing one, and then BeginMonitoring(IfNeeded) creates a new one. But I agree that that comment is still not quite right because it implies that some previous action would have already created the new WaveIn, but that is not the case. This is the one and only place where _waveIn gets set to something other than null.

tombogle:
___
*[SIL.Media/Naudio/AudioRecorder.cs, line 396 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-My84KPG804tgy3q_h8s-r2-396:-MySS6lZ4FZ4E-bC-eNX:bjmz7fa) ([raw file](https://github.com/sillsdev/libpalaso/blob/5038667d092ff08d78a7f5b9e45340ef1b37b80d/SIL.Media/Naudio/AudioRecorder.cs#L396)):*

Technically it would be better to capture _recordingState in a local variable inside the lock (or as part of the InvalidRecordingStateException). A race condition could conceivably change _recordingState between when you threw the InvalidRecordingStateException and when you now try to report the reason for it, resulting in a nonsensical message.
It would also be OK for this first catch to be inside the lock. It's the NotifyUserOfProblem that wants to be outside it.
</blockquote></details>

No longer relevant with the most recent changes.

tombogle:
___
*[SIL.Media.Tests/AudioRecorderTests.cs, line 164 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MySHTpeDwsch7mXZ1g5:-MySSIbfBjtJ1MQFbeU1:b-jesfcl) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media.Tests/AudioRecorderTests.cs#L164)):*

Polling more often might make this run faster. Any idea how long it really takes to BeginMonitoring?
</blockquote></details>

I started with the loop running up to 20 times and sleeping for a full second. Then I tweaked it down to 3 times @ 300 ms each. On my machine, it still passes at 2x100, but it actually seems to take longer. It fails at 2x50.  It might be able to be tightened up more. Looks like the real key is to sleep BEFORE the DoEvents call. On my machine, then I can bump it down to 90 ms and it only sleeps once.

tombogle:
___
*[SIL.Media.Tests/AudioRecorderTests.cs, line 68 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MySE9AV3sIgO9WgxxNG:-MySXbqu3fp5NXbR-5cA:bbzdul0) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media.Tests/AudioRecorderTests.cs#L68)):*

I think there's a syntax for this, something like Assert.That(recordingStopped, Is.True.After(maxTime,interval). Not a necessary change, but an option to consider.
</blockquote></details>

Cool. Great to know. Unfortunately, in the case the call to DoEvents is required, so that doesn't work here. If you can think of a way to restructure the test to not require that, I'd be happy to re-do it!

JohnThomson (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1139)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @FoolRunning)

___
*[SIL.Media.Tests/AudioRecorderTests.cs, line 164 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1139#-MySHTpeDwsch7mXZ1g5:-MyTPeph8nr5-wLY3o_F:b-5isyjh) ([raw file](https://github.com/sillsdev/libpalaso/blob/f1236feb1a6e763780356b528a05e6499f21b155/SIL.Media.Tests/AudioRecorderTests.cs#L164)):*

I started with the loop running up to 20 times and sleeping for a full second. Then I tweaked it down to 3 times @ 300 ms each. On my machine, it still passes at 2x100, but it actually seems to take longer. It fails at 2x50.  It might be able to be tightened up more. Looks like the real key is to sleep BEFORE the DoEvents call. On my machine, then I can bump it down to 90 ms and it only sleeps once.
</blockquote></details>

I was more wondering whether, depending on things like machine load and speed, you might want a higher count and shorter sleep. You don't want it to fail just because TeamCity is busy. But if it's not busy and manages to get monitoring started in 40ms, why wait 100? Then again, maybe the thread switching overhead and cost of DoEvents is significant. Sounds like you've done experiments and found a thoughtful compromise, so that's fine with me.

tombogle:
These tests can't run on TC anyway because they require a speaker and a mic. But, yeah, I think I've found the sweet spot. Anything less falls beneath the overhead of thread switching and doesn't consistently allow the other thread enough time to progress.

PR #1141, opened 2022-03-15
Merge branch libpalaso-6.0 into master

marksvc:
This brings in some changes for flatpak, such as to keyboarding,
platform utilities, and mutex.

+semver:feature

===
I get confused about how to propose branch merges in github pull requests.


.vscode/launch.json
```diff
@@ -16,11 +16,24 @@
...
+		},
+		{
+			"name": "Launch TestAppKeyboard.exe",
+			"env": {
+			  // Use this mono.
+			  // "PATH": "/opt/mono5-sil/bin:${env:PATH}",
+			},
+			"type": "mono",
+			"request": "launch",
+			"program": "${workspaceRoot}/output/DebugMonoStrongName/TestAppKeyboard.exe",
```
ermshiperete, line 35: On `master` the path is different. I think something like this should work:

```suggestion
			"program": "${workspaceRoot}/output/Debug/net461/TestAppKeyboard.exe",
```

SIL.Windows.Forms.Keyboarding/Linux/UnityIbusKeyboardRetrievingAdaptor.cs
```diff
@@ -62,6 +63,7 @@ private void RegisterKeyboards(IDictionary<string, uint> keyboards, (string, str
 					{
 						keyboard.SetIsAvailable(true);
 						keyboard.IBusKeyboardEngine = ibusKeyboard;
+							Logger.WriteEvent($"{System.Reflection.MethodBase.GetCurrentMethod().DeclaringType.Name}: Set keyboard as available: {keyboard}");
```
ermshiperete, line 66: I wonder if  using `GetType().Name` would be easier and/or more efficient?

ermshiperete (commented):
Looks mostly good, only a few comments.

tombogle:
This PR is really old and seems to have gone stale. Is there something here that we still need, or should this be closed?

marksvc:
The broader work associated with this got stalled from some difficulties/delays in communication. I think I need to learn more to understand how to proceed. 

I'll close this PR for now, but not delete the branch.

PR #1145, opened 2022-03-23
Upgrading SIL.Core.Desktop, SIL.Lift, and SIL.Linux.Logging to .NET Standard.

josephmyers:
I didn't dual-target SIL.Linux.Logging to Framework 4.6.1 (as appears to be the pattern), because the .NET Standard version of SIL.Core contains Mono.Posix.NetStandard, which required me to remove the Mono.Posix reference here. If this is desired, I can spend more time on it. **Edit: 4.6.1 support added back.**

I also added a preprocessor directive to Utilities.cs for Framework only, since .NET Standard doesn't support the .TemporaryFiles property.

For more information, see:
https://docs.microsoft.com/en-us/dotnet/api/system.xml.xsl.xslcompiledtransform.temporaryfiles?view=netframework-4.6.1
https://github.com/sillsdev/lift-standard/blame/4c3cd07950b58615549fd6005426304942d0b1f1/LIFTDotNet/LiftIO/Utilities.cs#L54

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1145)

### DIFF
--- a/SIL.Core.Desktop/IO/DirectoryUtilities.cs
+++ b/SIL.Core.Desktop/IO/DirectoryUtilities.cs
@@ -126,8 +126,8 @@ namespace SIL.IO
 
 			// get current settings
 			var everyone = new SecurityIdentifier(WellKnownSidType.WorldSid, null);
-			var fileInfo = new FileInfo(fullDirectoryPath);
-			var security = fileInfo.GetAccessControl(AccessControlSections.Access);
+			var directory = new DirectoryInfo(fullDirectoryPath);
+			var security = directory.GetAccessControl(AccessControlSections.Access);
 			var currentRules = security.GetAccessRules(true, true, typeof(SecurityIdentifier));
 
 			// if everyone already has full control, return now
@@ -150,7 +150,7 @@ namespace SIL.IO
 															PropagationFlags.None,
 															AccessControlType.Allow);
 				security.AddAccessRule(fullControl);
-				fileInfo.SetAccessControl(security);
+				directory.SetAccessControl(security);
 
 				returnVal = true;
 			}
--- a/SIL.Core.Desktop/IO/DirectoryUtilities.cs
+++ b/SIL.Core.Desktop/IO/DirectoryUtilities.cs
@@ -126,7 +126,8 @@ namespace SIL.IO
 
 			// get current settings
 			var everyone = new SecurityIdentifier(WellKnownSidType.WorldSid, null);
-			var security = Directory.GetAccessControl(fullDirectoryPath, AccessControlSections.Access);
+			var fileInfo = new FileInfo(fullDirectoryPath);
+			var security = fileInfo.GetAccessControl(AccessControlSections.Access);
 			var currentRules = security.GetAccessRules(true, true, typeof(SecurityIdentifier));
 
 			// if everyone already has full control, return now
@@ -149,7 +150,7 @@ namespace SIL.IO
 															PropagationFlags.None,
 															AccessControlType.Allow);
 				security.AddAccessRule(fullControl);
-				Directory.SetAccessControl(fullDirectoryPath, security);
+				fileInfo.SetAccessControl(security);
 
 				returnVal = true;
 			}
--- a/SIL.Core.Desktop/Progress/Commands/AsyncCommand.cs
+++ b/SIL.Core.Desktop/Progress/Commands/AsyncCommand.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using System.Threading.Tasks;
 
 //For UML diagram, see ProgressSystem.uml (done in StarUML)
 
@@ -146,9 +147,9 @@ namespace SIL.Progress.Commands
 		/// Override this method to invoke the actual
 		/// long operation using your preferred async mode.
 		/// </summary>
-		protected abstract void BeginInvokeCore();
+		protected abstract Task BeginInvokeCore();
 
-		protected abstract void BeginInvokeCore2(ProgressState progress);
+		protected abstract Task BeginInvokeCore2(ProgressState progress);
 
 		/// <summary>
 		/// Raises the Finish event.
--- a/SIL.Core.Desktop/Progress/Commands/BasicCommand.cs
+++ b/SIL.Core.Desktop/Progress/Commands/BasicCommand.cs
@@ -58,14 +58,14 @@ namespace SIL.Progress.Commands
 				_progressCallback,
 				_primaryStatusTextCallback,
 				_secondaryStatusTextCallback));
-			await EndWork(workTask).ConfigureAwait(false);
+			await EndWork(workTask).ConfigureAwait(false); //resume execution on any thread
 		}
 
 		protected override async Task BeginInvokeCore2(ProgressState progress)
 		{
 			WorkInvoker2 worker = DoWork2;
 			var workTask = Task.Run(() => worker.Invoke(progress));
-			await EndWork(workTask).ConfigureAwait(false);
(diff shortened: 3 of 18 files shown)


josephmyers:
The build is failing, because it's not doing a package restore beforehand?

ermshiperete:
> The build is failing, because it's not doing a package restore beforehand?

The problem is the line `<TargetFramework>netstandard2.0</TargetFramework>` because this causes both `TargetFramework` and `TargetFrameworks` to be set, and depending on the way of building one or the other gets used.

josephmyers:
> > The build is failing, because it's not doing a package restore beforehand?
> 
> The problem is the line `<TargetFramework>netstandard2.0</TargetFramework>` because this causes both `TargetFramework` and `TargetFrameworks` to be set, and depending on the way of building one or the other gets used.

Ok, thanks, I'm not sure why it builds locally. I will use TargetFrameworks instead.

SIL.Linux.Logging/SIL.Linux.Logging.csproj
```diff
@@ -4,12 +4,12 @@
     <RootNamespace>SIL.Linux.Logging</RootNamespace>
     <AssemblyTitle>SIL.Linux.Logging</AssemblyTitle>
     <Description>SIL.Linux.Logging provides a library to log to the syslog service.</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="GitVersion.MsBuild" Version="5.6.8" PrivateAssets="all" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
-    <PackageReference Include="Mono.Posix" Version="5.4.0.201" Condition="'$(OS)' == 'Windows_NT'" />
```
ermshiperete, line 12: You should also remove the `Mono.Posix` reference on Linux a few lines below.

I'm not sure where this assembly is used, but for compatibility it might be better to additionally compile SIL.Linux.Logging for `net461` so that we don't introduce an possibly incompatible different `Mono.Posix` dependency.

SIL.Lift/Utilities.cs
```diff
@@ -63,11 +63,13 @@ static public string ProcessLiftForLaterMerging(string inputPath)
 			{
 				transform.Transform(outputOfPassOne, new XsltArgumentList(), output);
 			}
+			#if NET461
```
ermshiperete, line 66: Can you add a comment why this is not needed with netstandard 2?

josephmyers: I had to add this, because in .NET Standard there is no .TemporaryFiles property. See the commit messages and the top of the PR page for more info!

SIL.Linux.Logging/SIL.Linux.Logging.csproj
josephmyers: Ok, thanks, looking into this now

SIL.Lift/Utilities.cs
josephmyers: To be clear, I can't guarantee that the .NET Standard version doesn't use temporary files. I only know that that property isn't supported. I am definitely open to suggestions

SIL.Core.Desktop/Progress/Commands/AsyncCommand.cs
```diff
@@ -146,9 +147,9 @@ public bool Canceling
 		/// Override this method to invoke the actual
 		/// long operation using your preferred async mode.
 		/// </summary>
-		protected abstract void BeginInvokeCore();
+		protected abstract Task BeginInvokeCore();
```
ermshiperete, line 150: Shouldn't this have `async`?

```suggestion
		protected abstract async Task BeginInvokeCore();
```

SIL.Core.Desktop/Progress/Commands/AsyncCommand.cs
```diff
@@ -146,9 +147,9 @@ public bool Canceling
...
 
-		protected abstract void BeginInvokeCore2(ProgressState progress);
+		protected abstract Task BeginInvokeCore2(ProgressState progress);
```
ermshiperete, line 152:
```suggestion
		protected abstract async Task BeginInvokeCore2(ProgressState progress);
```

SIL.Core.Desktop/Progress/Commands/BasicCommand.cs
```diff
@@ -51,24 +50,22 @@ public StatusCallback SecondaryStatusTextCallback
...
 		}
 
-		protected override void BeginInvokeCore2(ProgressState progress)
+		protected override async Task BeginInvokeCore2(ProgressState progress)
 		{
-			WorkInvoker2 worker = new WorkInvoker2(DoWork2);
-			worker.BeginInvoke(progress, new AsyncCallback(EndWork2), null);
+			WorkInvoker2 worker = DoWork2;
+			var workTask = Task.Run(() => worker.Invoke(progress));
+			await EndWork(workTask).ConfigureAwait(false);
```
ermshiperete, line 68: Can you add a comment why `ConfigureAwait` is necessary? I notice you have that information in the commit message, but IMO it would be good to have it in the code as well.

ermshiperete (changes requested):
Since this changes API it should have `+semver:major` in one of the commit messages.

Also,  `CHANGELOG.md` should mention that support for .NET Standard was added, and that USBDrive is only supported on .NET Framework.

SIL.Core.Desktop/Progress/Commands/BasicCommand.cs
josephmyers: It's not necessary, but it makes the code more compatible. By default, an await captures the context of the method and requires that context to resume once the Task finishes. This _can_ cause deadlock issues in some situations, but .ConfigureAwait(false) indicates that the resumption can occur on any thread. This can apply here, since no information from the Task is required to resume the method.

josephmyers: If you want me to put that in the code, I can, it's just kind of a brittle comment

SIL.Core.Desktop/IO/DirectoryUtilities.cs
```diff
@@ -126,7 +126,8 @@ public static bool SetFullControl(string fullDirectoryPath, bool showErrorMessag
 
 			// get current settings
 			var everyone = new SecurityIdentifier(WellKnownSidType.WorldSid, null);
-			var security = Directory.GetAccessControl(fullDirectoryPath, AccessControlSections.Access);
+			var fileInfo = new FileInfo(fullDirectoryPath);
```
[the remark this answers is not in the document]
josephmyers: This change is necessary, because Directory.GetAccessControl (etc) is not supported in .NET Standard; it's available in FileInfo/DirectoryInfo.

SIL.Core.Desktop/Progress/Commands/AsyncCommand.cs
josephmyers: It only needs async if there's an await command within. Since this abstract, there are no await's. And async doesn't modify the method signature, so implementing methods are free to choose whether or not they want to go asynchronous. Additionally, any method returning Task can be awaited -- no async keyword required :)

ermshiperete (approved):
> Reviewed 2 of 3 files at r1, 13 of 15 files at r3, 1 of 2 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1145)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #1147, opened 2022-03-23
Added option for alternate image metadata UI (BL-6259)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1147)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.WritingSystems] Allow specifying an alias to another Writing System for changing between upper- and lowercase
 - [SIL.Core] Extension method to get longest useful substring
 - [SIL.Core] Extension method IsLikelyWordForming to include letters, format characters, PUA and marks (diacritics, etc.)
+- [SIL.Windows.Forms] Caller can override the default save image metadata action from the image toolbox
 
 ### Changed
 
@@ -109,7 +110,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.DblBundle.Tests] Create nuget package
 - [SIL.Windows.Forms] Use the new Registered trademark logo (in About Box). Remove 132x148 logo.
 - [SIL.Windows.Forms] Use [Markdig](https://github.com/lunet-io/markdig) instead of
-    [MarkdownDeep.NET](https://www.toptensoftware.com/markdowndeep/)
+  [MarkdownDeep.NET](https://www.toptensoftware.com/markdowndeep/)
 
 ### Fixed
 
@@ -141,7 +142,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   on both Windows and Linux (although the limitations what works and what doesn't work remain the
   same)
 - [SIL.Media] cleanup of `AudioSession` API: rename `AudioIrrKlangSession` to `WindowsAudioSession`.
-  `AudioAlsaSession` and `WindowsAudioSession` are now  internal (they were never intended to
+  `AudioAlsaSession` and `WindowsAudioSession` are now internal (they were never intended to
   be used directly)
 - [SIL.Media] move some interfaces around so that they live in `SIL.Media` instead of
   `SIL.Media.Naudio`: `IAudioRecorder`, `RecordingState`, `IAudioPlayer`
--- a/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs
+++ b/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs
@@ -340,6 +340,14 @@ namespace SIL.Windows.Forms.ImageToolbox
 
 		}
 
+		/// <summary>
+		/// If set, this action will be used instead of the default (launching <see cref="MetadataEditorDialog"/>).
+		/// For example, the client may want to use a different UI to edit the `Metadata`.
+		/// The `Action<Metadata>` callback saves the modified `Metadata` to the image.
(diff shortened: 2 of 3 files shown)


tombogle:
___
*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 365 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1147#-MysOqRqBpVdBVcEegwB:-MysOqRqBpVdBVcEegwC:b-ynvo5v) ([raw file](https://github.com/sillsdev/libpalaso/blob/241681b7f7c398ef4a9a3260359a296b542efe9e/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L365)):*
> ```Smalltalk
> 			{
> 				EditMetadataActionOverride(_imageInfo.Metadata, SetNewImageMetadata);
> 				return;
> ```

Did you consider the possible merits of an event handler instead, with either an event args having a bool (Handled) that could be set or a bool return type indicating whether or not it was handled? This would allow the consumer to do something with the metadata (perhaps adjust it in some way) but still have it feed into the normal edit dialog. Depending on the current state of the metadata, it might be that a runtime decision should be made as to whether the custom editor would be required.

tombogle:
___
*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 349 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1147#-MysPzu12wf624442W5Q:-MysPzu12wf624442W5R:bvoarpd) ([raw file](https://github.com/sillsdev/libpalaso/blob/241681b7f7c398ef4a9a3260359a296b542efe9e/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L349)):*
> ```Smalltalk
> 		/// <see cref="SetNewImageMetadata(Metadata)"/>
> 		/// </summary>
> 		public Action<Metadata, Action<Metadata>> EditMetadataActionOverride { get; set; }
> ```

In both cases, the same action is used to save the metadata. Would it be better to have the override merely return the updated metadata and let this method handle setting it? Though I suppose then, you would need to return a value indicating whether the edit had been cancelled. Of course, if a mutable copy of the metadata were passed in, then this still might be a cleaner interface.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1147)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @tombogle)

___
*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 349 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1147#-MysPzu12wf624442W5Q:-MysRLKk730h2IuBfTsd:b-8y4h5j) ([raw file](https://github.com/sillsdev/libpalaso/blob/241681b7f7c398ef4a9a3260359a296b542efe9e/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L349)):*

In both cases, the same action is used to save the metadata. Would it be better to have the override merely return the updated metadata and let this method handle setting it? Though I suppose then, you would need to return a value indicating whether the edit had been cancelled. Of course, if a mutable copy of the metadata were passed in, then this still might be a cleaner interface.
</blockquote></details>

In our case, the "save" could be thought of as asynchronous. Basically, the new UI gets launched; then later when the user submits the new metadata, the save action is called.
Does that clear up the reason why this is needed? Perhaps I am the one not understanding.

tombogle:
___
*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 349 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1147#-MysPzu12wf624442W5Q:-MysSG4j-oYPSpMuQ5kY:bxxx4gf) ([raw file](https://github.com/sillsdev/libpalaso/blob/241681b7f7c398ef4a9a3260359a296b542efe9e/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L349)):*

In our case, the "save" could be thought of as asynchronous. Basically, the new UI gets launched; then later when the user submits the new metadata, the save action is called.
Does that clear up the reason why this is needed? Perhaps I am the one not understanding.
</blockquote></details>

That makes sense.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1147)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @tombogle)

___
*[SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs, line 365 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1147#-MysOqRqBpVdBVcEegwB:-MysUaGm6_MFk31Ax6mt:bxemv82) ([raw file](https://github.com/sillsdev/libpalaso/blob/241681b7f7c398ef4a9a3260359a296b542efe9e/SIL.Windows.Forms/ImageToolbox/ImageToolboxControl.cs#L365)):*

Did you consider the possible merits of an event handler instead, with either an event args having a bool (Handled) that could be set or a bool return type indicating whether or not it was handled? This would allow the consumer to do something with the metadata (perhaps adjust it in some way) but still have it feed into the normal edit dialog. Depending on the current state of the metadata, it might be that a runtime decision should be made as to whether the custom editor would be required. 
</blockquote></details>

Does understand the asynchronous nature of our use case clear this up, too? If not, we should discuss because I don't think I'm seeing it.

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1147)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #1151, opened 2022-03-28
Fix l10n.proj

ermshiperete:
- use GitVersion.MsBuild instead of deprecated GitVersionTask
- use TargetFramework**s** (plural) because that's defined in the parent Directory.Build.props
- 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1151)


tombogle (approved):

tombogle (changes requested):
Still doesn't solve the problem

ermshiperete:
The problem was that `<TargetFrameworks>` is already defined in `Directory.Build.props` and confused things.

tombogle (approved):

PR #1154, opened 2022-04-06
LT-21002: Prevent crash by filtering the sort list

mark-sil:
Filter the sort language list so that it only contains languages
that are installed on the machine.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1154)

### DIFF
--- a/SIL.Windows.Forms.WritingSystems/WritingSystemSetupModel.cs
+++ b/SIL.Windows.Forms.WritingSystems/WritingSystemSetupModel.cs
@@ -545,6 +545,11 @@ namespace SIL.Windows.Forms.WritingSystems
 				{
 					if (prohibitedList.Contains(cultureInfo.Name, StringComparison.OrdinalIgnoreCase))
 						continue;
+
+					// Do not add languages that are not installed.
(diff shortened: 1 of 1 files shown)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1154)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor)

*[a discussion](https://reviewable.io/reviews/sillsdev/libpalaso/1154#-N---hqZB7hY1DApDLe4:-N---hqZB7hY1DApDLe5:bx0g86t) (no related file):*
Hi Jason, after some initial investigation I would like to implement this solution instead of trying to create IcuRulesCollationDefinition’s for languages that are not installed. Using IcuRulesCollationDefinition’s is a much more complex solution than I anticipated so I would like to avoid the work unless we know it is something users would use.

---

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1154)* status: all files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor)

PR #1155, opened 2022-04-07
Another attempt to fix l10n.proj

ermshiperete:
During a regular build (of a .csproj) the build system sets `TargetFramework` for each framework in `TargetFrameworks`. This causes the `PkgL10NSharp_ExtractXliff` to get defined (in `obj\l10n.proj.nuget.g.props`). However, if `TargetFrameworks` is set,
the definition of the property gets an additional condition on `TargetFramework` so that the property isn't set when `UpdateCrowdin` runs.

This change sets `TargetFramework` (and sets `TargetFrameworks` to the empty string) to work around this problem.

See also https://github.com/NuGet/Home/issues/11721 and https://github.com/NuGet/Home/issues/11725

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1155)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms.ClearShare.WinFormsUI] Default to CC-BY for new CC licenses
 - [SIL.Media] Allow RecordingDeviceIndicator to find new sound input device when there was no selected device (state == NotYetStarted)
 - [SIL.Windows.Forms] Internationalized the ExceptionReportingDialog.
+- [SIL.Windows.Forms] Corrected typo in the name of AcquireImageControl.SetInitialSearchString
 
 ### Fixed
 
--- a/DistFiles/Palaso.en.xlf
+++ b/DistFiles/Palaso.en.xlf
@@ -518,6 +518,10 @@
         <note>ID: ImageToolbox.PictureFiles</note>
         <note>Shown in the file-picking dialog to describe what kind of files the dialog is filtering for</note>
       </trans-unit>
+      <trans-unit id="ImageToolbox.ProblemGettingImageFromDevice">
+        <source xml:lang="en">Problem Getting Image</source>
+        <note>ID: ImageToolbox.ProblemGettingImageFromDevice</note>
+      </trans-unit>
       <trans-unit id="ImageToolbox.ProblemLoadingImage">
         <source xml:lang="en">Sorry, there was a problem loading that image.</source>
         <note>ID: ImageToolbox.ProblemLoadingImage</note>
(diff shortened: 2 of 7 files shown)


tombogle:
This seems to work! It still require the build to be run twice because it fails the first time if the generated file is missing or out-of-date. I assume there might be some way to do the "RestartBuild" trick that we do in some of the other .proj files, but I guess the condition would have to be if the GetVersion target is not defined, and I'm not sure how to do that (if it's even possible). And then if something else goes wrong, I suppose that could lead to infinite recursion.

tombogle:
*[l10n/l10n.proj, line 13 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1155#-N-4bZSP3Tds1pcZ5Cj_:-N-4bZSP3Tds1pcZ5Cja:b-uhd85f) ([raw file](https://github.com/sillsdev/libpalaso/blob/600090b5618c02c33d7e445367656bcfb24052eb/l10n/l10n.proj#L13)):*
> ```XML
> 	<PackageReference Include="GitVersion.MsBuild" Version="5.9.0" GeneratePathProperty="true" />
> 	<PackageReference Include="L10NSharp.ExtractXliff" Version="4.1.0-beta0036" GeneratePathProperty="true" />
>     <PackageReference Include="NuGet.CommandLine" Version="5.4.0" GeneratePathProperty="true" />
> ```

Not really related to this fix, but what do you think of maybe going ahead and bumping this up to 6.1.0 just so as not to lag behind needlessly. I tried it and it works.

ermshiperete:
> _[l10n/l10n.proj, line 13 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1155#-N-4bZSP3Tds1pcZ5Cj_:-N-4bZSP3Tds1pcZ5Cja:b-uhd85f) ([raw file](https://github.com/sillsdev/libpalaso/blob/600090b5618c02c33d7e445367656bcfb24052eb/l10n/l10n.proj#L13)):_
> 
> > ```
> > 	<PackageReference Include="GitVersion.MsBuild" Version="5.9.0" GeneratePathProperty="true" />
> > 	<PackageReference Include="L10NSharp.ExtractXliff" Version="4.1.0-beta0036" GeneratePathProperty="true" />
> >     <PackageReference Include="NuGet.CommandLine" Version="5.4.0" GeneratePathProperty="true" />
> > ```
> 
> Not really related to this fix, but what do you think of maybe going ahead and bumping this up to 6.1.0 just so as not to lag behind needlessly. I tried it and it works.

Sure, good idea.

ermshiperete:
> I assume there might be some way to do the "RestartBuild" trick that we do in some of the other .proj files, but I guess the condition would have to be if the GetVersion target is not defined, and I'm not sure how to do that (if it's even possible).

We could restart if PkgL10NSharp_ExtractXliff is empty. The way to prevent infinite recursion is to have two targets. One depends on Restore and then restarts if necessary, passing the second target to msbuild. If we don't have to restart we call the second target directly. The second target does the actual work.

tombogle (approved):

tombogle:
*[l10n/l10n.proj, line 13 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1155#-N-4bZSP3Tds1pcZ5Cj_:-N-9R1xg68YZftYOtEl9:bb6un1j) ([raw file](https://github.com/sillsdev/libpalaso/blob/600090b5618c02c33d7e445367656bcfb24052eb/l10n/l10n.proj#L13)):*

Sure, good idea.
</blockquote></details>

Done

jasonleenaylor (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1155)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

PR #1159, opened 2022-04-22
PTX-21397 revised pasting verse ref so it works when CTRL-V is used as a shortcut key

jwickberg:
Previous changes didn't work in Paratext because a menu item had CTRL-V as a shortcut key and that would be processed before the KeyDown event happened. Changed code to use ProcessCmdKey instead of KeyDown.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1159)


SIL.Core/Linq/MiscOperators.cs
```diff
@@ -42,5 +42,29 @@ public static void ForEach<TSource> (this IEnumerable<TSource> source, Action<TS
 			foreach (TSource element in source)
 				action (element);
 		}
+
```
tombogle, line 45: GlyssenEngine.Utilities.LinqExtensions has a similar method called OnlyOrDefault, except my version of it does not take a predicate. However, some of the places I use it could benefit from the predicate version. (I just use a Where to pre-process the enumeration.) I think we should have a consistent name and put them in the same place (here). I think I have a slight preference for my name since it is more similar to SingleOrDefault and is clear about what happens when there is more than one. If you agree, I'll do my part o move mine over. I'm not 100% sure if it makes sense to keep the implementations distinct or if it would be better to have a single underlying implementation.

papeh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1159)* status: all files reviewed, 4 unresolved discussions (waiting on @FoolRunning and @jwickberg)

___
*[SIL.Core/Linq/MiscOperators.cs, line 47 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1159#-N0Hs2ylCZFVP_56xAN_:-N0Hs2ylCZFVP_56xANa:b1qbtsk) ([raw file](https://github.com/sillsdev/libpalaso/blob/f739d3e04c88675367c184870bd84b6e15cefae5/SIL.Core/Linq/MiscOperators.cs#L47)):*
> ```Smalltalk
> 
> 		// Like SingeOrDefault, but doesn't throw exception if more than one match is found
> 		// Returns found item or null if item isn't found or there is more than one match
> ```

XML comments would generate documentation; e.g.
```
/// <summary>Like SingleOrDefault</summary>
/// <returns>The one item or null</returns>
```
___
*[SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs, line 102 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1159#-N0HtLFA6vvOUpNcHH3g:-N0HtLFA6vvOUpNcHH3h:bj40y91) ([raw file](https://github.com/sillsdev/libpalaso/blob/f739d3e04c88675367c184870bd84b6e15cefae5/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs#L102)):*
> ```Smalltalk
> 		}
> 
> 		private void OnClick(object sender, EventArgs e)
> ```

I'd prefer calling this `OnPaste` or converting it to a lambda function
___
*[SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs, line 112 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1159#-N0HtBi7BaMpKEtzeIW5:-N0HtBi7BaMpKEtzeIW6:b-uucrjp) ([raw file](https://github.com/sillsdev/libpalaso/blob/f739d3e04c88675367c184870bd84b6e15cefae5/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs#L112)):*
> ```Smalltalk
> 		internal static bool CallProcessCmdKey(this VerseControl verseControl)
> 		{
> 			Keys keysData = Keys.Control | Keys.V;
> ```

Including Ctrl+V in the method name would improve clarity

tombogle:
*[SIL.Windows.Forms.Scripture/VerseControl.cs, line 363 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1159#-N0Hu3AnBl6hY2EYQIM_:-N0Hu3AnBl6hY2EYQIMa:b-a8thkd) ([raw file](https://github.com/sillsdev/libpalaso/blob/f739d3e04c88675367c184870bd84b6e15cefae5/SIL.Windows.Forms.Scripture/VerseControl.cs#L363)):*
> ```Smalltalk
> 		/// <param name="msg"></param>
> 		/// <param name="keyData"></param>
> 		/// <returns>true if CTRL-V was used and there was a valid reference in the clipboard</returns>
> ```

Maybe this is okay, but it seems a bit strange that if the clipboard contains something that cannot be interpreted as a valid reference, it will potentially go on to be handled by the Paste menu item (which is presumably what has a Ctrl-V shortcut). If the Verse Control has focus, do you really want a Ctrl-V to sometimes be handled by another control or window?

SIL.Core/Linq/MiscOperators.cs
```diff
@@ -42,5 +42,29 @@ public static void ForEach<TSource> (this IEnumerable<TSource> source, Action<TS
...
+		/// <summary>Like SingeOrDefault, but doesn't throw exception if more than one match is found.</summary>
```
ermshiperete, line 46:
```suggestion
		/// <summary>Like SingleOrDefault, but doesn't throw exception if more than one match is found.</summary>
```

SIL.Core/Linq/MiscOperators.cs
```diff
@@ -42,5 +42,29 @@ public static void ForEach<TSource> (this IEnumerable<TSource> source, Action<TS
...
+		/// <returns>found item or null if item isn't found or there is more than one match</returns>
+		public static TSource OnlyOrDefault<TSource>(this IEnumerable<TSource> source,
+			Func<TSource, bool> predicate)
+		{
+			if (source == null) throw new ArgumentNullException(nameof(source));
+			if (predicate == null) throw new ArgumentNullException(nameof(predicate));
+			return source.Where(predicate).OnlyOrDefault();
+		}
+
+		/// <summary>Like SingeOrDefault, but doesn't throw exception if more than one match is found.</summary>
```
ermshiperete, line 56:
```suggestion
		/// <summary>Like SingleOrDefault, but doesn't throw exception if more than one match is found.</summary>
```

papeh (approved):
:lgtm:

> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1159#-:-N0X8qHj4bus2Vu-r__s:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @FoolRunning, @jwickberg, and @tombogle)

tombogle:
*[`SIL.Core/Linq/MiscOperators.cs` line 45 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1159#-N0HolKbXjaSuupvD541-r1-45:-N0XC0qr27SbBT7DOiZG:btojbfd) ([raw file](https://github.com/sillsdev/libpalaso/blob/f739d3e04c88675367c184870bd84b6e15cefae5/SIL.Core/Linq/MiscOperators.cs#L45)):*
<details><summary><i>Previously, jwickberg wrote…</i></summary><blockquote>

I'll rename the method and make it have two versions - one with a predicate and one without - the predicate version will just be a convenience method for the other.
</blockquote></details>

Great!

PR #1160, opened 2022-04-22
Update IErrorReporter interface

sujeffreyl:
Normally goes through a simpler method which gives the IErrorReporter more control over the flow. The {resultIfAlternateButtonPressed} flow is only used if explicitly requested.

### DIFF
--- a/SIL.Core/Reporting/ErrorReport.cs
+++ b/SIL.Core/Reporting/ErrorReport.cs
@@ -17,10 +17,26 @@ namespace SIL.Reporting
 	public interface IErrorReporter
 	{
 		void ReportFatalException(Exception e);
+
+		/// <summary>
+		/// Notify the user of the {exception} and {message}, if {policy} permits, using the IErrorReporter's default options 
+		/// </summary>
+		/// <remarks>It is not required for the implementation to block (if the caller is not on the UI thread)
+		/// I would suggest that it should not block (if the caller is not on the UI thread)
+		/// By not blocking, you reduce headache and worry about the application deadlocking.
+		/// This will allow the caller to do an easy "fire and forget" of the problem notification mechanism</remarks>
+		void NotifyUserOfProblem(IRepeatNoticePolicy policy, Exception exception, string message);
+
+		/// <summary>
+		/// Notify the user of the {message}, if {policy} permits. Customize the {alternate button} label. 
+		/// </summary>
+		/// <remarks>The method should block and wait for the user to press the required button</remarks>
+		/// <returns>The method should return {resultIfAlternateButtonPressed} if the alternate button is clicked, ErrorResult.OK otherwise</returns>
 		ErrorResult NotifyUserOfProblem(IRepeatNoticePolicy policy,
 										string alternateButton1Label,
 										ErrorResult resultIfAlternateButtonPressed,
 										string message);
+
 		void ReportNonFatalException(Exception exception, IRepeatNoticePolicy policy);
 		void ReportNonFatalExceptionWithMessage(Exception error, string message, params object[] args);
 		void ReportNonFatalMessageWithStackTrace(string message, params object[] args);
@@ -473,23 +489,27 @@ namespace SIL.Reporting
 
 		public static ErrorResult NotifyUserOfProblem(IRepeatNoticePolicy policy, string messageFmt, params object[] args)
 		{
-			return NotifyUserOfProblem(policy, null, default(ErrorResult), messageFmt, args);
+			NotifyUserOfProblem(policy, null, messageFmt, args);
+			return ErrorResult.OK;
 		}
 
-		public static void NotifyUserOfProblem(Exception error, string messageFmt, params object[] args)
+		public static void NotifyUserOfProblem(Exception exception, string messageFmt, params object[] args)
 		{
-			NotifyUserOfProblem(new ShowAlwaysPolicy(), error, messageFmt, args);
+			NotifyUserOfProblem(new ShowAlwaysPolicy(), exception, messageFmt, args);
 		}
-
-		public static void NotifyUserOfProblem(IRepeatNoticePolicy policy, Exception error, string messageFmt, params object[] args)
+				
+		public static void NotifyUserOfProblem(IRepeatNoticePolicy policy, Exception exception, string messageFmt, params object[] args)
 		{
-			var result = NotifyUserOfProblem(policy, "Details", ErrorResult.Yes, messageFmt, args);
-			if (result == ErrorResult.Yes)
+			var message = string.Format(messageFmt, args);
+			if (s_justRecordNonFatalMessagesForTesting)
 			{
-				OnShowDetails(error, string.Format(messageFmt, args));
+				s_previousNonFatalMessage = message;
+				return;
 			}
 
-			UsageReporter.ReportException(false, null, error, String.Format(messageFmt, args));
+			_errorReporter.NotifyUserOfProblem(policy, exception, message);
+
+			UsageReporter.ReportException(false, null, exception, String.Format(messageFmt, args));
 		}
 
 		public static ErrorResult NotifyUserOfProblem(IRepeatNoticePolicy policy,
--- a/SIL.Core/Reporting/ConsoleErrorReporter.cs
+++ b/SIL.Core/Reporting/ConsoleErrorReporter.cs
@@ -16,22 +16,37 @@ namespace SIL.Reporting
 			WriteExceptionToConsole(error, null, Severity.Fatal);
 		}
 
-		public ErrorResult NotifyUserOfProblem(IRepeatNoticePolicy policy, string alternateButton1Label,
-			ErrorResult resultIfAlternateButtonPressed, string message)
+		/// <summary>
+		/// Notifies the user of problem by writing to console
+		/// </summary>
+		/// <param name="policy">The policy used to check if the message should be shown</param>
+		/// <param name="error">N/A. You may pass null. This parameter will be ignored.</param>
+		/// <param name="message">The message to print to console, if the policy allows</param>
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
Needs a changelog update -- should probably be explicit about the behavior changes and what devs would do about it.
Needs semver:major (correct me if I'm wrong) in the commit message

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-:-N0IY5LB-94iwmhOKydD:blw3cu7)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[SIL.Windows.Forms/Reporting/WinFormsErrorReporter.cs, line 40 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0IaPQU4Fodnpk04IVW:-N0IaPQU4Fodnpk04IVX:b-5h8ug6) ([raw file](https://github.com/sillsdev/libpalaso/blob/2a016de17f3199f8229387dc1ef5c5abb37b1b40/SIL.Windows.Forms/Reporting/WinFormsErrorReporter.cs#L40)):*
> ```Smalltalk
> 				ErrorReport.OnShowDetails(exception, message);
> 			}
> 		}
> ```

I guess you implemented it as blocking to try to preserve existing behavior?
It seems a little confusing that in the interface signature comment you suggest not blocking and then block here. Probably deserves more comments here.

SIL.Core/Reporting/ErrorReport.cs
```diff
@@ -473,23 +489,27 @@ public static void NotifyUserOfProblem(string message, params object[] args)
...
 
-		public static void NotifyUserOfProblem(Exception error, string messageFmt, params object[] args)
+		public static void NotifyUserOfProblem(Exception exception, string messageFmt, params object[] args)
 		{
-			NotifyUserOfProblem(new ShowAlwaysPolicy(), error, messageFmt, args);
+			NotifyUserOfProblem(new ShowAlwaysPolicy(), exception, messageFmt, args);
 		}
-
-		public static void NotifyUserOfProblem(IRepeatNoticePolicy policy, Exception error, string messageFmt, params object[] args)
+
```
ermshiperete, line 500: nit: whitespace

SIL.Core/Reporting/ErrorReport.cs
```diff
@@ -473,23 +489,27 @@ public static void NotifyUserOfProblem(string message, params object[] args)
...
 			{
-				OnShowDetails(error, string.Format(messageFmt, args));
+				s_previousNonFatalMessage = message;
+				return;
 			}
 
-			UsageReporter.ReportException(false, null, error, String.Format(messageFmt, args));
+			_errorReporter.NotifyUserOfProblem(policy, exception, message);
+
+			UsageReporter.ReportException(false, null, exception, String.Format(messageFmt, args));
```
ermshiperete, line 512:
```suggestion
			UsageReporter.ReportException(false, null, exception, message);
```

tombogle:
*[`SIL.Windows.Forms/Reporting/WinFormsErrorReporter.cs` line 40 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0IaPQU4Fodnpk04IVW:-N0WKaqZ1tj6-R2kY86I:bglnv2f) ([raw file](https://github.com/sillsdev/libpalaso/blob/2a016de17f3199f8229387dc1ef5c5abb37b1b40/SIL.Windows.Forms/Reporting/WinFormsErrorReporter.cs#L40)):*

I guess you implemented it as blocking to try to preserve existing behavior?
It seems a little confusing that in the interface signature comment you suggest not blocking and then block here. Probably deserves more comments here.
</blockquote></details>

I kind of think the WinForms implementation would have to block in order to be sensible. (I suppose it could maybe post a message to be handled later on the UI thread, but then it wouldn't necessarily appear in a context that makes sense to the user.)

sujeffreyl (commented):
Ah thanks, didn't know that. Done. See how it looks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-:-N0aQHa-BvsP9N99AkVB:bocn8iz)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @tombogle)

___
*[`SIL.Core/Reporting/ErrorReport.cs` line 33 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0WK5a13wgK8vUBN1Z-:-N0aFUWfBPKYFHIauasZ:b-ahwgi2) ([raw file](https://github.com/sillsdev/libpalaso/blob/2a016de17f3199f8229387dc1ef5c5abb37b1b40/SIL.Core/Reporting/ErrorReport.cs#L33)):*

This remark seems to contradict the above remark about not having to block. If it does not block, presumably it would always return ErrorResult.OK?
</blockquote></details>

Sorry for the confusion. The 2 methods have different requirements. I tried to make that clearer.

And yeah, this whole work was triggered because we found a case where it was prematurely returning ErrorResult.OK.
___
*[`SIL.Core/Reporting/ErrorReport.cs` line 512 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0IKygWgS74EYasM8Jh-r1-512:-N0W44mYBIoXVxmR49pp:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/2a016de17f3199f8229387dc1ef5c5abb37b1b40/SIL.Core/Reporting/ErrorReport.cs#L512)):*

```suggestion
			UsageReporter.ReportException(false, null, exception, message);
```
</blockquote></details>

Done.
___
*[`SIL.Windows.Forms/Reporting/WinFormsErrorReporter.cs` line 40 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0IaPQU4Fodnpk04IVW:-N0aFqmA0rzc9U8xl7lU:b-jm9afd) ([raw file](https://github.com/sillsdev/libpalaso/blob/2a016de17f3199f8229387dc1ef5c5abb37b1b40/SIL.Windows.Forms/Reporting/WinFormsErrorReporter.cs#L40)):*

I kind of think the WinForms implementation would have to block in order to be sensible. (I suppose it could maybe post a message to be handled later on the UI thread, but then it wouldn't necessarily appear in a context that makes sense to the user.) 
</blockquote></details>

Yeah, it was just to make it match the existing behavior more easily, no particularly good reason other than that.

tombogle:
*[`SIL.Core/Reporting/ErrorReport.cs` line 33 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0WK5a13wgK8vUBN1Z-:-N0aXtYwDrnHsfVC6K03:b-azwo2u) ([raw file](https://github.com/sillsdev/libpalaso/blob/2a016de17f3199f8229387dc1ef5c5abb37b1b40/SIL.Core/Reporting/ErrorReport.cs#L33)):*
<details><summary><i>Previously, sujeffreyl wrote…</i></summary><blockquote>

Sorry for the confusion. The 2 methods have different requirements. I tried to make that clearer.

And yeah, this whole work was triggered because we found a case where it was prematurely returning ErrorResult.OK.
</blockquote></details>

That makes more sense. In the XML comments, you can use <see cref="NotifyUserOfProblem(IRepeatNoticePolicy policy, string, ErrorResult, string)"/> to link to the specific version of the method you are talking about. This results in a nice hyperlink in the code so it is easy to find the method. Similarly in the comment below you can use a <see> element to ref to parameters. (If things referred to using the cref attributes in comments are ever renamed, the rename refactoring will also change the comment as well.)

tombogle:
*[`SIL.Core/Reporting/ConsoleErrorReporter.cs` line 23 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1160#-N0a_ygD7uKkvciF7REG:-N0a_ygD7uKkvciF7REH:be17ft3) ([raw file](https://github.com/sillsdev/libpalaso/blob/e71f2937dbb203a025a3d249a8bd9b883c7608d9/SIL.Core/Reporting/ConsoleErrorReporter.cs#L23)):*
> ```Smalltalk
> 		/// </summary>
> 		/// <param name="policy">The policy used to check if the message should be shown</param>
> 		/// <param name="error">N/A. You may pass null. This parameter will be ignored.</param>
> ```

I don't understand why this says "N/A" and claims that it will be ignored. The code definitely checks it for null and outputs it to the console if it is set.

tombogle (commented):
> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1160)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @sujeffreyl)

andrew-polk (approved):
> Reviewed 2 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1160)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @tombogle)

tombogle:
I also checked all 4 programs I work with an none of them ever pass null to the exception parameter.

tombogle (approved):

andrew-polk (approved):
> Dismissed @ermshiperete and @tombogle from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1160)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk, @ermshiperete, and @tombogle)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1160)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #1162, opened 2022-05-09
PTX-21397 revise book search based on Paratext testing

jwickberg:
Paratext testers found a couple of cases where the results from the pasting test in the verse control were not correct based on diacritics or RTL markers in the pasted text. The code was changed to remove diacritics and RTL markers from the the search text and use the BaseName of the book for all searches.

The UX team also decided that the best partial match should be returned rather than always requiring a unique match.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1162)

### DIFF
--- a/SIL.Scripture/MultilingScrBooks.cs
+++ b/SIL.Scripture/MultilingScrBooks.cs
@@ -113,7 +113,7 @@ namespace SIL.Scripture
         /// <summary>The maximum number of Scripture books that may be returned to the caller.</summary>
         public const int LastBook = 66;
         /// <summary>Regular expression used to parse verse reference strings.</summary>
-		public const string VerseRefRegex = @"(?<book>\w?.*[a-zA-Z])\s?((?<chapter>\d+)(\D+(?<verse>\d+))?)?";
+		public const string VerseRefRegex = @"(?<book>\w?.*\p{L})\s?((?<chapter>\d+)(\D+(?<verse>\d+))?)?";
 
 		/// <summary>// Indicates whether to process deutero-canonical book names.</summary>
 		protected bool m_fProcessDeuteroCanonical = false;
--- a/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
+++ b/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
@@ -82,6 +82,7 @@ namespace SIL.Windows.Forms.Scripture.Tests
 		[TestCase("ngs 3:3", ExpectedResult = "MAT 1:1")] // search should fail since search text doesn't start on a word boundary
 		[TestCase("M 6:3", ExpectedResult = "MIC 6:3")]
 		[TestCase("M 28:18", ExpectedResult = "MAT 28:18")]
+		[TestCase("{bad...[regex", ExpectedResult = "MAT 1:1")] // verify that bad regex express in search text doesn't cause exception
 		public string PastedTextGetsExpectedResult(string text)
 		{
 			m_verseCtrl.VerseRef = new VerseRef("MAT", "1", "1", ScrVers.English);
--- a/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
+++ b/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
@@ -83,6 +83,7 @@ namespace SIL.Windows.Forms.Scripture.Tests
 		[TestCase("M 6:3", ExpectedResult = "MIC 6:3")]
 		[TestCase("M 28:18", ExpectedResult = "MAT 28:18")]
 		[TestCase("{bad...[regex", ExpectedResult = "MAT 1:1")] // verify that bad regex express in search text doesn't cause exception
+		[TestCase("CO 3:7", ExpectedResult = "COL 3:7")] // verify that we get COL 3:7 and not 1CO 3:7
 		public string PastedTextGetsExpectedResult(string text)
 		{
 			m_verseCtrl.VerseRef = new VerseRef("MAT", "1", "1", ScrVers.English);
--- a/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
+++ b/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
@@ -70,11 +70,11 @@ namespace SIL.Windows.Forms.Scripture.Tests
 		[TestCase("jhn 4:5", ExpectedResult = "JHN 4:5")]
 		[TestCase("ACT 99:888", ExpectedResult = "ACT 28:31")]
 		[TestCase("2 Cor 3:18", ExpectedResult = "2CO 3:18")]
-		[TestCase("2 Co\u0301r 3:18", ExpectedResult = "2CO 3:18")] // verify that diacritic will be skipped
-		[TestCase("2 C\u00f3r 3:18", ExpectedResult = "2CO 3:18")] // verify composed character will work
+		[TestCase("2 Co\u0301r 3:18", ExpectedResult = "2CO 3:18", ExcludePlatform = "Linux", Reason = "Clipboard problems with text")] // verify that diacritic will be skipped
+		[TestCase("2 C\u00f3r 3:18", ExpectedResult = "2CO 3:18", ExcludePlatform = "Linux", Reason = "Clipboard problems with text")] // verify composed character will work
 		[TestCase("2 C 3:18", ExpectedResult = "2CH 3:17")] // Partial match finds 2CH, but there are only 17 verses in chapter 3
 		[TestCase("2CO 3:18", ExpectedResult = "2CO 3:18")]
-		[TestCase("2CO 3\u200f:\u200e18", ExpectedResult = "2CO 3:18")]
+		[TestCase("2CO 3\u200f:\u200e18", ExpectedResult = "2CO 3:18", ExcludePlatform = "Linux", Reason = "Clipboard problems with text")]
 		[TestCase("2CO3:18", ExpectedResult = "2CO 3:18")]
 		[TestCase("2CO 3.18", ExpectedResult = "2CO 3:18")]
 		[TestCase("2CO 3", ExpectedResult = "2CO 3:1")]
--- a/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
+++ b/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs
@@ -59,36 +59,36 @@ namespace SIL.Windows.Forms.Scripture.Tests
 			Assert.AreEqual(32, m_verseCtrl.VerseRef.VerseNum);
 		}
 
-		[TestCase("2 Corinthians 3:18", "2CO 3:18")]
-		[TestCase("2 Corinthians3:18", "2CO 3:18")]
-		[TestCase("2 Corinthians3*18", "2CO 3:18")]
-		[TestCase("2 Corinthians", "2CO 1:1")]
-		[TestCase("PSA 119:176", "PSA 119:176")]
-		[TestCase("MRK 1:0", "MRK 1:0")]
-		[TestCase("MRK", "MRK 1:1")]
-		[TestCase("LUK 2.3", "LUK 2:3")]
-		[TestCase("jhn 4:5", "JHN 4:5")]
-		[TestCase("ACT 99:888", "ACT 28:31")]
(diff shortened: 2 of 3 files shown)


tombogle:
*[`SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs` line 85 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1doR9HBKCXrQA8V0Wk:-N1doR9HBKCXrQA8V0Wl:b-8ne5ip) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs#L85)):*
> ```Smalltalk
> 		[TestCase("M 6:3", "MIC 6:3")]
> 		[TestCase("M 28:18", "MAT 28:18")]
> 		public void PastedTextGetsExpectedResult(string text, string expectedResult)
> ```

Not a big deal, but in case you didn't know, you can give a unit test method a return type of string, use the ExpectedResult attribute, and then return the actual result and have NUnit do the comparison.

tombogle:
*[`SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs` line 82 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1doqPY2w7JrhI8cCAM:-N1doqPY2w7JrhI8cCAN:bpbquvb) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs#L82)):*
> ```Smalltalk
> 		[TestCase("2CO 3", "2CO 3:1")]
> 		[TestCase("Songs 3:3", "SNG 3:3")]
> 		[TestCase("ngs 3:3", "MAT 1:1")] // shouldn't find something that start on a word boundary
> ```

should be "starts" instead of "start" - but I think this comment could be worded more clearly. Perhaps: No BaseName contains a word that starts with "ngs"

tombogle:
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1098 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1dpq9l2O6vTnKqtlW_:-N1dq_N311Zo0c6h0nZb:brrllwv) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture/VerseControl.cs#L1098)):*

Might be more readable and more robust to use a Regex.
</blockquote></details>

In fact, you could build a Regex based on the value of mustStartWithText and simplify the code.

tombogle:
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1104 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1drY7I5nG0qkxOY48Y:-N1drY7I5nG0qkxOY48Z:bg2mff4) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture/VerseControl.cs#L1104)):*
> ```Smalltalk
> 					return true;
> 
> 				// English versification seems to be pretty complete
> ```

I think this is true, and at least for most common scenarios, this is probably adequate, but since this control has an IScrVerseRef member and that interface allows you to get at the actual last chapter for a book, it seems that would be a clearer and more robust approach.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1162)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @jwickberg and @tombogle)

jwickberg:
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1098 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1dpq9l2O6vTnKqtlW_:-N1e7UimBnPm6ouIjErS:b-9s6to) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture/VerseControl.cs#L1098)):*

In fact, you could build a Regex based on the value of mustStartWithText and simplify the code.
</blockquote></details>

I don't want to use a regex at this point because it would be adding \s to user input which could contain just about anything since the regex used to get the book has a ".*" in it.

jwickberg (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1162)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @tombogle)

___
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1104 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1drY7I5nG0qkxOY48Y:-N1e9d5iBcAIB53NRKxT:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture/VerseControl.cs#L1104)):*

I think this is true, and at least for most common scenarios, this is probably adequate, but since this control has an IScrVerseRef member and that interface allows you to get at the actual last chapter for a book, it seems that would be a clearer and more robust approach.
</blockquote></details>

Done.
___
*[`SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs` line 82 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1doqPY2w7JrhI8cCAM:-N1e8BIfDGaAD1xJal7A:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs#L82)):*

should be "starts" instead of "start" - but I think this comment could be worded more clearly. Perhaps: No BaseName contains a word that starts with "ngs"
</blockquote></details>

Done.
___
*[`SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs` line 85 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1doR9HBKCXrQA8V0Wk:-N1e8D448z0zpPtACrye:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture.Tests/VerseControlTests.cs#L85)):*

Not a big deal, but in case you didn't know, you can give a unit test method a return type of string, use the ExpectedResult attribute, and then return the actual result and have NUnit do the comparison.
</blockquote></details>

Done.

tombogle:
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1098 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1dpq9l2O6vTnKqtlW_:-N1j98JK90WwKmfZagqm:b-xfcw6d) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture/VerseControl.cs#L1098)):*
> Is an ordinary space the only type of word boundary a BaseName could have?
It is possible to "escape" the searchText such that it would be safe to use as part of a regex, but doing it the way you are is probably okay if you can be quite certain that only a "normal" space could be used as a word boundary. I can't tell for sure from your test cases whether BaseName is always the English name of the book, or if it could be in some other language. Certainly the vast majority of languages will use regular spaces, but a scriptio continua language might use a narrow or zero-width spacing character. (Or they might not use any space at all, in which case this logic is not going to work.) Maybe that is rare enough that we don't need to build out this "nice-to-have" feature to accommodate it. In any case, probably either a comment to say that BaseName is always English or an ANHANCE comment to point out the possible need to handle scriptio continua languages would be fitting.

tombogle:
*[`SIL.Scripture/MultilingScrBooks.cs` line 116 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1jLE5OAgAb_RWQZusS:-N1jLE5PC2hrYMJNi1Y0:biw7phd) ([raw file](https://github.com/sillsdev/libpalaso/blob/4b0bb8b423cbd308ad0d4c09f9d4c77fbb38a06c/SIL.Scripture/MultilingScrBooks.cs#L116)):*
> ```Smalltalk
>         public const int LastBook = 66;
>         /// <summary>Regular expression used to parse verse reference strings.</summary>
> 		public const string VerseRefRegex = @"(?<book>\w?.*\p{L})\s?((?<chapter>\d+)(\D+(?<verse>\d+))?)?";
> ```

I realize that this is not what changed, but I'm not able to come up with any reason for the the leading \w? in the book. Since it is followed by a .*, it's the same as not having it at all. if we wanted to insist that a book name must begin with either a letter or a number (which is what I suspect this is intended to do, we want:
(?<book>[\p{L}\d].*\p{L})\s?((?<chapter>\d+)(\D+(?<verse>\d+))?)?



_Code quote:_
```Smalltalk
?<book>\w?.*\p{L})\s?((?<chapter>\d+)(\D+(?<verse>\d+))?)?
```

tombogle:
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1093 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1jOfx_AUNs0FDvq0Q0:-N1jOfx_AUNs0FDvq0Q1:b-tdfqic) ([raw file](https://github.com/sillsdev/libpalaso/blob/00bce64085cae0136968e0ca286453ac64462548/SIL.Windows.Forms.Scripture/VerseControl.cs#L1093)):*
> ```Smalltalk
> 				if (BaseName.Length < searchText.Length)
> 					return false;
> 				if (!Regex.IsMatch(BaseName, $"\\b{Regex.Escape(searchText)}", RegexOptions.IgnoreCase))
> ```

This change looks like a change in the logic such that you no longer prefer a match that occurs at the start of the name. Is that intentional? Wouldn't that mean that "Co 3:1" would match 1 Corinthians and stop before finding Colossians?

jwickberg:
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1093 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1jOfx_AUNs0FDvq0Q0:-N1jRQ-vE4dioGhjgCok:bh53g9y) ([raw file](https://github.com/sillsdev/libpalaso/blob/00bce64085cae0136968e0ca286453ac64462548/SIL.Windows.Forms.Scripture/VerseControl.cs#L1093)):*

This change looks like a change in the logic such that you no longer prefer a match that occurs at the start of the name. Is that intentional? Wouldn't that mean that "Co 3:1" would match 1 Corinthians and stop before finding Colossians?
</blockquote></details>

I thought 3 searches would be sufficient, but you did spot something that I missed. I'll go back to 4 searches.

jwickberg (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1162)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jwickberg and @tombogle)

___
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1098 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1162#-N1dpq9l2O6vTnKqtlW_:-N1jVUt_B3Tu0wHn4yjX:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/ec2afbc59e96d48e851f97a1f3d125251d40fe0f/SIL.Windows.Forms.Scripture/VerseControl.cs#L1098)):*

> Is an ordinary space the only type of word boundary a BaseName could have?
It is possible to "escape" the searchText such that it would be safe to use as part of a regex, but doing it the way you are is probably okay if you can be quite certain that only a "normal" space could be used as a word boundary. I can't tell for sure from your test cases whether BaseName is always the English name of the book, or if it could be in some other language. Certainly the vast majority of languages will use regular spaces, but a scriptio continua language might use a narrow or zero-width spacing character. (Or they might not use any space at all, in which case this logic is not going to work.) Maybe that is rare enough that we don't need to build out this "nice-to-have" feature to accommodate it. In any case, probably either a comment to say that BaseName is always English or an ANHANCE comment to point out the possible need to handle scriptio continua languages would be fitting.
</blockquote></details>

Done.

tombogle (approved):

PR #1169, opened 2022-05-19
Added several methods to IetfLanguageTag class to support getting language names.

tombogle:
This code was mostly pulled over from Bloom.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1169)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   (It is acceptable for implementers to just fill some parameters then call the original method)
   `ConsoleErrorReporter` and `WinFormsErrorReporter` implement `IErrorReporter`'s new interface method
 - [SIL.Core] Added override of SerializeToFileWithWriteThrough to simplify error handling.
+- [SIL.WritingSystems] Added several methods to IetfLanguageTag class to support getting language names.
 
 ### Changed
 
--- a/Palaso.sln.DotSettings
+++ b/Palaso.sln.DotSettings
@@ -15,32 +15,47 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=BBCCCVVV/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=bidi/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=bldr/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Canonicalizes/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Clipboarding/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Crowdin/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Dari/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Deuterocanon/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=deuterocanonical/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=else_0027s/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=espa_00F1ol/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Ethnologue/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=etic/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=filenames/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=fran_00E7ais/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=glist/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Iana/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Ibus/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=IETF/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=IMDI/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=ingl_00E9s/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=JAARS/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Jira/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Keyman/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Klavier/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Latn/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=LDML/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=libxklavier/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=macrolanguage/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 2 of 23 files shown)


tombogle:
Remaining work to get the tests to pass on Linux.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1169)* status: all files reviewed, 7 unresolved discussions (waiting on @tombogle)

___
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 944 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YN_Lc-s_Byq3xZABn:-N2YS2OV-KuXnVhqgfz-:byz78zt) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L944)):*

Is it a problem that these assume ICU will not be used?
</blockquote></details>

Note that Linux may normally have access to full ICU?

tombogle:
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 944 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YN_Lc-s_Byq3xZABn:-N2vigqz0477ALcIKqbL:b-uhfayi) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L944)):*

Note that Linux may normally have access to full ICU?
</blockquote></details>

Yes, that seems to be why some of them are failing on Linux.

tombogle:
*[`Palaso.sln.DotSettings` line 61 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2XU1dq6vWXUG-E7DF6:-N2vo-xq8NkJnQmGusMt:b-n4kxux) ([raw file](https://github.com/sillsdev/libpalaso/blob/e0e248ce9f0fbe3054c925fc7d70950b3ae49b7e/Palaso.sln.DotSettings#L61)):*

At the end of the day, it really doesn't matter, but you've added some things here which I wouldn't normally add to my "user dictionary" such as `_0020static_0020strin` and `Qaaa`, etc.
</blockquote></details>

I found and removed a few clunkers (including _0020static_0020strin). I don't really understand how those get in there, but it must be some kind of hiccup in Resharper. I've caught it in the act once or twice, but obviously I sometimes check them in without noticing. I wish there were some (clean and unobtrusive) way to be able to add non-English words such that they would only be permitted in Tests. Most of the time, I just don't bother adding them, but sometimes it feels expedient. For Qaaa, it seemed unlikely to ever be an accidental misspelling of a real word, and it's so riddled through the code that adding it made sense.

tombogle:
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 921 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YNAJICzgVuT51Bpl2:-N2vqWj01QXRATKN0los:b2chu5f) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L921)):*

Sorry; I'm having a hard time understanding exactly what is going on, but you changed something with the special logic here, and this seems to return a different result than what Bloom is currently getting, which is "Chinese (Simplified)".

Hm. Except I see that in the test below it gets "Simplified" so it must be two different paths. And what I'm looking at in Bloom is the one with the English if needed which is the test below.

But should these results be the same?
</blockquote></details>

I don't know. I noticed that they are different. Can you look to see where GetLocalizedLanguageName is used in Bloom to see if it would matter or make more sense one way or the other? If the UI is in English, his method currently just returns the DisplayName if available or the English Name as a fallback. Logically, I would expect them to both be set and to always/usually be the same, but for "zh-CN" they are different. DisplayName is "Chinese (Simplified, PRC)" and the EnglishName is "Chinese (Simplified, China)". Clearly, neither of those is the sanitized/simplified string we return as the English Subtitle.



_Code quote:_
```Smalltalk
[TestCase("zh-CN", "en", ExpectedResult = "Chinese (Simplified, PRC)")]
```

andrew-polk (changes requested):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1169)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 921 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YNAJICzgVuT51Bpl2:-N2xV9995jGFT9cXVn8o:bfktugf) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L921)):*

I don't know. I noticed that they are different. Can you look to see where GetLocalizedLanguageName is used in Bloom to see if it would matter or make more sense one way or the other? If the UI is in English, his method currently just returns the DisplayName if available or the English Name as a fallback. Logically, I would expect them to both be set and to always/usually be the same, but for "zh-CN" they are different. DisplayName is "Chinese (Simplified, PRC)" and the EnglishName is "Chinese (Simplified, China)". Clearly, neither of those is the sanitized/simplified string we return as the English Subtitle.
</blockquote></details>

We definitely use it, but I'm having a hard time tracking down exactly how and if it matters which version of the string appears.
We should probably chat about the change you made to the comma/parenthesis logic and if we want to bother making things any more consistent than Bloom had them or just leave it alone.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1169)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 921 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YNAJICzgVuT51Bpl2:-N35aVqNA6tugT_A8E45:b-1i0wf2) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L921)):*

We definitely use it, but I'm having a hard time tracking down exactly how and if it matters which version of the string appears.
We should probably chat about the change you made to the comma/parenthesis logic and if we want to bother making things any more consistent than Bloom had them or just leave it alone.
</blockquote></details>

I would certainly be ok with anything which preserves Bloom's existing behavior. 
Since we are having a hard time connecting synchronously, maybe just try to explain your take on the slight change in comma/parenthesis behavior so I can try to understand if Bloom would care.

In `GetNativeLanguageNameWithEnglishSubtitle`
In Bloom:
```
					// I have seen more cruft after the country name a few times, so remove that as well.
					// The parenthetical expansion always seems to start "(Simplified", which we want to keep.
					// We need double close parentheses because there's one open parenthesis before
					// "Chinese" and another open parenthesis before "Simplified" (which precedes ", China").
					// Also, we don't worry about the parenthetical content of the native Chinese name.
					var idxCountry = englishNameSuffix.IndexOf(", China");
					if (englishNameSuffix.Length > 0 && idxCountry > 0)
						englishNameSuffix = englishNameSuffix.Substring(0, idxCountry) + "))";
```
in this PR:
```
					// I have seen more cruft after the country name a few times, so remove that
					// as well. The parenthetical expansion always seems to start "(Simplified",
					// which we want to keep. We need double close parentheses because there's one
					// open parenthesis before "Chinese" and another open parenthesis before
					// "Simplified" (which precedes ", China" or ", PRC"). Also, we don't worry
					// about the parenthetical content of the native Chinese name.
					var idxCountry = englishNameSuffix.IndexOf(", ", StringComparison.Ordinal);
					if (englishNameSuffix.Length > 0 && idxCountry > 0)
						englishNameSuffix = englishNameSuffix.Substring(0, idxCountry) + "))";

```

Looking again, I see this is explicitly for the `zh-CN` tag. Presumably, you found a case (ICU?) where PRC is returned instead of China and decided that the intent of the existing logic would be to strip that, too. 

If I understand correctly, then I'm ok with that and with merging this PR as is.
If you want to discuss further whether it is worth trying to bring more consistency between methods, we can do that.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1169)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`SIL.WritingSystems/SIL.WritingSystems.csproj` line 10 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2XV-1ZBZ89sv61xf8b:-N2vppzd9kyRHCbF82BP:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/e0e248ce9f0fbe3054c925fc7d70950b3ae49b7e/SIL.WritingSystems/SIL.WritingSystems.csproj#L10)):*

wants `PrivateAssets="all" added so it doesn't get treated as a dependency for things which consume this dll
</blockquote></details>

Done.
___
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 734 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YMWxaA-xYvAFwxVkN:-N2vq73j7ITkI8GGflFD:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L734)):*

ReturnsVariant
</blockquote></details>

Done.
___
*[`SIL.WritingSystems.Tests/IetfLanguageTagTests.cs` line 946 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1169#-N2YNfrC4ckP26NVLyIC:-N2vmPgnEGSNokv_eB_6:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/f48674e92dce40b4202491a70678413c0e2a3867/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs#L946)):*

the fallback is not always English
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1169)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

PR #1187, opened 2022-07-01
Replace and/or update Mono.Posix library

### LINKED ISSUE #1186: Switch from deprecated Mono.Posix library to Mono.Unix
The SIL.Threading.GlobalMutex.LinuxGlobalMutexAdapter class (from GlobalMutex.cs) uses Mono.Posix.NETStandard 1.0.0 in its `netstandard2.0` build. But the [documentation for Mono.Posix](http://docs.go-mono.com/index.aspx?link=N%3AMono.Posix) now says "This namespace ... has been deprecated in favor of the new [Mono.Unix](http://docs.go-mono.com/monodoc.ashx?link=N%3aMono.Unix) namespace."

I believe that https://github.com/sillsdev/LfMerge/issues/279 is caused by the use of Mono.Posix.NETStandard 1.0.0, and would be resolved by switching to the https://www.nuget.org/packages/Mono.Unix package instead. It's possible that https://www.nuget.org/packages/Mono.Posix.NETStandard/5.20.1-preview would also work, which might suffice as a temporary measure. But since the Mono.Unix namespace is the correct one to use moving forward, we will eventually need to switch to it.

rmunn:
Fixes #1186.

This is currently a first-pass attempt, only replacing Mono.Posix with Mono.Unix in the .NET Standard builds, while keeping Mono.Posix in the net461 builds (though updating it to the most recent release of Mono.Posix). If this builds and passes tests, we'll then try replacing Mono.Posix with Mono.Unix everywhere and see what that does.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1187)

### DIFF
--- a/SIL.Core/SIL.Core.csproj
+++ b/SIL.Core/SIL.Core.csproj
@@ -10,12 +10,12 @@
     <PackageReference Include="JetBrains.Annotations" Version="2021.2.0" PrivateAssets="All" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
     <PackageReference Include="SIL.ReleaseTasks" Version="2.3.4-*" PrivateAssets="All" />
-    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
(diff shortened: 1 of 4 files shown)


ermshiperete:
Superseded by #1199

PR #1188, opened 2022-07-01
PTX-22682: Add copying/pasting from context menu in BCV control

rolfheij-sil:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1188)

### DIFF
--- a/SIL.Windows.Forms.Scripture/VerseControl.Designer.cs
+++ b/SIL.Windows.Forms.Scripture/VerseControl.Designer.cs
@@ -31,14 +31,14 @@ namespace SIL.Windows.Forms.Scripture
 			this.components = new System.ComponentModel.Container();
 			this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
 			this.uiVerseSpinner = new SIL.Windows.Forms.Widgets.HorizontalSpinner();
-			this.uiChapter = new SIL.Windows.Forms.Widgets.EnterTextBox();
+			this.uiChapter = new VCEnterTextBox();
 			this.uiChapter.CopyEvent += HandleCopy;
 			this.uiChapter.PasteEvent += HandlePaste;
-			this.uiVerse = new SIL.Windows.Forms.Widgets.EnterTextBox();
+			this.uiVerse = new VCEnterTextBox();
 			this.uiVerse.CopyEvent += HandleCopy;
 			this.uiVerse.PasteEvent += HandlePaste;
 			this.uiChapterSpinner = new SIL.Windows.Forms.Widgets.HorizontalSpinner();
-			this.uiBook = new SIL.Windows.Forms.Widgets.SafeComboBox();
+			this.uiBook = new VCSafeComboBox();
 			this.uiBook.CopyEvent += HandleCopy;
 			this.uiBook.PasteEvent += HandlePaste;
 			this.uiToolTip = new System.Windows.Forms.ToolTip(this.components);
@@ -169,10 +169,10 @@ namespace SIL.Windows.Forms.Scripture
 
 		private SIL.Windows.Forms.Widgets.HorizontalSpinner uiVerseSpinner;
 		private SIL.Windows.Forms.Widgets.HorizontalSpinner uiChapterSpinner;
-		private SIL.Windows.Forms.Widgets.EnterTextBox uiChapter;
-		private SIL.Windows.Forms.Widgets.EnterTextBox uiVerse;
+		private VCEnterTextBox uiChapter;
+		private VCEnterTextBox uiVerse;
 		private System.Windows.Forms.TableLayoutPanel tableLayoutPanel1;
 		private System.Windows.Forms.ToolTip uiToolTip;
-		private SIL.Windows.Forms.Widgets.SafeComboBox uiBook;
+		private VCSafeComboBox uiBook;
 	}
 }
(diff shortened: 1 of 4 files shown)


FoolRunning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1188)* status: all files reviewed, 3 unresolved discussions (waiting on @rolfheij-sil)

___
*[`SIL.Windows.Forms/Widgets/EnterTextBox.cs` line 12 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1188#-N6DPRVh0WXwjKznn0P3:-N6DPRVh0WXwjKznn0P4:bxh6e99) ([raw file](https://github.com/sillsdev/libpalaso/blob/6942338184724d7fb6ca85c0a8cb93549207c1c4/SIL.Windows.Forms/Widgets/EnterTextBox.cs#L12)):*
> ```Smalltalk
> 	/// default button on the form is selected instead.
> 	/// Also copy and paste event are overridden,
> 	/// so complete VerseRefs are copied/pasted
> ```

This is a generic textbox and doesn't know anything about a verse reference. I would expect the code changes done here will break normal copy/paste when this textbox is used normally.
___
*[`SIL.Windows.Forms/Widgets/SafeComboBox.cs` line 21 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1188#-N6DLwua9dLwyAocyw67:-N6DLwua9dLwyAocyw68:b-jwgku2) ([raw file](https://github.com/sillsdev/libpalaso/blob/6942338184724d7fb6ca85c0a8cb93549207c1c4/SIL.Windows.Forms/Widgets/SafeComboBox.cs#L21)):*
> ```Smalltalk
> 
> 		[DllImport("user32.dll", EntryPoint = "FindWindowExA", CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Ansi)]
> 		private static extern IntPtr FindWindowEx(IntPtr hwndParent, IntPtr hwndChildAfter, string lpszClass, string lpszWindow);
> ```

Using direct Win32 calls is not a good idea since this can run on Linux. Admittedly, this code probably doesn't ever run on Linux, but Palaso is a cross-platform library.
___
*[`SIL.Windows.Forms/Widgets/SafeComboBox.cs` line 38 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1188#-N6DPqbwBplYwq4fnurl:-N6DPqbwBplYwq4fnurm:b-k4bksi) ([raw file](https://github.com/sillsdev/libpalaso/blob/6942338184724d7fb6ca85c0a8cb93549207c1c4/SIL.Windows.Forms/Widgets/SafeComboBox.cs#L38)):*
> ```Smalltalk
> 
> 				textBoxHandle.CopyEvent += OnCopyEvent;
> 				textBoxHandle.PasteEvent += OnPasteEvent;
> ```

Same worry as for the textbox above. This combo box is not specific to the VerseControl, so I expect that normal copy/paste from this combobox is now broken.

rolfheij-sil:
I was under the impression that overriding the normal copy/paste behavior was the intended goal of this issue. How would you describe the desired behavior?

About using user32.dll: There already were 38 uses of it in the libpalaso code, so I though it wouldn't hurt adding one more

FoolRunning:
@rolfheij-sil, I think what you did was the correct behavior, but it was done on the wrong controls. The problem is that we only want this special copy/paste behavior on the VerseControl. To get the behavior, you changed the SafeComboBox and the EnterTextBox - which are other public controls that are used separately from the VerseControl. What you probably need to do is create private versions of these classes inside the VerseControl where you can override the behavior there.

Yes, you are correct about the Win32 thing. I didn't realize it was used so much in Palaso. However, all uses of the external code is inside a ``Platform.IsWindows`` check. I think this should be as well:
```
if (Platform.IsWindows)
    GetTextBoxHandle();
```

rolfheij-sil:
@FoolRunning I think the SafeComboBox and EnterTextBox controls are only used in the VerseControl. I can't find any references to them outside of VerseControl.

I'll add the IsWindows check

rolfheij-sil:
@ermshiperete Clear. I'll update the implementation

FoolRunning (changes requested):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1188)* status: all files reviewed, 1 unresolved discussion (waiting on @rolfheij-sil)

___
*[`SIL.Windows.Forms.Scripture/VerseControl.cs` line 1207 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1188#-N6IQbs-92T29vRDKOKB:-N6IQbs-92T29vRDKOKC:bce2mu1) ([raw file](https://github.com/sillsdev/libpalaso/blob/f5b206ed08e6146a5c116472d9ce43b1775a7e77/SIL.Windows.Forms.Scripture/VerseControl.cs#L1207)):*
> ```Smalltalk
> 	/// that is triggered from the context menu, so by using the mouse.
> 	/// </summary>
> 	internal class VCSafeComboBox : SafeComboBox
> ```

Ideally, these classes should be private and inside the VerseControl (there probably isn't much use for them outside of the VerseControl). Is there a reason you made them internal instead?

rolfheij-sil:
@FoolRunning Nesting classes isn't really used in C++, so I think that's why I placed them outside the VerseControl class. And I couldn't mark them 'private' within the scope of the namespace, so I went with 'internal'.
Anyway, I've nested the classes now

PR #1192, opened 2022-07-08
Improved CheckedComboBox

tombogle:
- correctly display summary information
- display ItemCheck event in Designer
- display drowndown using correct width

Also enhanced TestApp to illustrate correct behavior of the CheckedComboBox

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1192)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1192)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete and @tombogle)

___
*[`SIL.Windows.Forms/CheckedComboBox/CheckedComboBox.cs` line 245 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1192#-N6igrJC4fq6UwShgHSj:-N6igrJC4fq6UwShgHSk:blqeedp) ([raw file](https://github.com/sillsdev/libpalaso/blob/99e4752eda592c5c79784e9cb34ab65e29256ef9/SIL.Windows.Forms/CheckedComboBox/CheckedComboBox.cs#L245)):*
> ```Smalltalk
> 		}
> 
> 		public new string DisplayMember
> ```

Does removing this mean this is a +semver:major change (and should be explicitly documented in the changelog)?

Or now it just inherits the prop from the parent?

tombogle:
*[`SIL.Windows.Forms/CheckedComboBox/CheckedComboBox.cs` line 245 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1192#-N6igrJC4fq6UwShgHSj:-N6iwju-0DptGQnyoELu:b-41md2j) ([raw file](https://github.com/sillsdev/libpalaso/blob/99e4752eda592c5c79784e9cb34ab65e29256ef9/SIL.Windows.Forms/CheckedComboBox/CheckedComboBox.cs#L245)):*

Does removing this mean this is a +semver:major change (and should be explicitly documented in the changelog)?

Or now it just inherits the prop from the parent?
</blockquote></details>

Looking at this article: https://docs.microsoft.com/en-us/dotnet/core/compatibility/ (which doesn't actually mention this scenario) and attempting to think it through, I'm guessing that this is not a breaking change. The property existed before with the same type and visibility. There would be a hideously subtle behavior change if someone had used the previous version to access the base class property through a variable of type ComboBox (one of the reasons to make this change is to eliminate that code smell). In practice, I'm pretty sure no one else has attempted to use this new control, so if I'm wrong in my analysis, I guess I'll still get away with it. But I will update the changelog to point this out (in case it matters) and also to mention the other change that already made this a semver minor change.



_Code quote:_
```Smalltalk
protected override void OnDisplayMemberChanged(EventArgs e)
```

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1192)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

PR #1199, opened 2022-07-14
Replace deprecated Mono.Posix* with Mono.Unix

### LINKED ISSUE #1186: Switch from deprecated Mono.Posix library to Mono.Unix

ermshiperete:
Fixes #1186.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1199)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - [SIL.Windows.Forms.WritingSystems] Moved (internal) extension method InitializeWithAvailableUILocales to SIL.Windows.Forms.
 - [SIL.Windows.Forms.WritingSystems] Added additional optional localizationIncompleteViewModel parameter to ToolStripExtensions.InitializeWithAvailableUILocales.
+- Replace deprecated `Mono.Posix` dependency with `Mono.Unix` (#1186)
 
 ### Removed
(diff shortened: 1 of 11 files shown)


papeh (approved):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1199)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

PR #1202, opened 2022-07-21
Allowed for alternate way of getting Unicode character categories in  IsLikelyWordForming

tombogle:
Also added unit tests for this method

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1202)

### DIFF
--- a/SIL.Core.Tests/Extensions/StringExtensionTests.cs
+++ b/SIL.Core.Tests/Extensions/StringExtensionTests.cs
@@ -1,12 +1,9 @@
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.Design;
-using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
-using Microsoft.SqlServer.Server;
 using NUnit.Framework;
 using SIL.Extensions;
 using static System.String;
@@ -17,6 +14,18 @@ namespace SIL.Tests.Extensions
 	[TestFixture]
 	public class StringExtensionTests
 	{
+		[SetUp]
+		public void Setup()
+		{
+			Assert.That(StringExtensions.AltImplGetUnicodeCategory, Is.Null);
+		}
+
+		[TearDown]
+		public void TearDown()
+		{
+			StringExtensions.AltImplGetUnicodeCategory = null;
+		}
+
 		[Test]
 		public void SplitTrimmed_StringHasSpacesOnly_GivesEmptyList()
 		{
@@ -729,12 +738,10 @@ namespace SIL.Tests.Extensions
 		[TestCase("\u094C")] // SpacingCombiningMa
 		[TestCase("\u01C8")] // TitlecaseLetter:
 		[TestCase("Q")] // Uppercase letter
-		public void IsLikelyWordForming_WordformingCharactersKnownToDotNet_ReturnsTrue(string s)
+		public void IsLikelyWordForming_WordFormingCharactersKnownToDotNet_ReturnsTrue(string s)
 		{
-			Assert.That(StringExtensions.AltImplGetUnicodeCategory, Is.Null);
 			Assert.IsTrue(s.IsLikelyWordForming(0, false));
 		}
-
 
 		[TestCase(")")] // ClosePunctuation
 		[TestCase("\u203F")] // ConnectorPunctuation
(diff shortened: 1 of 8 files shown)


SIL.Core.Tests/Extensions/StringExtensionTests.cs
```diff
@@ -716,6 +720,79 @@ public void GetLongestUsefulCommonSubstring_SurrogatePairsWithSymbolsAsWordBreak
...
+				s == vietnameseAlternateReadingMark ? UnicodeCategory.SpacingCombiningMark :
+				CharUnicodeInfo.GetUnicodeCategory(s, i);
+
+			try
+			{
+				Assert.IsTrue(vietnameseAlternateReadingMark.IsLikelyWordForming(0, false));
+			}
+			finally
+			{
+				StringExtensions.AltImplGetUnicodeCategory = null;
```
ermshiperete, line 792: We could also do this in a setup method that runs before each test. That way we wouldn't have to check for `StringExtensions.AltImplGetUnicodeCategory` being `null` at the start of each test, and we wouldn't need this try/finally block here.

tombogle:
*[`SIL.Core/Extensions/StringExtensions.cs` line 604 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1202#-N7XaEgI8lTQ-nFidttL-r1-604:-N7aST8m-oGlzdUIwiZ5:b4xwmgy) ([raw file](https://github.com/sillsdev/libpalaso/blob/003f2ede1a0fd475402ab6fac4eaba9012cb4892/SIL.Core/Extensions/StringExtensions.cs#L604)):*

It might be good to add an example in the comment how you'd use ICU to get the character category, or add a unit test that demonstrates this.
</blockquote></details>

I did one better. I created and hooked up the only implementation anyone will _probably_ ever need as part of the Sldr initialization..

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1202)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @ermshiperete, @JohnThomson, and @papeh)

___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 775 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1202#-N7XaEgI8lTQ-nFidttK-r1-775:-N7aZjVmDgbzlaPfy_di:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/003f2ede1a0fd475402ab6fac4eaba9012cb4892/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L775)):*

I find this test hard to understand - it's not very clear what we currently get (without the AltImplGetUnicodeCategory extension) and what we expect (ok, it says it in the test name...).

So I wonder if it'd be better to assert that `IsLikelyWordForming` will return the expected result (which will always fail) and add a `Ignore` attribute to the test with the reasoning for this test.
</blockquote></details>

Done.
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 782 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1202#-N7XaEgI8lTQ-nFidttK-r1-782:-N7aUtmt7LS9EC__ojv3:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/003f2ede1a0fd475402ab6fac4eaba9012cb4892/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L782)):*

Renaming the variables to `str` and `index` would improve the documentation.
</blockquote></details>

Done.
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 783 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1202#-N7XaEgI8lTQ-nFidttK-r1-783:-N7aVcpc7QLcyCz6mqgH:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/003f2ede1a0fd475402ab6fac4eaba9012cb4892/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L783)):*

To improve the documentation purpose of this test, we should compare a substring:

```suggestion
				s.Substring(i, vietnameseAlternateReadingMark.Length)  == vietnameseAlternateReadingMark ? UnicodeCategory.SpacingCombiningMark :
```
</blockquote></details>

Done.
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 792 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1202#-N7XaEgI8lTQ-nFidttK-r1-792:-N7aWZ1GEVSMFXGiiOS6:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/003f2ede1a0fd475402ab6fac4eaba9012cb4892/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L792)):*

We could also do this in a setup method that runs before each test. That way we wouldn't have to check for `StringExtensions.AltImplGetUnicodeCategory` being `null` at the start of each test, and we wouldn't need this try/finally block here.
</blockquote></details>

Done.

SIL.WritingSystems/IcuUCharCategoryExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace SIL.WritingSystems
+{
+	public static class IcuUCharCategoryExtensions
+	{
+		public static UnicodeCategory ToUnicodeCategory(this Character.UCharCategory category)
+		{
+			switch (category)
+			{
+				case Character.UCharCategory.UNASSIGNED:
```
ermshiperete, line 12: Too bad that `Character.UCharCategory` and `UnicodeCategory` don't have the same values...

ermshiperete (approved):
:lgtm:

tombogle:
*[`SIL.WritingSystems/IcuUCharCategoryExtensions.cs` line 12 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1202#-N7aRofVkdEbHukzxaFV-r2-12:-N7b1FgqEEQJ5rfH8epB:bqog80t) ([raw file](https://github.com/sillsdev/libpalaso/blob/fec56aa306df156a4f06a628e15f1e1efff75320/SIL.WritingSystems/IcuUCharCategoryExtensions.cs#L12)):*

Too bad that `Character.UCharCategory` and `UnicodeCategory` don't have the same values...
</blockquote></details>

Agreed

PR #1203, opened 2022-07-22
Made ICU-based implementation of method to get Unicode category of a character more robust

tombogle:
Made the method testable and made it handle the case where the index passed in to IsLikelyWordForming is the character between the high and low surrogates.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1203)

### DIFF
--- a/SIL.WritingSystems.Tests/SldrTests.cs
+++ b/SIL.WritingSystems.Tests/SldrTests.cs
@@ -444,5 +444,31 @@ namespace SIL.WritingSystems.Tests
 				Assert.That(File.Exists(langTagsPath), Is.False);
 			}
 		}
+
+		[Test]
+		public void GetUnicodeCategoryBasedOnICU_HighSurrogate_GetsCorrectCategory()
+		{
+			Assert.That(Sldr.GetUnicodeCategoryBasedOnICU("\U00020056\U000200D9", 0),
+				Is.EqualTo(UnicodeCategory.OtherLetter));
+			Assert.That(Sldr.GetUnicodeCategoryBasedOnICU("\U00020056\U000200D9", 2),
+				Is.EqualTo(UnicodeCategory.OtherLetter));
+		}
+
+		[Test]
+		public void GetUnicodeCategoryBasedOnICU_LowSurrogate_GetsCorrectCategory()
+		{
+			Assert.That(Sldr.GetUnicodeCategoryBasedOnICU("\U00020056\U000200D9", 1),
+				Is.EqualTo(UnicodeCategory.OtherLetter));
+			Assert.That(Sldr.GetUnicodeCategoryBasedOnICU("\U00020056\U000200D9", 3),
+				Is.EqualTo(UnicodeCategory.OtherLetter));
+		}
+
+		[Test]
+		public void GetUnicodeCategoryBasedOnICU_LowSurrogateAtStartOfString_ThrowsArgumentException()
+		{
+			var bogusString = "\U00020056\U000200D9".Substring(1);
+			Assert.That(() => Sldr.GetUnicodeCategoryBasedOnICU(bogusString, 0),
+				Throws.ArgumentException);
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


SIL.WritingSystems.Tests/SldrTests.cs
```diff
@@ -444,5 +444,31 @@ public void LanguageTags_NoCachedLangTagsSldrOffline_UseEmbeddedLangTags()
 				Assert.That(File.Exists(langTagsPath), Is.False);
 			}
 		}
+
+		[Test]
+		public void GetUnicodeCategoryBasedOnICU_HighSurrogate_GetsCorrectCategory()
+		{
+			Assert.That(Sldr.GetUnicodeCategoryBasedOnICU("\U00020056\U000200D9", 0),
+				Is.EqualTo(UnicodeCategory.OtherLetter));
+			Assert.That(Sldr.GetUnicodeCategoryBasedOnICU("\U00020056\U000200D9", 2),
```
ermshiperete, line 453: I personally would prefer to make this two test cases. That way it's clearer that we're testing two scenarios instead of one scenario with two asserts.

SIL.WritingSystems/Sldr.cs
```diff
@@ -161,21 +161,38 @@ private static void InitializeGetUnicodeCategoryBasedOnIcu()
...
+			if (index < 0 || index >= s.Length)
+				throw new ArgumentOutOfRangeException(nameof(index), index,
+					"Index out of range. Must be a positive number less than the" +
+					$" length ({s.Length}) of the string: {{s}}");
+
+			// Just to be a bit more robust (like the default
+			// "UnicodeInfo.GetUnicodeCategory" implementation), if they gave us the index
+			// of the low surrogate and there is a valid high surrogate character preceding
+			// it, we'll fix things up so it doesn't throw an exception.
+			if (char.IsLowSurrogate(s, index) && index > 0 &&
```
ermshiperete, line 190: This won't work on big-endian machines, but I don't think we have to worry about that for libpalaso :smile:

ermshiperete (approved):

SIL.WritingSystems/Sldr.cs
tombogle: Hmmm... Might be worth a comment.

FoolRunning: Doesn't endian-ness only affect the order of bytes in a character (i.e. the two bytes that make up each character)? I would expect the order of the characters to be unaffected by endian-ness.

tombogle: That's what this is. If the user incorrectly specifies the position between the low and the high surrogate, regardless of endian-ness, ICU considers it an invalid position. On a small-endian implementation, you can detect this (and "fix" the index) by looking to see if the preceding position is the low surrogate and specified one is high. But on a big-endian machine, it will be the opposite.

ermshiperete: > Doesn't endian-ness only affect the order of bytes in a character (i.e. the two bytes that make up each character)? I would expect the order of the characters to be unaffected by endian-ness.

yes, it affects the order of the bytes in a character, but it also affects the order of the high and low surrogate "characters" - that's why there is a BOM also for UTF-16 and UTF-32 texts.

FoolRunning: > That's what this is.

No, this method is dealing with characters in a string (each character is two bytes). **index** is the character index, not the byte index.

> but it also affects the order of the high and low surrogate "characters"

Can you point to some documentation about this? I can't find any evidence that endian-ness affects anything but the bytes of each character (UTF-16 and UTF-32 need BOM characters because each character is multi-byte).

tombogle: I think Tim is right: https://unicodebook.readthedocs.io/unicode_encodings.html. Unfortunately, I jumped the gun (twice) and merged this without getting this resolved, and then I pushed a little change to add the comment, which I think now is incorrect. (That one went in without a PR by mistake. I didn't realize that master was not protected in this repo.)

ermshiperete: @FoolRunning I guess you're right :smile: I got confused by the "high" and "low". The Unicode website also talks about  leading and trailing for high and low surrogates which makes it a bit clearer that the leading/high surrogate always comes first.

PR #1207, opened 2022-08-04
Fire event when invalid reference pasted. Prevent input of too long v…

rolfheij-sil:
…erse/chapter values

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1207)

### DIFF
--- a/SIL.Windows.Forms.Scripture/VerseControl.cs
+++ b/SIL.Windows.Forms.Scripture/VerseControl.cs
@@ -829,11 +829,11 @@ namespace SIL.Windows.Forms.Scripture
 			if (string.IsNullOrEmpty(chapter))
 				chapter ="1";
 			else if (chapter.Length > 3 && Regex.IsMatch(chapter, @"^\d+$"))
-				chapter = chapter.Substring(0, 3);
+				return false;
 			if (string.IsNullOrEmpty(verse))
 				verse = "1";
 			else if (verse.Length > 3 && Regex.IsMatch(verse, @"^\d+$"))
-				verse = verse.Substring(0, 3);
+				return false;
 			if (Canon.IsBookIdValid(searchBook))
 			{
 				book = searchBook;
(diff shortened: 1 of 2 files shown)


SIL.Windows.Forms.Scripture/VerseControl.cs
```diff
@@ -817,8 +828,12 @@ private void HandlePasteScriptureRef()
 			// chapter and verse is optional in regex, make it 1 if not given
 			if (string.IsNullOrEmpty(chapter))
 				chapter ="1";
+			else if (chapter.Length > 3 && Regex.IsMatch(chapter, @"^\d+$"))
```
tombogle, line 831: I suspect that the actual resulting behavior here would not be what a user would expect. I think maybe if the chapter or verse portion is longer than 3 digits, we would just want to treat it like any other invalid pasted reference. It's most likely a typo or something and any one of the extra digits could be the errant one.

tombogle (changes requested):

tombogle (changes requested):

SIL.Windows.Forms.Scripture/VerseControl.cs
```diff
@@ -57,6 +57,8 @@ public partial class VerseControl : UserControl
 		public event EventHandler InvalidBookEntered;
 		/// <summary>Fired when any part of the reference is invalid</summary>
 		public event EventHandler InvalidReferenceEntered;
+		/// <summary>Fired when any part of the pasted reference is invalid</summary>
+		public event EventHandler InvalidReferencePasted;
```
tombogle, line 61: Since this is a new public event, it should have an entry in the ### Added section of CHANGELOG.md (and it should have +semver:minor in the comment)

tombogle (changes requested):
See comment about adding entry to CHANGELOG

tombogle (approved):

PR #1210, opened 2022-08-19
Test for normalised DateTimes

papeh:
In Germany, sometimes a dot is used instead of the colon to separate hours, minutes, and seconds: "16.12" for "16:12".

* Use DateTimeExtensions to write DateTimes to files (in the few places that didn't already)
* Test that LDML data is written in an invariant culture (LIFT writing was already tested).

https://jira.sil.org/browse/LT-20698

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1210)

### DIFF
--- a/SIL.DictionaryServices.Tests/Lift/LiftWriterTests.cs
+++ b/SIL.DictionaryServices.Tests/Lift/LiftWriterTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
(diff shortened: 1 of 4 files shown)


tombogle:
I'll need to look at this in detail next week when I have a fresh brain and more time. I'm assuming that at least part of the purpose of this is to ensure round-tripping of saved data. I see some of the classes used by SayMore are affected, so I want to study this. I know we've done some things in the past to deal with round-tripping date data.

papeh:
Thanks, @tombogle. The only change is that `ToISO8601TimeFormatDateOnlyString()` also passes CultureInfo.InvariantCulture to ToString. I don't expect this to make a difference when the precise format is already specified, but always calling the central method makes it more discoverable and reduces the chance of typos.

SIL.DictionaryServices.Tests/Lift/LiftWriterTests.cs
```diff
@@ -902,13 +902,12 @@ public void EntryHasDateModified()
...
 		/// <summary>
-		/// Regression: WS-34576
+		/// Regression: WS-34576, LT-20698
 		/// </summary>
 		[Test]
 		public void Add_CultureUsesPeriodForTimeSeparator_DateAttributesOutputWithColon()
 		{
-			var culture = new CultureInfo("en-US");
-			culture.DateTimeFormat.TimeSeparator = ".";
+			var culture = new CultureInfo("de-DE") {DateTimeFormat = {TimeSeparator = "."}};
```
tombogle, line 910: Might not be strictly necessary, but maybe it would be good to make two test cases: one for en-US and one for de-DE. At least that would kind of preserve the history of this test's original behavior.

tombogle (changes requested):

tombogle (commented):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1210)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @papeh and @tombogle)

SIL.DictionaryServices.Tests/Lift/LiftWriterTests.cs
papeh: done

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1210)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

PR #1221, opened 2022-08-31
Add tests for valid LDML files

papeh:
Add a by-hand test that the LDML files in the FieldWorks repository are valid.
Add a test that downloads and verifies several LDML files from the SLDR.

Inspired by https://jira.sil.org/browse/LT-21098


SIL.WritingSystems.Tests/SldrTests.cs
```diff
@@ -274,7 +288,39 @@ public void GetLdmlFile_NotModified_DoesntDownloadNewFile()
...
+		private static void DownloadAndVerifyLDML(string isStaging, string ietfLanguageTag)
+		{
+			using var environment = new TestEnvironment(false);
+			var originalStagingValue = Environment.GetEnvironmentVariable(Sldr.SldrStaging);
+			try
+			{
+				Environment.SetEnvironmentVariable(Sldr.SldrStaging, isStaging);
+				var file = Path.Combine(environment.FilePath, environment.TryGetLdmlFile(ietfLanguageTag));
+				var adaptor = new LdmlDataMapper(new TestWritingSystemFactory());
+				adaptor.Read(file, new WritingSystemDefinition("en", "Latn", "US", string.Empty, "eng", false), ex => Assert.Fail(ex.ToString()));
```
ermshiperete, line 315: Hmm, I don't see much verification happening - unless that's happening in the LdmlDataMapper.

At a minimum I'd expect that we assert that `adaptor.Read` returns something.

SIL.WritingSystems.Tests/SldrTests.cs
```diff
@@ -113,7 +127,7 @@ public void GetLdmlFile_UnknownLanguage_StatusFileNotFound()
 			using var environment = new TestEnvironment(false);
 			const string ietfLanguageTag = "qaa";
 
-			var sldrStatus = environment.GetLdmlFile(ietfLanguageTag, out var filename);
+			var sldrStatus = environment.GetLdmlFile(ietfLanguageTag, out _);
 			if (sldrStatus == SldrStatus.UnableToConnectToSldr)
```
ermshiperete, line 131: Use `TryGetLdmlFile` here?

SIL.WritingSystems.Tests/SldrTests.cs
```diff
@@ -31,6 +31,20 @@ public TestEnvironment(bool sldrOffline = true, DateTime? embeddedAllTagsTime =
 
 			private TemporaryFolder FolderContainingLdml { get; }
 
+			/// <summary>
+			/// Tries to get the LDML file for the requested ietfLanguageTag from the SLDR.
+			/// If unable to connect to SLDR, ignores the remainder of the test.
+			/// </summary>
+			/// <returns>The name the downloaded LDML file (to get the full path, combine with <see cref="FilePath"/></returns>
+			public string TryGetLdmlFile(string ietfLanguageTag)
```
ermshiperete, line 39: I wonder if it would be better to modify the existing `GetLdmlFile` method and add the `Assert.Ignore` there? It seems that the other places you modified where `GetLdmlFile` gets called could also use this check. But I didn't check if it's used elsewhere where this Assert doesn't make sense.

ermshiperete (changes requested):

SIL.WritingSystems.Tests/SldrTests.cs
```diff
@@ -264,17 +272,52 @@ public void GetLdmlFile_NotModified_DoesntDownloadNewFile()
...
+		[Test]
+		[Category("LongRunning")] // ~2 minutes
+		public void GetLdmlFile_GetsValidLDML([Values("false", "true")] string isStaging,
+			[Values("ar", "az", "bn", "de", "en", "es", "en-GB", "fa", "fr", "hi", "hu", "id", "km", "ko", "ml", "ms", "my",
+				"ne", "pt", "ru", "rw", "sw", "ta", "te", "th", "tr", "ur", "vi", "zh", "zh-CN")] string ietfLanguageTag)
+		{
+			DownloadAndVerifyLDML(isStaging, ietfLanguageTag);
+		}
+
+		[Test]
```
ermshiperete, line 292: Could add a comment that this test runs with every build whereas we don't normally run the above test in CI builds.

ermshiperete (approved):

PR #1224, opened 2022-09-08
PTX-22860: Allow any single non-digit character as separator between …

rolfheij-sil:
…chapter and verse

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1224)

### DIFF
--- a/SIL.Scripture.Tests/MultilingScrBooksTests.cs
+++ b/SIL.Scripture.Tests/MultilingScrBooksTests.cs
@@ -187,6 +187,9 @@ namespace SIL.Scripture.Tests
 		[TestCase("Luk 5,15", ExpectedResult = 42005015)]
 		[TestCase("luk 5.15", ExpectedResult = 42005015)]
 		[TestCase("LUK5:15", ExpectedResult = 42005015)]
+		[TestCase("LUK5/15", ExpectedResult = 42005015)]
+		[TestCase("LUK5_15", ExpectedResult = 42005015)]
+		[TestCase("LUK5;15", ExpectedResult = 42005015)]
 		[TestCase("4T1:5", ExpectedResult = 55001005)]
 		public int ParseRefString_EdgeCases(string input)
 		{
--- a/SIL.Scripture.Tests/MultilingScrBooksTests.cs
+++ b/SIL.Scripture.Tests/MultilingScrBooksTests.cs
@@ -184,13 +184,16 @@ namespace SIL.Scripture.Tests
 		/// Tests edge cases when parsing a reference string
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
+		[TestCase("4T1:5", ExpectedResult = 55001005)]
 		[TestCase("Luk 5,15", ExpectedResult = 42005015)]
 		[TestCase("luk 5.15", ExpectedResult = 42005015)]
 		[TestCase("LUK5:15", ExpectedResult = 42005015)]
-		[TestCase("LUK5/15", ExpectedResult = 42005015)]
+		[TestCase("LUK 5/15", ExpectedResult = 42005015)]
 		[TestCase("LUK5_15", ExpectedResult = 42005015)]
 		[TestCase("LUK5;15", ExpectedResult = 42005015)]
-		[TestCase("4T1:5", ExpectedResult = 55001005)]
+		[TestCase("LUK 5३15", ExpectedResult = 42001001)]
+		[TestCase("LUK 5a15", ExpectedResult = 1001001)]
+		[TestCase("LUK5α15", ExpectedResult = 1001001)]
 		public int ParseRefString_EdgeCases(string input)
 		{
 			return m_mlscrBook.ParseRefString(input).BBCCCVVV;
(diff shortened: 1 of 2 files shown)


rolfheij-sil:
PTX-22860 is a rework item that was created because a number of characters that should be allowed as separators were rejected by the current implementation. These characters were '_' '/' and ';'.

The original issue is PTX-21397 which states:
image

For this reason I decided to change the RegEx accordingly.

Let me know if you want some more insight into the issues!

tombogle:
Unless we know of a case where we would need to allow for a letter as a separator or have a fairly plausible sense that it might be legitimate, I would still lean toward saying it needs to be either whitespace or punctuation (or _maybe_ a symbol). If there is a letter, I'm inclined to guess that it is just garbage and we shouldn't try to interpret it as something meaningful.

rolfheij-sil:
I see your point.
Would [\W_] work for you? So that basically excludes all letters and numbers as accepted separators
\W = [^a-zA-Z0-9_]

tombogle:
It's sometimes a little tricky to know when to tell RegEx whether to exclude what you don't want or include what you do want. Using the various predefined RegEx character classes makes it especially tricky because these are implementation/locale dependent. I _think_ that for the purpose of this regex we _probably_ don't need/want it to match on chapter or verse numbers that are not the "plain" decimal digits 0-9. But I'm not 100% sure about that. If we test it against a verse reference that uses Hindi digits, what do we want? In any case, I'm pretty sure that if we're trying to preclude the use of /a/ as a chapter/verse separator, we also want to exclude /α/. So if .net's idea of \W is [^a-zA-Z0-9_], then that won't work. I think you're better off including the specific Unicode character classes we think are valid.You might also need to add a unit test or two to demonstrate that stuff like JHN 5a6 and JHN 5a6 does not result in a match.

tombogle:
So maybe [this](https://regex101.com/r/bZPkNN/1).

tombogle (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1224)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @rolfheij-sil and @tombogle)

___
*[`SIL.Scripture/MultilingScrBooks.cs` line 116 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1224#-NBS8_AS0eRh6_mNKkAV:-NBm-QZXDxr262810okA:bb6ywmx) ([raw file](https://github.com/sillsdev/libpalaso/blob/2f50e816abc7aa9705e7fd774b75ef567abe8448/SIL.Scripture/MultilingScrBooks.cs#L116)):*

Is this going too far? Should letters be allowed to be C/V separators?
</blockquote></details>

LGTM
___
*[`SIL.Scripture.Tests/MultilingScrBooksTests.cs` line 187 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1224#-NBm-nZV9L1TCCfeQWGu:-NBm-nZV9L1TCCfeQWGv:bvglrg5) ([raw file](https://github.com/sillsdev/libpalaso/blob/8766c8323f63273cd39d113c88161c18c7cad07e/SIL.Scripture.Tests/MultilingScrBooksTests.cs#L187)):*
> ```Smalltalk
> 		/// </summary>
> 		[TestCase("4T1:5", ExpectedResult = 55001005)]
> ```

It's not clear what "4T" is supposed to be/mean. At a minimum, a comment saying that book 55 is 2TI would be helpful. But why should 4T even correspond to that? Seems like a fairly arbitrary guess. (I realize this might not be new behavior.)
___
*[`SIL.Scripture.Tests/MultilingScrBooksTests.cs` line 194 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1224#-NBm0eHP6AFzLWG4Fh03:-NBm0eHP6AFzLWG4Fh04:b95qth6) ([raw file](https://github.com/sillsdev/libpalaso/blob/8766c8323f63273cd39d113c88161c18c7cad07e/SIL.Scripture.Tests/MultilingScrBooksTests.cs#L194)):*
> ```Smalltalk
> 		[TestCase("LUK5_15", ExpectedResult = 42005015)]
> 		[TestCase("LUK5;15", ExpectedResult = 42005015)]
> 		[TestCase("LUK 5३15", ExpectedResult = 42001001)]
> ```

Probably good to add a comment here noting that ३ is Devanagari digit 3. Also, as I look at this test case, it's not clear whether it fails to identify the chapter and verse because a) the number is too large, b) there is only one number, or c) ३ is not in the correct Unicode character class. Maybe it would help to add a test case like "LUK ३:५" so that it would be more clear whether we are expecting to be able to process digits from other scripts or not.



_Code quote:_
```Smalltalk
5३
```

rolfheij-sil:
I also don't understand why 4T translates to 2TI, but the test was already there, so I decided to leave it. I've added a new test case for LUK ३:५, and added comments

tombogle (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1224)* status: all files reviewed, 2 unresolved discussions (waiting on @rolfheij-sil)

tombogle:
I approved and merged it.

PR #1227, opened 2022-09-21
Fix `GetDirectoryDistributedWithApplication`

ermshiperete:
When running under net6.0 our previous detection of unit tests no longer works because `GetEntryAssembly()` now returns the test runner. This change additionally searches the directory where SIL.Core.dll is located.

This change also moves the list of subdirectories where we look for files or directories into a private field. When looking for
directories we now also look directly in the application directory, which is a breaking change from previous behavior where we only looked in `DistFiles`, `common`, and `src` subdirectories. This matches the behavior of `GetFileDistributedWithApplication`.

+semver:major

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1227)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.DblBundle] Made UsxNode abstract.
 - [SIL.DblBundle] Made UsxNode.StyleTag virtual. Calling UsxChapter.StyleTag on a chapter end node returns null instead of throwing an exception.
 - [SIL.DblBundle] Made UsxChapter sealed.
+- [SIL.Core] `FileLocationUtilities.GetDirectoryDistributedWithApplication` not only checks in
+  `DistFiles`, `common`, and `src` subdirectories, but also directly in the application or solution directory.
 
 ## [10.1.0] - 2022-08-26
 
--- a/SIL.Core/EntryAssembly.cs
+++ b/SIL.Core/EntryAssembly.cs
@@ -3,7 +3,7 @@ using System.Reflection;
 
 namespace SIL
 {
-	public class EntryAssembly
+	public static class EntryAssembly
 	{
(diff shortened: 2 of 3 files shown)


tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1227)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

ermshiperete:
@tombogle Could I request your re-review? In fixing the failing tests on the Windows build agent I realized that this is actually a breaking change - the two lists with sub-directories that I moved into a field weren't completely identical - `GetFileDistributedWithApplication` also had the empty string in it's list, meaning it looked for the file also directly in the application directory. `GetDirectoryDistributedWithApplication` now has the same behavior. Technically this is a breaking change, but I think the chances of having two files with the same name but different content, one below the project root directory, the other under `DistFiles`, `src`, or `common` is pretty unlikely.
But I'd still appreciate a second pair of eyes on this :smile: Thanks!

tombogle:
Since we're making a (barely) breaking change related to this, why don't we remove the obsolete versions of DirectoryOfApplicationOrSolution  and DirectoryOfTheApplicationExecutable from FileLocator now? It doesn't look as though anything is using them now.

tombogle (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1227)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

ermshiperete:
> Since we're making a (barely) breaking change related to this, why don't we remove the obsolete versions of DirectoryOfApplicationOrSolution and DirectoryOfTheApplicationExecutable from FileLocator now? It doesn't look as though anything is using them now.

Good idea. I'll do that in a separate PR.

Issue #987, opened 2020-11-05
Improve how SLDR downloads langtag.json file


jwickberg:
When SLDR is initialized, it currently does a synchronous request to download the langtags.json file and this request can be slow on machines with slow internet connections.

It would be better if the download of the langtags.json file could be separated out from the initialization code and allow applications to decide when to make the call.

Also, the SLDR code currently uses the IF-MODIFIED-SINCE header and for a while this was broken on the server. It is working now, but WSTech would like us to switch to using the ETag method instead since this is now more common. They gave this as a reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

ermshiperete:
>It would be better if the download of the langtags.json file could be separated out from the initialization code and allow applications to decide when to make the call.

@jwickberg  Don't we have that already? There's `Sldr.InitializeLanguageTags()` which does the download of `langtags.json`. The application can decide when that gets called. If it has been called before (i.e. if `_languageTags` is not null) we won't do the download again. Or do you mean that we should always use the local `langtags.json` file, and do the download in a separate method that the application can call when it pleases to do so?

jwickberg:
@ermshiperete I created this when I found that Paratext was starting slowly on some machines because of the download of the langtags.json file - this was made worse because the IF-MODIFIED-SINCE wasn't working which meant the file was always being downloaded. That was corrected on the server, but WSTech suggested we do the check a different way.
If we are sure that the timestamp check is working - the flag is less important.
Probably need to check with WSTech again to see the best way to make sure that the timestamp works.

ermshiperete:
I think what needs to happen:

1. extract downloading code to a new public method
2. change signature of `Sldr.InitializeLanguageTags()` to `Sldr.InitializeLanguageTags(bool downloadLangTags = true)` and call the new download method only if `downloadLangTags` is `true`. This deserves a `+semver:major` in the commit message.
3. change download method to use ETag (IF-NONE-MATCH header) instead of IF-MODIFIED-SINCE header. This requires a way to store the ETag of the downloaded file somewhere, maybe in a separate file. Downloading should re-populate `_languageTags`.

PR #1232, opened 2022-10-14
Langtag download issue 987

elisunger:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1232)


SIL.WritingSystems/Sldr.cs
```diff
@@ -409,49 +410,90 @@ private static void LoadLanguageTagsIfNecessary()
...
+				var webRequest = (HttpWebRequest)WebRequest.Create(Uri.EscapeUriString(langtagsUrl));
+				webRequest.UserAgent = UserAgent;
+				webRequest.Timeout = 10000;
+				webRequest.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
+				using var webResponse = (HttpWebResponse)webRequest.GetResponse();
+				if (File.Exists(etagPath))
+				{
+					etag = File.ReadAllText(etagPath);
+					currentEtag = webResponse.Headers.Get("Etag");
+					if (etag == "")
```
ermshiperete, line 462: No need to check for empty `etag` after we got the `Etag` header. You can unconditionally write the new value (but maybe do it in the `webResponse.StatusCode != HttpStatusCode.NotModified` block).

SIL.WritingSystems/Sldr.cs
```diff
@@ -409,49 +410,90 @@ private static void LoadLanguageTagsIfNecessary()
...
 						// delete the old langtags.json file if a newer embedded one is available.
 						// this can happen if the application is upgraded to use a newer version of SIL.WritingSystems
 						// that has an updated embedded langtags.json file.
 						File.Delete(cachedAllTagsPath);
+						File.Delete(etagPath);
+					}
 					else
 						sinceTime = fileTime;
 				}
 				sinceTime += TimeSpan.FromSeconds(1);
```
ermshiperete, line 430: Since we don't send `sinceTime` to the server, we no longer need that variable here - we only need it above in line 419 so that we can compare it to `fileTime`.

SIL.WritingSystems/Sldr.cs
```diff
@@ -390,9 +390,10 @@ public static SldrStatus GetLdmlFile(string destinationPath, string canonicalLan
 			}
 		}
 
-		public static void InitializeLanguageTags()
+		public static void InitializeLanguageTags(bool downloadLangTags = true)
 		{
 			LoadLanguageTagsIfNecessary();
+			if (downloadLangTags) LoadLanguageTags();
```
ermshiperete, line 396: We should download the tags before we call `LoadLanguageTagsIfNecessary`

SIL.WritingSystems/Sldr.cs
```diff
@@ -409,49 +410,90 @@ private static void LoadLanguageTagsIfNecessary()
...
-					// it will throw WebException or have status HttpStatusCode.NotModified if file is unchanged or not get it
-					var langtagsUrl =
-						$"{SldrRepository}index.html?query=langtags&ext=json{StagingParameter}";
-					var webRequest = (HttpWebRequest) WebRequest.Create(Uri.EscapeUriString(langtagsUrl));
-					webRequest.UserAgent = UserAgent;
-					webRequest.IfModifiedSince = sinceTime;
-					webRequest.Timeout = 10000;
-					webRequest.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
-					using var webResponse = (HttpWebResponse) webRequest.GetResponse();
+		public static void LoadLanguageTags()
```
ermshiperete, line 436: We now have two methods with almost identical names (`LoadLanguageTags` and `LoadLanguageTagsIfNecessary`) which do different things which is confusing. Maybe rename this method to `DownloadLanguageTags`?

ermshiperete (changes requested):

ermshiperete (changes requested):
We're getting closer. There are a few things that still need to be changed.

I also added some unit tests that you can use to debug and verify the correct behavior. You'll have to run `git pull git@github.com:elisunger/libpalaso.git --rebase langtagDownload-issue-987` to get them.

SIL.WritingSystems/Sldr.cs
```diff
@@ -470,55 +471,90 @@ private static void LoadLanguageTagsIfNecessary()
...
+				// it will throw WebException or have status HttpStatusCode.NotModified if file is unchanged or not get it
+				var langtagsUrl =
+					$"{SldrRepository}index.html?query=langtags&ext=json{StagingParameter}";
+				var webRequest = (HttpWebRequest)WebRequest.Create(Uri.EscapeUriString(langtagsUrl));
+				webRequest.Headers.Set("If-None-Match", etag);
+				webRequest.UserAgent = UserAgent;
+				webRequest.Timeout = 10000;
+				webRequest.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
+				using var webResponse = (HttpWebResponse)webRequest.GetResponse();
+				if (webResponse.StatusCode != "OK")
```
ermshiperete, line 520: `StatusCode` returns an enum, so you can't compare to `string`:

```suggestion
				if (webResponse.StatusCode != HttpStatusCode.OK)
```

SIL.WritingSystems/Sldr.cs
```diff
@@ -470,55 +471,90 @@ private static void LoadLanguageTagsIfNecessary()
...
+			string etag;
+			string currentEtag;
+			try
+			{
+				// get SLDR langtags.json from the SLDR api compressed
+				// it will throw WebException or have status HttpStatusCode.NotModified if file is unchanged or not get it
+				var langtagsUrl =
+					$"{SldrRepository}index.html?query=langtags&ext=json{StagingParameter}";
+				var webRequest = (HttpWebRequest)WebRequest.Create(Uri.EscapeUriString(langtagsUrl));
+				webRequest.Headers.Set("If-None-Match", etag);
```
ermshiperete, line 515: The compiler doesn't like this line and fails with `Error CS0165 : Use of unassigned local variable 'etag'`. You should read `etagPath` before this line (you can copy lines 525 and 527).

imnasnainaec:
@ermshiperete This pr's author has no GitHub contributions since October 2022. Is the change worth having somebody else  finish it?

ermshiperete:
@imnasnainaec I'm not sure how important #987 is. But might be worth having someone else finish this PR.

imnasnainaec:
Continued in #1383

PR #1235, opened 2022-11-14
Fix requesting LDML file with default user id

ermshiperete:
Previous implementation caused us to request `uid=unknown` to which the server replied with `501 not implemented`. Since that was treated the same as the server not being available this wasn't noticeable unless you ran the tests marked as Explicit.

This change compares the uid with the default user id and omits the parameter if necessary. That causes the server to send a proper response and causes the explicit tests to pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1235)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1235)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

PR #1243, opened 2023-01-19
Updated Changelog to prepare for release of 11.0.0

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1243)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [11.0.0] - 2023-01-19
+
 ### Added
 
 - [SIL.Core] Added `SIL.Reporting.FontAnalytics` class.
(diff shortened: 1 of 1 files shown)


jwickberg (approved):
No objections to the change. Paratext 9.3 probably won't upgrade, Paratext 9.4 has a while to go before release so can update when needed.

Still learning about semver - what requires this to go to version 11? Only change I see is an added class (FontAnalytics). Seems that would fit the definition of a minor change - new functionality that is backward compatible.

tombogle:
There are quite a few things in the "Changed" and "Removed" sections. As a general rule, most things in those categories require a major change. Although Paratext 9.3 does not upgrade, releasing a new (major or minor) version should make it possible for plugins to use the new version and load it alongside the version that Paratext loads. Two different plugins (or a plugin + Paratext) can't load two different beta versions.

ermshiperete (approved):

tombogle:
> No objections to the change. Paratext 9.3 probably won't upgrade, Paratext 9.4 has a while to go before release so can update when needed.
> 
> Still learning about semver - what requires this to go to version 11? Only change I see is an added class (FontAnalytics). Seems that would fit the definition of a minor change - new functionality that is backward compatible.

@jwickberg - Just realized that since Paratext 9.3 (as well as 9.4) is targeting a beta version of the SIL v11 libraries, it will still be impossible for plugins targeting the stable v 11.0 release to load the needed version. This is why it is generally not recommended to target a beta version of a library in a release. If Paratext 9.3 can't/won't upgrade, then the only options are for plugins to remain "broken" (i.e., if the user installs two plugins that use L10nSharp for localization but don't have the same UI languages available, the user will get weird messages about a localization not being available), or for us to release yet another version of libpalaso (with no actual changes) to force .Net to recognize that the versions are incompatible. Fortunately, I know of no other release plugin that uses L10nSharp, so for now this probably only affects me, but because I have already fixed the localization bug in Transcelerator, I can't run it anymore, or easily release bug fixes for it. (I could, but I'd have to go back in history and branch from before that bug fix.)

tombogle:
@jwickberg Upon further review, I realized that I have not yet merged/release those changes, so I can just wait if needed.

PR #1250, opened 2023-05-24
This is to fix running on Ubuntu 22x

johnml1135:
- they don't have libdl.so - but only libdl.so.2
- see https://stackoverflow.com/questions/75855053/how-to-address-crash-due-to-missing-libdl-so-on-ubuntu-22

This work is untested

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1250)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - [SIL.Windows.Forms] Make `PalasoImage.FromFile(Robustly)` methods more robust
+- [SIL.Windows.Forms] Update dll to `libdl.so.2` to make compatible with Ubuntu 22.x.  Affects multiple projects.
 - [SIL.Core] Fixed `BulkObservableList.MoveRange` method when moving a single item forward.
 
 ## [12.0.0] - 2023-02-14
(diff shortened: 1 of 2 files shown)


lyonsil:
In ICU.net, this was fixed [between version 2.8.1 and 2.9.0](https://github.com/sillsdev/icu-dotnet/commit/0ab57ff02027cb9d1248ec8fd17f47952a5d3e13#diff-b50e994cb29a0a36086ad24043f3f413591a87ddc727ee111acdc8dbea95966f) to make the icu.net.dll.config file work properly.  The solution that ICU.net took will work on Macs, too, so it is something to consider if support beyond Linux is needed.

ermshiperete:
> I came across this when developing docker images for machine.py which uses SIL.WritingSystems. It will not initialize in Ubuntu 22 containers.
> 

That's a slightly different error which won't be fixed by this PR. The most likely reason is that the `icu.net.dll.config` file is missing which needs to exist next to `icu.net.dll`. And if you're running .NET Core 3.x or .NET 5+ you'll need icu.net 2.9+.

ermshiperete (approved):
LGTM, but can you please add a line to `CHANGELOG.md`? Thanks!

johnml1135:
Added comment to changelog - I believe in the right format.

ermshiperete (approved):

PR #1252, opened 2023-06-02
fix bug where a dll was treated as a folder when trying to find a directory

hahn-kev:
When calling `GetDirectoryDistributedWithApplication` and it doesn't find the path on the first try, it will attempt to use the location of the executing assembly. However the code incorrectly assumes that `Assembly.GetExecutingAssembly().Location` returns a directory path, when it actually returns a file path. This change fixes that.

I have not run any test. However I checked `Assembly.GetExecutingAssembly().Location` via linqpad against .Net 7 and .Net framework and they both return a file path. I have verified that `Path.GetDirectoryName` returns the full path, not just the name of the parent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1252)

### DIFF
--- a/SIL.Core/IO/FileLocationUtilities.cs
+++ b/SIL.Core/IO/FileLocationUtilities.cs
@@ -168,7 +168,7 @@ namespace SIL.IO
 			if (Directory.Exists(path))
 				return path;
 
-			var thisDirectory = GetProjectDirectory(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
+			var thisDirectory = GetProjectDirectory(Path.GetDirectoryName(typeof(FileLocationUtilities).Assembly.Location));
 			if (thisDirectory != DirectoryOfApplicationOrSolution)
 			{
 				path = GetDirectoryDistributedWithApplication(thisDirectory, subPath);
--- a/SIL.Core/IO/FileLocationUtilities.cs
+++ b/SIL.Core/IO/FileLocationUtilities.cs
@@ -168,7 +168,7 @@ namespace SIL.IO
 			if (Directory.Exists(path))
 				return path;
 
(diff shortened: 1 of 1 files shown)


ermshiperete (approved):
LGTM. Can you please add a line to `CHANGELOG.md`  ("Unreleased" section)? Thanks!

tombogle:
*[`SIL.Core/IO/FileLocationUtilities.cs` line 171 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1252#-NXH5v080HL7Q0BdINbg:-NXH5v09-jEefNSbumpn:bue78o) ([raw file](https://github.com/sillsdev/libpalaso/blob/021314c293a827456dfccef46a16f25b0b42434d/SIL.Core/IO/FileLocationUtilities.cs#L171)):*
> ```Smalltalk
> 				return path;
> 
> 			var thisDirectory = GetProjectDirectory(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
> ```

Although not common, there are situations where GetExecutingAssembly() can return null. I realize that this problem already existed in the code as it was, but since you're fixing this bug, it might make sense to also guard against a null reference exception.

hahn-kev:
@tombogle Taking a look at the value that it returns and the description "Gets the assembly that contains the code that is currently executing." Meaning the assembly that contains the calling function `GetDirectoryDistributedWithApplication`. To that end I think we should replace the call from `Assembly.GetExecutingAssembly()` to `typeof(FileLocationUtilities).Assembly` which is never null. 

They even have a comment in the [remarks](https://learn.microsoft.com/en-us/dotnet/api/System.Reflection.Assembly.GetExecutingAssembly?view=netstandard-2.0#remarks) for `GetExecutingAssembly` about using a type instead of this function. I suspect this function would be useful in generated code when you don't have a consistent type name to use to get a reference to the assembly.

Would this be an acceptable fix?

PR #1253, opened 2023-06-02
change detection of a unit test context

hahn-kev:
Previously `DirectoryOfTheApplicationExecutable` was determining that it was in a testing context by checking if `Assembly.GetEntryAssembly() == null` this is no longer always true, for example running tests in Rider.

This PR changes the code to use the presence of a loaded unit testing dll as the indicator that we're running tests. I have preserved the check for `Assembly.GetEntryAssembly() == null` so as to not break backwards compatibility with any code that currently depends on the unit test path running when the assembly is null. This should reduce this to only changing functionality when running tests and not in any other cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1253)

### DIFF
--- a/SIL.Core/Reflection/ReflectionHelper.cs
+++ b/SIL.Core/Reflection/ReflectionHelper.cs
@@ -1,7 +1,9 @@
 using SIL.IO;
 using System;
 using System.IO;
+using System.Linq;
 using System.Reflection;
+using Mono.Unix.Native;
 
 namespace SIL.Reflection
 {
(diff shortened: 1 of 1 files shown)


ermshiperete (approved):

SIL.Core/Reflection/ReflectionHelper.cs
```diff
@@ -461,13 +463,21 @@ public static string LongVersionNumberString
...
+			);
+
 		public static string DirectoryOfTheApplicationExecutable
 		{
 			get
 			{
 				string path;
-				bool unitTesting = Assembly.GetEntryAssembly() == null;
-				if (unitTesting)
+				// checking for assembly == null is for backwards compatibility with old code that may unintentionally call take the unit test path
```
tombogle, line 479: I think there is a typo (or some other source of unintelligible grammer) in this comment.

tombogle (approved):

PR #1254, opened 2023-06-02
Fix crash adding metadata to jpg from MS Paint

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1254)


papeh:
star
Only this file (star.jpg, created with mspaint on Windows 10) causes the crash. Another jpg created with mspaint on Windows 11 does not.

imnasnainaec:
We should probably do one of the following:
- delete this branch
- create an issue

PR #1256, opened 2023-06-16
Update Interop.WIA.dll to be AnyCpu instead of x86

jasonleenaylor:
- Regenerate Interop.WIA.dll with these steps
- Add a COM reference to Microsoft Windows Image Aquisition
- Set Embed=false
- Copy generated assembly from the output into the lib folder
- Remove the COM reference

(I wasn't able to easily generate and pack this assembly in the build)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1256)


andrew-polk (approved):

jasonleenaylor:
I tested, and the msil Interop.WIA.dll does not work in a 32bit app, so more work is required.

jasonleenaylor:
BloomDesktop only ships 32bit right now, so there is need.

PR #1271, opened 2023-07-07
+semver:major Moved versions of CanWriteToDirectories and CanWriteToDirectory

tombogle:
One version of each of these methods does not require Windows Forms. So by putting them into SIL.Core, they will be available for non-Forms applications.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1271)


papeh:
Seems reasonable.
We should leave the original API with [Obsolete] to give clients an opportunity to transition.

tombogle:
Maybe, but since I just added these a couple days ago and the originals have only gone out as betas, it's extremely unlikely that anyone has used them, and I did add the remarks that refer to the other versions.

papeh (approved):

PR #1280, opened 2023-07-19
add audio monitoring without exception catching

nabalone:
For Bloom https://issues.bloomlibrary.org/youtrack/issue/BL-12406/Dont-show-error-Dialog-if-mic-access-is-off
Create a version of BeginMonitoring that doesn't catch the exceptions so that they can be handled by the caller

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)

### DIFF
--- a/SIL.Media/Naudio/AudioRecorder.cs
+++ b/SIL.Media/Naudio/AudioRecorder.cs
@@ -175,9 +175,6 @@ namespace SIL.Media.Naudio
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// For the IAudioRecorder interface. <see cref="BeginMonitoring(bool)"/> </summary>
-		/// <remarks>
-		/// Methods in this class should not call this method (i.e., from inside code that has
-		/// a lock on this). Instead use <see cref="BeginMonitoringIfNeeded()"/>. </remarks>
 		/// ------------------------------------------------------------------------------------
 		public virtual void BeginMonitoring()
 		{
@@ -206,9 +203,6 @@ namespace SIL.Media.Naudio
 		/// thrown as a result of this call (besides the documented exception) will be caught
 		/// and reported via <see cref="ErrorReport.NotifyUserOfProblem(string,object[])"/>.
 		/// </param>
-		/// <remarks> Methods in this class should not call this method (i.e., from inside
-		/// code that has a lock on this). Instead use <see cref="BeginMonitoringIfNeeded()"/>.
-		/// </remarks>
 		/// ------------------------------------------------------------------------------------
 		public virtual void BeginMonitoring(bool catchAndReportExceptions)
 		{
--- a/SIL.Media/Naudio/AudioRecorder.cs
+++ b/SIL.Media/Naudio/AudioRecorder.cs
@@ -216,6 +216,19 @@ namespace SIL.Media.Naudio
 		}
 
 		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Same as BeginMonitoring method above but without catching exceptions
+		/// ------------------------------------------------------------------------------------
+		public void BeginMonitoringWithThrow()
+		{
+			bool monitoringStarted = false;
+			monitoringStarted = BeginMonitoringIfNeeded();
+			if (!monitoringStarted)
+				throw new InvalidOperationException("Only call this once for a new WaveIn device" +
+					" (i.e., when RecordingState is NotYetStarted or Stopped.");
+		}
+
+		/// ------------------------------------------------------------------------------------
 		/// <summary>Begins monitoring if in the correct state to do so</summary>
 		/// <returns><c>true</c> if monitoring was started; <c>false</c> if not in a state where
 		/// monitoring could be started (e.g., already monitoring or recording, or stopped
--- a/SIL.Media/Naudio/AudioRecorder.cs
+++ b/SIL.Media/Naudio/AudioRecorder.cs
@@ -174,6 +174,19 @@ namespace SIL.Media.Naudio
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
+		/// For the IAudioRecorder interface. See BeginMonitoring(bool catchAndReportException)
+		/// </summary>
+		/// <remarks>
+		/// Methods in this class should not call this method (i.e., from inside code that has
+		/// a lock on this). Instead use the private internal version.</remarks>
+		/// ------------------------------------------------------------------------------------
+		public virtual void BeginMonitoring()
+		{
+			BeginMonitoring(true);
+		}
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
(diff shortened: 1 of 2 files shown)


SIL.Media/Naudio/AudioRecorder.cs
```diff
@@ -215,6 +215,19 @@ public virtual void BeginMonitoring()
 
 		}
 
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Same as BeginMonitoring method above but without catching exceptions
+		/// ------------------------------------------------------------------------------------
+		public void BeginMonitoringWithThrow()
```
nabalone, line 222: Or would "BeginMonitoringWithoutCatchingExceptions" be a better name?

SIL.Media/Naudio/AudioRecorder.cs
```diff
@@ -215,6 +215,19 @@ public virtual void BeginMonitoring()
 
 		}
```
nabalone, line 216: Would it be better to have BeginMonitoring call BeginMonitoringWithThrow so as not to duplicate the little bit of logic and exception throwing? i.e.
`public virtual void BeginMonitoring()
{
			try
			{
				BeginMonitoringWithThrow();
			}
			catch (InvalidOperationException e)
			{
				// TODO narrow this to catching only the exception excplicitly thrown in BeginMonitoringWithThrow
				throw e;
			}
			catch (Exception e)
			{
				ErrorReport.NotifyUserOfProblem(new ShowOncePerSessionBasedOnExactMessagePolicy(),
					e, "There was a problem starting up volume monitoring.");
			}`

tombogle:
*[`SIL.Media/Naudio/AudioRecorder.cs` line 226 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1280#-N_jTQmV4YxJbMcxNVKe:-N_jTQmV4YxJbMcxNVKf:bi4rfll) ([raw file](https://github.com/sillsdev/libpalaso/blob/d7e61b4e9b0a04577289d820f6fee38e9ff8dc1a/SIL.Media/Naudio/AudioRecorder.cs#L226)):*
> ```Smalltalk
> 			bool monitoringStarted = false;
> 			monitoringStarted = BeginMonitoringIfNeeded();
> 			if (!monitoringStarted)
> ```

To avoid duplication of code and prevent a scenario where a future improvement to the message might cause the two versions to get out of sync, maybe better to factor out this code to throw the exception. OR, have a single implementation method that takes a reportExceptionToUser parameter and in the catch blow merely re-throws the caught exception if !reportExceptiontoUser.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: all files reviewed, 3 unresolved discussions (waiting on @nabalone and @tombogle)

tombogle:
*[`SIL.Media/Naudio/AudioRecorder.cs` line 181 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1280#-N_ndLsECgdvFM487aJd:-N_ndLsECgdvFM487aJe:b-k5w7de) ([raw file](https://github.com/sillsdev/libpalaso/blob/4822466c80b27d4f7c6cf0151b6267f999c41b83/SIL.Media/Naudio/AudioRecorder.cs#L181)):*
> ```Smalltalk
> 		/// <remarks>
> 		/// Methods in this class should not call this method (i.e., from inside code that has
> 		/// a lock on this). Instead use the private internal version.</remarks>
> ```

It looks like the other version is also public.

nabalone:
The `Methods in this class should not call this method (i.e., from inside code that has a lock on this). Instead use the private internal version.` remarks comment was on the original `BeginMonitoring` method. Could it have referred to the private `BeginMonitoringIfNeeded` method? or is it just out of date?

tombogle:
You might need to look at the file history to see if you can figure out what it referred to. In any case, if doesn't make sense and there is no way to make it make sense, then better to remove it.

andrew-polk (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: all files reviewed, 6 unresolved discussions (waiting on @nabalone and @tombogle)

andrew-polk (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: all files reviewed, 6 unresolved discussions (waiting on @nabalone and @tombogle)

nabalone (commented):
Looked into it more. It looks likely to me that `BeginMonitoringIfNeeded` was at one point `BeginMonitoring_Internal` ? Regardless, I don't see a reason why it would be necessarily better to internally use`BeginMonitoringIfNeeded` over `BeginMonitoring` which thinly wraps `BeginMonitoringIfNeeded`. Could someone double check that though? I've removed the remarks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280#-:-N_tJs3y1GWE0hmZWHlK:bi4h83w)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @tombogle)

andrew-polk (approved):
@tombogle Thanks for reviewing this. Are you ready to approve it?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280#-:-NaEGffK6T14bnNZhFFw:blp1qk6)* status: all files reviewed, 6 unresolved discussions (waiting on @nabalone and @tombogle)

andrew-polk (approved):
@tombogle Also, it looks like you have draft comments pending.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280#-:-NaEb286CM4ht3cghD2E:buy9hqs)* status: all files reviewed, 6 unresolved discussions (waiting on @nabalone and @tombogle)

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @nabalone)

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

andrew-polk (commented):
> Dismissed @nabalone from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @nabalone)

andrew-polk (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1280)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

andrew-polk (approved):

PR #1281, opened 2023-07-21
Added some utility methods, ConfirmFileOverwriteDlg and new WaitCursor members

tombogle:
The only "change" (that probably only just barely warrants making a new major version) was to make the WaitCursor class (which previously only contained static methods and probably therefore should have been a static class) implement IDisposable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1281)

### DIFF
--- a/Palaso.sln.DotSettings
+++ b/Palaso.sln.DotSettings
@@ -75,7 +75,6 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Qaaa/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=ransomware/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=reentrant/@EntryIndexedValue">True</s:Boolean>
-	<s:Boolean x:Key="/Default/UserDictionary/Words/=Resharper/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Saami/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=SLDR/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=standarderror/@EntryIndexedValue">True</s:Boolean>
--- a/Palaso.sln.DotSettings
+++ b/Palaso.sln.DotSettings
@@ -75,6 +75,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Qaaa/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=ransomware/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=reentrant/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Resharper/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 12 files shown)


SIL.Windows.Forms/FileSystem/ConfirmFileOverwriteDlg.cs
```diff
@@ -0,0 +1,82 @@
+// ---------------------------------------------------------------------------------------------
+#region // Copyright (c) 2013, SIL International.
```
papeh, line 2: 2023

SIL.Core/Data/StrLengthComparer.cs
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Comparison method
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		public int Compare(string obj1, string obj2)
+		{
+			return m_asc * ((obj1?.Length ?? -1) - (obj2?.Length ?? - 1));
```
papeh, line 43: Elegant!

TestApps/SIL.Windows.Forms.TestApp/TestAppForm.cs
```diff
@@ -490,5 +491,23 @@ private void btnMediaFileInfo_Click(object sender, EventArgs e)
...
+			{
+				@"c:\folder\file.txt",
+				@"My Documents\another.doc",
+				@"LastOne.png"
+			};
+			foreach (var file in filenames)
+			{
+				using (var dlg = new ConfirmFileOverwriteDlg(file))
+				{
+					if (dlg.ShowDialog(this) == DialogResult.No && dlg.ApplyToAll)
```
papeh, line 507: Can "yes" not also be applied to all?
This is not the behaviour I'd expect, but I suppose it's ok for a test app.

Palaso.sln.DotSettings
```diff
@@ -75,6 +75,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Qaaa/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=ransomware/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=reentrant/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Resharper/@EntryIndexedValue">True</s:Boolean>
```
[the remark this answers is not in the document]
papeh: ("ReSharper" is already in the default dictionary)

TestApps/SIL.Windows.Forms.TestApp/TestAppForm.cs
tombogle: Apply to All can (of course) be interpreted however the client wishes to interpret it. But, yes, my normal expectation is that it would indeed apply to all. Since in this test app, we're not dealing with real files and not actually overwriting them, I just did this as a mindless way to demonstrate that both the dialog result and the value of ApplyToAll are being set as expected. I'll try to make it a little better.

SIL.Windows.Forms/FileSystem/ConfirmFileOverwriteDlg.cs
tombogle: It was originally written in 2013. Just moved it from Transcelerator. But the namespace did change (and it does look weird in the history), so I've updated it.

TestApps/SIL.Windows.Forms.TestApp/TestAppForm.cs
papeh: Much nicer!

papeh (approved):

PR #1284, opened 2023-10-03
Fixed i18n problem: SettingsProtection.LauncherButtonLabel was ID for two different strings

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1284)


tombogle:
Note that as a follow-on to this, we'll want to fix all existing translations in Crowdin because the string that survived the extraction was the one whose ID was changed.

PR #1293, opened 2023-10-13
Fix 4 img metadata methods that could fail due to cloud or scanning interference

hatton:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1293)


ermshiperete (approved):
This could use an entry in `CHANGELOG.md`, but otherwise LGTM

PR #1299, opened 2023-10-24
Fix hyperlink wrapped text cutoff in image dialog: LT-21652

aror92:
First part of issue [LT-21652](https://jira.sil.org/browse/LT-21652).

### DIFF
--- a/SIL.Windows.Forms/ClearShare/WinFormsUI/MetadataDisplayControl.cs
+++ b/SIL.Windows.Forms/ClearShare/WinFormsUI/MetadataDisplayControl.cs
@@ -97,9 +97,11 @@ namespace SIL.Windows.Forms.ClearShare.WinFormsUI
 			using (var g = this.CreateGraphics())
 			{
 				var w = this.Width - 10;
-				// UserControl does not have UseCompatibleTextRendering.
-				var h = TextRenderer.MeasureText(g, label, this.Font, new System.Drawing.Size(w, Int32.MaxValue), TextFormatFlags.WordBreak).Height;
-				var linkLabel = new LinkLabel() {Text = label, Width = this.Width - 10, Height = (int) (h + 5)};
+				var linkLabel = new LinkLabel() { Padding = Padding.Empty, Text = label };
+
+				// Have link label automatically determine its height, capping width at w.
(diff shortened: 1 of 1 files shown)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1299)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aror92 and @hatton)

___
*[`SIL.Windows.Forms/ClearShare/WinFormsUI/MetadataDisplayControl.cs` line 104 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1299#-NhXKNj02IK5erB8dPwH:-Ni0R9se2vlc1IixnFI2:bmxbp3n) ([raw file](https://github.com/sillsdev/libpalaso/blob/48056d75332e46f01237d4c0fdc09c37774b744d/SIL.Windows.Forms/ClearShare/WinFormsUI/MetadataDisplayControl.cs#L104)):*

@StephenMcConnel I have a faint recollection that maybe this doesn't work on mono (Linux)?
</blockquote></details>

There's no sign of that in the Bloom code:  we never take any precautions in setting LinkLabel.AutoSize to true.  (I'm not sure we ever use it quite like you do here, restricting the width but allowing the height to vary automatically.)  It's possible that may be one of the bugs we fixed in Mono over the years, although I don't have a record of fixing it myself.  I did find a mono bug report dated 2019 that fixed a display issue with LinkLabels in the mono code, but I don't think it had anything to do with AutoSize.
In short, I don't have a problem submitting this code, but it might be worth testing on Linux if possible.

aror92:
Thanks for the info. I don't have access to a Linux machine, but if someone else is able to test on Linux, that would be great.

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1299)* status: all files reviewed, 1 unresolved discussion (waiting on @aror92 and @hatton)

andrew-polk (commented):
> Dismissed @hatton from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1299)* status: all files reviewed, 1 unresolved discussion (waiting on @aror92)

PR #1314, opened 2024-03-28
Checked for disposed log box or caller-requested cancel in SafeInvoke

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1314)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
 - [SIL.Archiving] Upgraded to L10nSharp 7.0.0
 - [SIL.Windows.Forms] Upgraded to L10nSharp 7.0.0
 - [SIL.Windows.Forms.DblBundle] Upgraded to L10nSharp 7.0.0
 - [SIL.Windows.Forms.Keyboarding] Upgraded to L10nSharp 7.0.0
 - [SIL.Windows.Forms.WritingSystems] Upgraded to L10nSharp 7.0.0
+- [SIL.Core] `RaiseExceptionIfFailed` no longer throws an exception if user cancelled
 
 ## [13.0.1] - 2024-01-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - [SIL.Core] Fixed bug in extension method GetLongestUsefulCommonSubstring when string ends with an Object replacement character
+- [SIL.Core] LogBox: Checked for disposed log box or caller-requested cancel in SafeInvoke so we don't try to write messages or scroll.
 
 ## [13.0.0] - 2023-12-07
 
--- a/Palaso.sln.DotSettings
+++ b/Palaso.sln.DotSettings
@@ -31,6 +31,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=bidi/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=bldr/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=browsable/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=cancellability/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Canonicalizes/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=choco/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Clipboarding/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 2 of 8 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1314)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.Windows.Forms/Progress/LogBox.cs` line 290 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1314#-Nu585HX4w1hdTcyMb5n:-Nu585HX4w1hdTcyMb5o:b3lr0k2) ([raw file](https://github.com/sillsdev/libpalaso/blob/063e317ea775e7f4461950b58e9aeb9f4692ba9d/SIL.Windows.Forms/Progress/LogBox.cs#L290)):*
> ```Smalltalk
> 		{
> 			if (box.IsDisposed || box.Disposing || CancelRequested)
> 				return;
> ```

It is not clear to me why you would not allow the code to add more messages after the operation is cancelled.
The obvious example where you would want to do so is to add a message stating the operation has been cancelled. But seems like there could be others.

But maybe this only applies to an error condition?
Still I'm not convinced you wouldn't want to log an error after the user cancelled.

At a more meta level, should a method named something as generic as "SafeInvoke" never complete because the user cancelled an operation?

tombogle:
*[`SIL.Windows.Forms/Progress/LogBox.cs` line 290 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1314#-Nu585HX4w1hdTcyMb5n:-Nu5FjkWD2YQCfP27FT3:bp0tfpk) ([raw file](https://github.com/sillsdev/libpalaso/blob/063e317ea775e7f4461950b58e9aeb9f4692ba9d/SIL.Windows.Forms/Progress/LogBox.cs#L290)):*

It is not clear to me why you would not allow the code to add more messages after the operation is cancelled.
The obvious example where you would want to do so is to add a message stating the operation has been cancelled. But seems like there could be others.

But maybe this only applies to an error condition?
Still I'm not convinced you wouldn't want to log an error after the user cancelled.

At a more meta level, should a method named something as generic as "SafeInvoke" never complete because the user cancelled an operation?
</blockquote></details>

Fair questions. At least in HearThis, the dialog box that contains the log box is getting disposed when the user cancels. But maybe that would not always be the case. Perhaps I should only do this if it's disposed or disposing.

andrew-polk (changes requested):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1314)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.Core/CommandLineProcessing/CommandLineRunner.cs` line 314 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1314#-NueA8JUCUAqZ0f2-hvW:-NueA8JUCUAqZ0f2-hvX:b-3tycuv) ([raw file](https://github.com/sillsdev/libpalaso/blob/37f55d3b7dd8ca64afad1612db6c77adafe125d2/SIL.Core/CommandLineProcessing/CommandLineRunner.cs#L314)):*
> ```Smalltalk
> 			// should always already know if the user cancelled, not sure there is any need to
> 			// raise an exception in this case.
> 			if (ExitCode == 0 || UserCancelled)
> ```

This seems like a logical change.
And the only users of this method (AFAICT via github) are HearThis and Bloom.

Do you need to do anything in HearThis to account for this? If so, Bloom does too as it looks like our calling code was copied from HearThis.

tombogle:
*[`SIL.Core/CommandLineProcessing/CommandLineRunner.cs` line 314 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1314#-NueA8JUCUAqZ0f2-hvW:-Nue_ZRY6aaKsz5LxBiN:b-vpi1wk) ([raw file](https://github.com/sillsdev/libpalaso/blob/37f55d3b7dd8ca64afad1612db6c77adafe125d2/SIL.Core/CommandLineProcessing/CommandLineRunner.cs#L314)):*

This seems like a logical change.
And the only users of this method (AFAICT via github) are HearThis and Bloom.

Do you need to do anything in HearThis to account for this? If so, Bloom does too as it looks like our calling code was copied from HearThis.
</blockquote></details>

Good sleuthing. I'm making the change in order to try to prevent a current crashing bug in HearThis whenever the user cancels an export. I don't think I'll need to make any changes in HearThis itself, but I'm not 100% sure. Because it is a functional change that could theoretically break something, maybe it should be semver:major. I think it will already be part of a major version change when released.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1314)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tombogle)

___
*[`SIL.Core/CommandLineProcessing/CommandLineRunner.cs` line 314 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1314#-NueA8JUCUAqZ0f2-hvW:-NuizVx_B1RO0enwOViC:b-iw92rm) ([raw file](https://github.com/sillsdev/libpalaso/blob/37f55d3b7dd8ca64afad1612db6c77adafe125d2/SIL.Core/CommandLineProcessing/CommandLineRunner.cs#L314)):*

Good sleuthing. I'm making the change in order to try to prevent a current crashing bug in HearThis whenever the user cancels an export. I don't think I'll need to make any changes in HearThis itself, but I'm not 100% sure. Because it is a functional change that could theoretically break something, maybe it should be semver:major. I think it will already be part of a major version change when released.
</blockquote></details>

Maybe I'm just being lazy or helping you be lazy, but I'm ok with it going as is.

PR #1316, opened 2024-04-10
add usb drive info to net standard

hahn-kev:
previously drive info was excluded from the netstandard package, however this is blocking us from upgrading a number of projects to dotnet 8. This adds usb drive info back in using the `DriveInfo.GetDrives()` API which is avalible on all dotnet platforms. 

One thing that's not clear to me is how I would go about testing this. I wouldn't be suprised if there were differences between the versions being used, but I'm not even sure where to start here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1316)

### DIFF
--- a/SIL.Core.Desktop/UsbDrive/UsbDriveInfo.cs
+++ b/SIL.Core.Desktop/UsbDrive/UsbDriveInfo.cs
@@ -1,13 +1,10 @@
-﻿#if !NETSTANDARD
-// Copyright (c) 2009-2016 SIL International
+﻿// Copyright (c) 2009-2016 SIL International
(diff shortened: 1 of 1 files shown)


tombogle (approved):
Probably not feasible to unit test as part of the CI build because we wouldn't have control over the environment. (I imagine there would usually not be any USB drives,)  Perhaps create a "ByHand" unit test that just spits out the information so it can be verified manually. Maybe you could save the "report" to a known stable test location, and whenever the .Net Standard test runs, it could look for and (if found) compare its results the .net Framework results, and vice versa. You'd still get false negatives if a USB drive had been added or removed since the last time the test had been run using the other build, but it would at least give a heads-up to investigate.

PR #1318, opened 2024-04-26
Fixed backwards logic for LocalizationIncompleteViewModel.ShouldShowDialog

tombogle:
(Technically this is a breaking contractual change, since effectively the behavior is the opposite of the original implementation, but the name so clearly indicates the desired behavior that it seems unlikely any subclass implementation would have implemented the logic according to the previously expected backwards behavior.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1318)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


tombogle:
When this is merged and released and Transcelerator gets it, there is a TODO in the Transcelerator code to fix the implementation to work correctly.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1318)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tombogle)

PR #1319, opened 2024-05-08
Added static SilResources class with property SilLogo101x113.

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1319)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- [SIL.Windows.Forms] Added static SilResources class with property SilLogo101x113.
+
 ### Fixed
 
 - [SIL.Windows.Forms] Fixed backwards logic for LocalizationIncompleteViewModel.ShouldShowDialog (Technically this is a breaking contractual change, since effectively the behavior is the opposite of the original implementation, but the name so clearly indicates the desired behavior that it seems unlikely any subclass implementation would have implemented the logic according to the previously expected backwards behavior.)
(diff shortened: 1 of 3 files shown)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1319)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.Windows.Forms/Widgets/SilResources.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1319#-NxPPgMIA_bCcxAgzvpN:-NxPPgMIA_bCcxAgzvpO:b-w0vhf8) ([raw file](https://github.com/sillsdev/libpalaso/blob/54ebd03ab993a30cf07f5166686bbed776991a6e/SIL.Windows.Forms/Widgets/SilResources.cs#L10)):*
> ```Smalltalk
> 	{
> 		[PublicAPI]
> 		public static Bitmap SilLogo101x113 => Resources.SILLogoBlue101x113;
> ```

I wonder if the public API really wants to have the dimensions in the property name, but I guess that would be the only way to distinguish if we ever had more than one size.

tombogle:
*[`SIL.Windows.Forms/Widgets/SilResources.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1319#-NxPPgMIA_bCcxAgzvpN:-NxSOC2o66nWYBLcDc9i:b-xkic0e) ([raw file](https://github.com/sillsdev/libpalaso/blob/54ebd03ab993a30cf07f5166686bbed776991a6e/SIL.Windows.Forms/Widgets/SilResources.cs#L10)):*

I wonder if the public API really wants to have the dimensions in the property name, but I guess that would be the only way to distinguish if we ever had more than one size.
</blockquote></details>

That's what I was thinking

PR #1325, opened 2024-06-04
Renamed SIL.Media to SIL.Windows.Forms.Media

josephmyers:


tombogle:
I wonder if it might be best to split SIL.Media into a utility DLL and a Forms DLL, as we have with the other things. There are only a small handful of WinForms-specific things here.

josephmyers:
If it's a small handful, would there be any value in absorbing it into an existing library? Is the benefit of a separate dll enough to justify its cost?

tombogle:
I don't see any logical fit anywhere else. The only possible place would be Core, but _everything_ uses thats, so I think it would be less than ideal to put it there. Eventually, HearThis will probably become a non-WinForms app (or get a non-WinForms cousin), so breaking the functionality up really does make sense. I don't know enough about the architecture of Bloom, but it would be another possible candidate for using stuff in SIL.Media if it were not WinFoms dependent.

josephmyers:
I'm also fine leaving it as is. The only thing this PR does is make the name less misleading.

andrew-polk:
Bloom will need to have this teased apart eventually, so that is going to be helpful whenever it happens.
However, I think this rename is useful as a standalone change.
It needs a +semver:major in the commit message.
Also, unfortunately, this has languished long enough to have many merge conflicts.

josephmyers:
Yes, and I've changed roles as well, so this unfortunately now falls outside my purview. It may be easier to start from scratch.

PR #1326, opened 2024-06-06
Added Framework 4.8 to all projects

josephmyers:
And fixed all warnings other than those warning about Framework 4.6.1.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1326)

### DIFF
--- a/SIL.Core.Desktop/SIL.Core.Desktop.csproj
+++ b/SIL.Core.Desktop/SIL.Core.Desktop.csproj
@@ -6,7 +6,7 @@
     <Company>SIL</Company>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);netstandard2.0</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="GitVersion.MsBuild" Version="5.11.1" PrivateAssets="all" />
--- a/SIL.Core.Desktop/SIL.Core.Desktop.csproj
+++ b/SIL.Core.Desktop/SIL.Core.Desktop.csproj
@@ -11,7 +11,7 @@
   <ItemGroup>
     <PackageReference Include="GitVersion.MsBuild" Version="5.11.1" PrivateAssets="all" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="NDesk.DBus" Version="0.15.0" Condition="'$(TargetFramework)' != 'netstandard2.0'" />
+    <PackageReference Include="NDesk.DBus" Version="0.15.0" Condition=" $(DefineConstants.Contains('NETFRAMEWORK')) " />
     <PackageReference Include="SIL.ReleaseTasks" Version="2.5.0" PrivateAssets="All" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="6.0.0" />
     <PackageReference Include="System.Drawing.Common" Version="6.0.0" />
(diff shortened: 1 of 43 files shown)


SIL.Core.Desktop/SIL.Core.Desktop.csproj
```diff
@@ -20,7 +20,7 @@
   <ItemGroup>
     <ProjectReference Include="..\SIL.Core\SIL.Core.csproj" />
   </ItemGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
hahn-kev, line 23: I'd do this instead to ensure it only runs when targeting framework, otherwise if this ever references more recent dotnet the current setup will include this block
```suggestion
  <ItemGroup Condition=" $(DefineConstants.Contains('NETFRAMEWORK')) ">
```

josephmyers: Done

SIL.Core/IO/DirectoryHelper.cs
```diff
@@ -82,7 +82,7 @@ public static bool AreEquivalent(DirectoryInfo dirInfo1, DirectoryInfo dirInfo2)
 				: StringComparison.InvariantCulture;
 			var backslash = new[]
 			{
-				'\\', '/'
+				'\\', '/', '.'
```
hahn-kev, line 85: this this? why is there a '.' on the end of a directory name now?

hahn-kev (commented):
I left a question about a change.

Also, I figured out why `dotnet test` wasn't working and you had to pass in a list of dlls. It's because each test package needs to reference `<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />`. However when I did that I started getting some test failures. That said they seem to be due to some weird reference issues. If I start clean and use dotnet test then I get failures. However if I run a dotnet build first, then dotnet test works and the tests pass. So there's something weird going on. I'd say we would like to be able to use dotnet test without anything else, so we should look into that.

josephmyers:
SIL.WritingSystems.Tests takes a ridiculous amount of time to finish. It's 8-10 minutes on my machine, and enough on the build server to time out. This is exacerbated by increasing our supported frameworks. Seems an appropriate time to look into how we can get those tests speedier.

josephmyers:
The majority of the pain seems to be coming from the SldrTests. We could just not run these on the build server.

josephmyers:
[4a5ad40](https://github.com/sillsdev/libpalaso/pull/1326/commits/4a5ad402649638aa224e8df13b58c5c5264314de) succeeded in 10 minutes. Why did the previous time out at 30?

hahn-kev:
@josephmyers I was seeing some tests hang when working on my writing systems tests, there's now a timeout in dotnet test to detect that so if it happens you should have some details about it at least, this is now in master.

hahn-kev (approved):
LGTM

hahn-kev (changes requested):
oops, it looks like the following projects actually override `TargetFrameworks` meaning they don't pickup net48 from `Directory.Build.props`

- AddSortKey
- SIL.Core.Desktop
- SIL.Core
- SIL.DblBundle
- SIL.DictionaryServices
- SIL.Lexicon
- SIL.Lift
- SIL.Linux.Logging
- SIL.Scripture
- SIL.TestUtilities
- SIL.WritingSystems

josephmyers:
> oops, it looks like the following projects actually override `TargetFrameworks` meaning they don't pickup net48 from `Directory.Build.props`
> 
> * AddSortKey
> * SIL.Core.Desktop
> * SIL.Core
> * SIL.DblBundle
> * SIL.DictionaryServices
> * SIL.Lexicon
> * SIL.Lift
> * SIL.Linux.Logging
> * SIL.Scripture
> * SIL.TestUtilities
> * SIL.WritingSystems

Good catch. So why is the test run doubled? Looks to me like it's running everything, though based on the code, it's likely not

LibPalaso Tests
   34 files  +   17     34 suites  +17   9m 32s ⏱️ - 3m 43s
4 815 tests  -    68  4 584 ✅  -    68  231 💤 ±  0  0 ❌ ±0 
9 668 runs  +4 766  9 180 ✅ +4 522  488 💤 +244  0 ❌ ±0 

Results for commit https://github.com/sillsdev/libpalaso/commit/d51226f79643393b134126283aa26325aabef746. ± Comparison against base commit https://github.com/sillsdev/libpalaso/commit/269c813ab4ec593d4cedfa5072b7b23948929db7.

hahn-kev (approved):
LGTM

PR #1327, opened 2024-06-06
Add a macOS-compatible global lock

lyonsil:
The Linux approach for generating a lock doesn't work on macOS because `/var/lock` doesn't exist on macOS.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1327)

### DIFF
--- a/SIL.Core/Threading/GlobalMutex.cs
+++ b/SIL.Core/Threading/GlobalMutex.cs
@@ -15,6 +15,10 @@ namespace SIL.Threading
 	///
 	/// This is needed because Mono does not support system-wide, named mutexes. Mono does implement the Mutex class,
 	/// but even when using the constructors with names, it only works within a single process.
+	///
+	/// Note that in .NET 8.0 and later (and perhaps starting sooner than version 8), a named mutex can be used across
+	/// processes in Linux and macOS. However, software using the previous method of locking would not recognize the
+	/// new method of locking, so Linux continues to use the old, file-based approach internally.
 	/// </summary>
 	public class GlobalMutex : DisposableBase
 	{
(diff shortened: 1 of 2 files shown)


SIL.Core/Threading/GlobalMutex.cs
```diff
@@ -28,10 +28,12 @@ public class GlobalMutex : DisposableBase
 		public GlobalMutex(string name)
```
tombogle, line 28: Do we need to document that (on Mac) the actual mutex might not be named as requested? (Probably not important since it will always work like this.)

lyonsil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1327)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.Core/Threading/GlobalMutex.cs` line 28 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1327#gh-1630006175:-Nziq1RGEiAJR_I9SiJN:b-4tk9gd) ([raw file](https://github.com/sillsdev/libpalaso/blob/0aed06d6c7a7c9bc8663cf2b97b6de543abf070e/SIL.Core/Threading/GlobalMutex.cs#L28)):*

Do we need to document that (on Mac) the actual mutex might not be named as requested? (Probably not important since it will always work like this.)
</blockquote></details>

I added more comments to explain things in more detail.

tombogle (approved):

tombogle (changes requested):
There should be an entry in the changelog.md. In once sense, perhaps, this should be marked as semver:minor. On Mac, it is a new bit of functionality even though technically the interface itself didn't change.

lyonsil (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1327#-:-Nzj7PsB88Hzt2St0gwx:bb6un1j)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tombogle)

tombogle (approved):

tombogle (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1327)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lyonsil)

PR #1330, opened 2024-07-31
run writingsystem tests against dotnet 8 to show a bug in the integration between icu.net and SIL.WritingSystems

hahn-kev:
See explanation of the issue https://github.com/sillsdev/icu-dotnet/pull/202, 

this PR just adds a test to use `IcuRulesCollator.GetSortKey` which shows the current issue. Fixing this depends on the above PR being merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1330)


hahn-kev:
It looks like a lot of our WritingSystem tests (especially related to sorting) are not expecting dotnet 8 to use ICU natively. To resolve this I went ahead and used a configuration flag so that instead of dotnet using ICU it will use [NLS](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/globalization#nls) which is what .Net Framework uses, this meets the expectations of our tests. That said many of these tests may be over specified, so we should consider if we want to change them or not.

hahn-kev:
closed in favor of #1336 which is not based on a fork meaning there's no issue publishing the test results

PR #1335, opened 2024-08-01
Simulate a FileOk handler when choosing an image file

StephenMcConnel:
This enforces the image file filter even when users type or paste in a file path that defeats the filter.  This fix is motivated by https://issues.bloomlibrary.org/youtrack/issue/BL-13552.  Using the DialogAdapters interface prevents having an actual FileOk handler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1335)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs
+++ b/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs
@@ -81,59 +81,124 @@ namespace SIL.Windows.Forms.ImageToolbox
 		{
 			try
 			{
+				// Used to simulate having a FileOk event handler that can cancel the OK button click.
+				bool invalidFileChosen;
 				SetMode(Modes.SingleImage);
-				// The primary thing that OpenFileDialogWithViews buys us is the ability to default to large icons.
-				// OpenFileDialogWithViews still doesn't let us read (and thus remember) the selected view.
-				using (var dlg = new OpenFileDialogWithViews(OpenFileDialogWithViews.DialogViewTypes.Large_Icons))
+				do
 				{
-					if (string.IsNullOrEmpty(ImageToolboxSettings.Default.LastImageFolder))
+					invalidFileChosen = false;
+					// The primary thing that OpenFileDialogWithViews buys us is the ability to default to large icons.
+					// OpenFileDialogWithViews still doesn't let us read (and thus remember) the selected view.
+					using (var dlg = new OpenFileDialogWithViews(OpenFileDialogWithViews.DialogViewTypes.Large_Icons))
 					{
-						dlg.InitialDirectory = Environment.GetFolderPath(Environment.SpecialFolder.MyPictures);
-					}
-					else
-					{
-						dlg.InitialDirectory = ImageToolboxSettings.Default.LastImageFolder; ;
-					}
-
-					dlg.Title = "Choose Picture File".Localize("ImageToolbox.FileChooserTitle", "Title shown for a file chooser dialog brought up by the ImageToolbox");
-
-					//NB: disallowed gif because of a .net crash:  http://jira.palaso.org/issues/browse/BL-85
-					dlg.Filter = "picture files".Localize("ImageToolbox.PictureFiles", "Shown in the file-picking dialog to describe what kind of files the dialog is filtering for") + "(*.png;*.tif;*.tiff;*.jpg;*.jpeg;*.bmp)|*.png;*.tif;*.tiff;*.jpg;*.jpeg;*.bmp;";
-
-					if (DialogResult.OK == dlg.ShowDialog(this.ParentForm))
-					{
-						ImageToolboxSettings.Default.LastImageFolder = Path.GetDirectoryName(dlg.FileName);
-						ImageToolboxSettings.Default.Save();
-
-						try
+						if (string.IsNullOrEmpty(ImageToolboxSettings.Default.LastImageFolder))
 						{
-							_currentImage = PalasoImage.FromFile(dlg.FileName);
+							dlg.InitialDirectory = Environment.GetFolderPath(Environment.SpecialFolder.MyPictures);
 						}
-						catch (Exception err) //for example, http://jira.palaso.org/issues/browse/BL-199
+						else
 						{
-							var msg = "Sorry, there was a problem loading that image.".Localize(
-								"ImageToolbox.ProblemLoadingImage");
-							if (ImageLoadingExceptionReporter!= null)
-							{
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
Thinking through the implications of semantic versioning here....
in one sense, this could be considered a breaking change if any client were expecting to allow the loophole of pasting whatever file you want. But given that the library won't do anything sensible (I assume) with other file types, I think we can probably rule that out.
I think I'm ok with considering this a bug fix.

This is a good compromise, btw. Good thinking!

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1335#-:-O3E1xx88uimL87OxuqY:b-nbmbsi)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`CHANGELOG.md` line 27 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1335#-O3E2eHt-7Qa-m1IehDv:-O3E2eHt-7Qa-m1IehDw:b-gy0k8h) ([raw file](https://github.com/sillsdev/libpalaso/blob/5f048f83bfe627fd5ea246789bf1229b689217b2/CHANGELOG.md#L27)):*
> ```Markdown
> ### Fixed
> 
> - [SIL.Window.Forms] When choosing a file in the ImageToolbox.AcquireImageControl, a FileOk handler is simulated that verifies the selected file passes the given filter.  Users can defeat the filter mechanism by pasting or typing the file name.  While the returned filename does not pass the filter, the dialog is reopened until the user either chooses a proper filename or cancels the dialog.  The native FileOk handler can prevent the dialog from closing: we can't achieve that.  (See BL-13552.)
> ```

I think this has to go into a new "Fixed" section under "Unreleased" instead of 14.1.1.
___
*[`SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs` line 110 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1335#-O3E2lyx8QyevznYo_O5:-O3E2lyx8QyevznYo_O6:b-43x18z) ([raw file](https://github.com/sillsdev/libpalaso/blob/cc4c4f86495ca8d6ebdbc828bebfa79f3f92a1ce/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs#L110)):*
> ```Smalltalk
> 						if (DialogResult.OK == dlg.ShowDialog(this.ParentForm))
> 						{
> 							// Simulate having a FileOk event handler that can cancel the OK button click.
> ```

Probably worth calling out that we would prefer to use FileOk but that it is tied up in DialogAdapters so not worth changing at this time.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1335)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`CHANGELOG.md` line 27 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1335#-O3E2eHt-7Qa-m1IehDv:-O3E6kyY5w7EXi_rR6Pb:b-y7ua0g) ([raw file](https://github.com/sillsdev/libpalaso/blob/5f048f83bfe627fd5ea246789bf1229b689217b2/CHANGELOG.md#L27)):*

I think this has to go into a new "Fixed" section under "Unreleased" instead of 14.1.1.
</blockquote></details>

Good catch.  My eyes saw "Fixed" but didn't see the version number above it.
___
*[`SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs` line 110 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1335#-O3E2lyx8QyevznYo_O5:-O3E8ZmADtMYQAe2Y8dn:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/cc4c4f86495ca8d6ebdbc828bebfa79f3f92a1ce/SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs#L110)):*

Probably worth calling out that we would prefer to use FileOk but that it is tied up in DialogAdapters so not worth changing at this time.
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1335)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

SIL.Windows.Forms/ImageToolbox/AcquireImageControl.cs
```diff
@@ -81,61 +81,126 @@ private void GetImageFileFromSystem()
...
 						{
-							var msg = "Sorry, there was a problem loading that image.".Localize(
-								"ImageToolbox.ProblemLoadingImage");
-							if (ImageLoadingExceptionReporter!= null)
+							// Simulate having a FileOk event handler that can cancel the OK button click.
+							// We would prefer to actually use the event handler since that can prevent the
+							// dialog from closing, but we can't do that because of using DialogAdapters for
+							// cross-platform compatibility.  It's not worth messing with DialogAdapters at
+							// this point just to get this feature, which might not be cross-platform.
+							if (!DoubleCheckFileFilter(dlg.Filter, dlg.FileName))
```
megahirt, line 115: Do we still need DialogAdapters given that we are no longer shipping Bloom (or other software) on Mono Linux?  It would seem like a good idea to just remove the "cross-platform" code since it's not actually used and go with the simpler native implementation of FileOK for Windows.Forms.  Am I missing something?

andrew-polk: We thought Flex was still using the mono code in libpalaso. Is that incorrect?

megahirt: Can you fact-check the following for me, @jasonleenaylor ?

FieldWorks won't be shipping a new version on mono linux, so we're safe to remove mono cross-platform code from libpalaso master and reduce complexity for everyone.  I gather this PR would have been simpler without the mono requirement.  We don't even have build infrastructure anymore to build FW mono on Linux.

WinForms in libpalaso should be assumed to be Windows-only runtime.

jasonleenaylor: Yes @megahirt you are correct. We have stopped packaging for linux and are focusing on rearchitecting with the hope handle future cross-platform versions without mono. Sorry for any extra effort that went in because that wasn't widely known. When I saw this comment I assumed Bloom was still shipping for linux with mono.

PR #1348, opened 2024-10-03
.NET8 SIL.Core.Tests

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1348)


tombogle (approved):

hahn-kev (approved):
looks good to me.

One thing that might be nice is to make `SIL.Core.Tests` reference the CommandLineRunner test app project, then it'll always be built before the tests run.

PR #1349, opened 2024-10-04
+semver:major Added properties to ArchivingDlgViewModel to facilitate customizing the initial summary

tombogle:
Added unit tests and changed test app to illustrate intended usage of the new properties. Improved/added some comments in ArchivingDlgViewModel

Note: I considered removing OverrideDisplayInitialSummary because with this change, I anticipate that it will no longer be needed, but since it is inexpensive to leave in and allows for maximum flexibility, I decided to leave it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1349)

### DIFF
--- a/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
+++ b/SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
@@ -366,21 +366,16 @@ namespace SIL.Archiving.Tests
 				Assert.Fail($"Initialization threw an exception: {ex}");
 			}
 
-			Assert.That(messagesDisplayed.Count, Is.EqualTo(7));
-			Assert.That(messagesDisplayed[0].Item1, Is.EqualTo("First pre-archiving message"));
-			Assert.That(messagesDisplayed[0].Item2, Is.EqualTo(Warning));
-			Assert.That(messagesDisplayed[1].Item1, Is.EqualTo("Second pre-archiving message"));
-			Assert.That(messagesDisplayed[1].Item2, Is.EqualTo(Indented));
-			Assert.That(messagesDisplayed[2].Item1, Is.EqualTo("Frogs"));
-			Assert.That(messagesDisplayed[2].Item2, Is.EqualTo(Success));
-			Assert.That(messagesDisplayed[3].Item1, Is.EqualTo("green.frog"));
-			Assert.That(messagesDisplayed[3].Item2, Is.EqualTo(Bullet));
-			Assert.That(messagesDisplayed[4].Item1, Is.EqualTo("Label: Toads"));
-			Assert.That(messagesDisplayed[4].Item2, Is.EqualTo(Success));
-			Assert.That(messagesDisplayed[5].Item1, Is.EqualTo("red.toad"));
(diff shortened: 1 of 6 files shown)


SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
```diff
@@ -263,4 +265,123 @@ public void SetAbstract_MultipleLanguages_AddsDescriptionToCorpusImdiFile()
...
+			try
+			{
+				await model.Initialize(progress, new CancellationToken()).ConfigureAwait(false);
+			}
+			catch (Exception ex)
+			{
+				Assert.Fail($"Initialization threw an exception: {ex}");
+			}
+
+			Assert.That(messagesDisplayed.Count, Is.EqualTo(7));
```
hahn-kev, line 369: Might I suggest something like this instead
```c#
			Assert.That(messagesDisplayed, Is.EqualTo(new[]
			{
				("First pre-archiving message", Warning).ToTuple(),
				("Second pre-archiving message", Indented).ToTuple(),
				("Frogs", Success).ToTuple(),
				("green.frog", Bullet).ToTuple(),
				("Label: Toads", Success).ToTuple(),
				("red.toad", Bullet).ToTuple(),
				("blue.toad", Bullet).ToTuple()
			}));
```

SIL.Archiving/ArchivingDlgViewModel.cs
```diff
@@ -268,41 +323,87 @@ await Task.Run(() =>
...
+					DisplayMessage(Path.GetFileName(file), MessageType.Bullet);
+			}
+
+			Console.WriteLine($"{ArchiveType} Tests TEMP: {nameof(DisplayInitialSummary)} end");
 		}
 
 		protected virtual string FileGroupDisplayMessage(string groupKey)
 		{
-			return groupKey;
+			return OverrideGetFileGroupDisplayMessage == null ? groupKey : OverrideGetFileGroupDisplayMessage(groupKey);
```
hahn-kev, line 406: since `OverrideGetFileGroupDisplayMessage` could return null we should do `OverrideGetFileGroupDisplayMessage?.Invoke(groupKey) ?? groupKey` instead, then we will not return null unless groupKey was null, which would have been the case prior to this change.

hahn-kev (commented):
this looks good to me, I left a couple suggestions, though I'm not sure why this is a semver major change, I don't see any breaking changes, just new features added.

tombogle:
I think it is a technically a major version number change because I renamed a parameter. In most cases, it will recompile with no actual code changes, but if a caller were to use named parameters, then it could break. (Originally, I had intended to remove a method as well, but I decided to leave it.)

SIL.Archiving.Tests/IMDIArchivingDlgViewModelTests.cs
tombogle: Done

SIL.Archiving/ArchivingDlgViewModel.cs
tombogle: Done

hahn-kev (approved):

PR #1350, opened 2024-10-10
Changed copyrights from SIL International to SIL Global

tombogle:
Also added PublicAPI attribute to a lot of public members not used internally to make ReSharper happier. Did a bit of additional refactoring, code cleanup and fixing typos.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1350)


andrew-polk (changes requested):
I suppose it doesn't hurt anything, but the PublicAPI annotations everywhere feel like a lot of clutter, IMO.
If we were principled in applying it across the board where we think/know these are things clients want to or might want to use, it would be different. But I don't like just adding it to make the IDE happy. There's probably some global setting which can be modified instead?

The worst is when you use a libpalaso package and suddenly have a new dependency on the Jetbrains.Annotations package. I believe there is a setting to say this dependency should not be passed on to clients. (Is it `PrivateAssets="all"`?) That should be used everywhere for the Annotations package if we are going to keep using it.

> Reviewed 136 of 155 files at r1, 110 of 110 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1350#-:-OA-O3ZMDDcHVEdqNK5b:b9zcp9l)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.Core.Desktop/UsbDrive/Linux/UsbDriveInfoUDisks2.cs` line 2 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1350#-OA-P2jbDvE8jTcyGyxd:-OA-P2jbDvE8jTcyGyxe:b9ymyr8) ([raw file](https://github.com/sillsdev/libpalaso/blob/fe8c1bc8b0e48a22f23a162f2e5d2fc0126a8206/SIL.Core.Desktop/UsbDrive/Linux/UsbDriveInfoUDisks2.cs#L2)):*
> ```Smalltalk
> #if !NETSTANDARD
> // Copyright (c) 2024 SIL Global
> ```

Odd to have the copyright inside conditional code.

tombogle:
Yes, it's PrivateAssets="all". I was under the impression that that package
always got added that way, but if not, I should definitely fix that. There
probably is a setting to just turn off that complaint in the IDE. I'll also
look into that. (Though, of course, technically it is possible for public
members to be unintentionally public, so the alert isn't necessarily bad.)

andrew-polk (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1350)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tombogle)

tombogle (commented):
Fixed

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1350#-:-OA4UVXDCvI_xq-xptv7:b-886gsx)* status: 247 of 248 files reviewed, all discussions resolved (waiting on @tombogle)

___
*[`SIL.Core.Desktop/UsbDrive/Linux/UsbDriveInfoUDisks2.cs` line 2 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1350#-OA-P2jbDvE8jTcyGyxd:-OA4SAig0YDezmbdCJ5f:bd0pi9m) ([raw file](https://github.com/sillsdev/libpalaso/blob/fe8c1bc8b0e48a22f23a162f2e5d2fc0126a8206/SIL.Core.Desktop/UsbDrive/Linux/UsbDriveInfoUDisks2.cs#L2)):*

Odd to have the copyright inside conditional code.
</blockquote></details>

True. But that's how it was. Anyway, I've changed it.

PR #1357, opened 2024-10-30
Add an optional parameter to XmlUtils.WriteNode to preserve specific namespaces

jasonleenaylor:
We need this for FlexBridge to avoid some occasional data mangling during Send/Receive
I left the original behavior intact assuming that it was intentional, if it wasn't we could simply always write out the namespace names.

I also updated one of our gha dependencies in this PR, that is unrelated to the core change. (If someone else decides to merge this please rebase rather than merge to keep the changes separate)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1357)

### DIFF
--- a/SIL.Core.Tests/Xml/XmlUtilsTests.cs
+++ b/SIL.Core.Tests/Xml/XmlUtilsTests.cs
@@ -176,6 +176,62 @@ namespace SIL.Tests.Xml
 			}
 			Assert.That(output.ToString(), Is.EqualTo(expectedOutput));
 		}
+
+		[Test]
+		public void WriteNode_PreserveNamespacesArePreserved()
+		{
+			string input = @"<text><span class='bold' xml:space='preserve'> </span></text>";
+			string expectedOutput =
+				"<?xml version=\"1.0\" encoding=\"utf-16\"?>\r\n"
+				+ "<root>\r\n"
+				+ "	<text>\r\n"
+				+ "		<span\r\n"
+				+ "			class=\"bold\"\r\n"
+				+ "			xml:space=\"preserve\"> </span>\r\n"
+				+ "	</text>\r\n"
+				+ "</root>";
+			var output = new StringBuilder();
+			var preserveNamespace = new HashSet<string>();
+			preserveNamespace.Add("xml");
+			using (var writer = XmlWriter.Create(output, CanonicalXmlSettings.CreateXmlWriterSettings()))
+			{
+				writer.WriteStartDocument();
+				writer.WriteStartElement("root");
+				XmlUtils.WriteNode(writer, input, new HashSet<string>(), preserveNamespace);
+				writer.WriteEndElement();
+				writer.WriteEndDocument();
+			}
+			Assert.That(output.ToString(), Is.EqualTo(expectedOutput));
+		}
+
+		[Test]
+		public void WriteNode_ProtectsAgainstXmlnsFormatThrashing()
+		{
+			string input = @"<text><span class='bold' xmlns:fw='http://software.sil.org/fieldworks' fw:special='yes' xml:space='preserve'> </span></text>";
+			string expectedOutput =
+				"<?xml version=\"1.0\" encoding=\"utf-16\"?>\r\n"
+				+ "<root>\r\n"
+				+ "	<text>\r\n"
+				+ "		<span\r\n"
+				+ "			class=\"bold\"\r\n"
+				+ "			xmlns:fw=\"http://software.sil.org/fieldworks\"\r\n"
+				+ "			fw:special=\"yes\"\r\n"
+				+ "			xml:space=\"preserve\"> </span>\r\n"
+				+ "	</text>\r\n"
+				+ "</root>";
+			var output = new StringBuilder();
+			var preserveNamespace = new HashSet<string>();
+			preserveNamespace.Add("xml");
+			preserveNamespace.Add("xmlns");
+			preserveNamespace.Add("fw");
+			using (var writer = XmlWriter.Create(output, CanonicalXmlSettings.CreateXmlWriterSettings()))
+			{
+				writer.WriteStartDocument();
+				writer.WriteStartElement("root");
+				Assert.Throws<ArgumentException>(()=> XmlUtils.WriteNode(writer, input, new HashSet<string>(), preserveNamespace));
+			}
+		}
+
 		/// <summary>
 		/// This verifies that suppressing pretty-printing of children works for spans nested in spans nested in text.
 		/// </summary>
@@ -191,8 +247,10 @@ namespace SIL.Tests.Xml
 				+ "				class=\"italic\">bit</span>bt</span></text>\r\n"
 				+ "</root>";
 			var output = new StringBuilder();
-			var suppressIndentingChildren = new HashSet<string>();
-			suppressIndentingChildren.Add("text");
+			var suppressIndentingChildren = new HashSet<string>
+			{
+			   "text"
+			};
 			using (var writer = XmlWriter.Create(output, CanonicalXmlSettings.CreateXmlWriterSettings()))
 			{
 				writer.WriteStartDocument();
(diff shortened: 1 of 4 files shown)


andrew-polk:
Please add a changelog entry.
And, by strict semver, this wants a minor version bump.

SIL.Core/Xml/XmlUtils.cs
```diff
@@ -997,11 +999,18 @@ public static void WriteNode(XmlWriter writer, XElement dataToWrite, HashSet<str
...
-		private static void WriteElementTo(XmlWriter writer, XElement element, HashSet<string> suppressIndentingChildren)
+		/// <param name="preserveNamespaces">a set of namespaces to preserve when writing out elements</param>
+		private static void WriteElementTo(XmlWriter writer, XElement element, HashSet<string> suppressIndentingChildren, HashSet<string> preserveNamespaces = null)
 		{
 			writer.WriteStartElement(element.Name.LocalName);
 			foreach (var attr in element.Attributes())
-				writer.WriteAttributeString(attr.Name.LocalName, attr.Value);
+			{
+				var ns = attr.Name.NamespaceName;
+				if(preserveNamespaces != null && preserveNamespaces.Contains(ns))
```
hahn-kev, line 1009: what if I wanted to preserve all namespaces?

SIL.Core.Tests/Xml/XmlUtilsTests.cs
```diff
@@ -176,6 +176,36 @@ public void WriteNode_DoesNotIndentChildWhenSuppressed()
...
+		public void WriteNode_PreserveNamespacesArePreserved()
+		{
+			string input = @"<text><span class='bold' xml:space='preserve'> </span></text>";
+			string expectedOutput =
+				"<?xml version=\"1.0\" encoding=\"utf-16\"?>\r\n"
+				+ "<root>\r\n"
+				+ "	<text>\r\n"
+				+ "		<span\r\n"
+				+ "			class=\"bold\"\r\n"
+				+ "			 xml:space=\"preserve\"> </span>\r\n"
```
hahn-kev, line 190: I'm assuming this is what you want to keep around, but I'm not clear what it would look like otherwise, would it just be `space=`?

hahn-kev (approved):
looks good to me.

SIL.Core.Tests/Xml/XmlUtilsTests.cs
jasonleenaylor: yes, without the xml namespace the standard writer will ignore the 'space="preserve"' and drop that content.

jasonleenaylor:
@andrew-polk There is a +semver:minor on the relevant commit already and I added a changelog entry.

jasonleenaylor:
*[`SIL.Core/Xml/XmlUtils.cs` line 1009 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1357#gh-1823807772:-OAZWze13YfHsoH3BzQP:b1rsmk5) ([raw file](https://github.com/sillsdev/libpalaso/blob/b5efcc93f6de24c97fa0e4b7c12929393a88ff71/SIL.Core/Xml/XmlUtils.cs#L1009)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

what if I wanted to preserve all namespaces?
</blockquote></details>

Then you want a different change, I had considered making this an all on or all off, but landed on this solution.
It seems like most people don't care, and I only care about one.

andrew-polk (commented):
Thanks. Sorry, I had missed the `+semver:minor`

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1357#-:-OAZXMyb59RxaOVvjtBc:bz4e0zf)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @hahn-kev)

SIL.Core.Tests/Xml/XmlUtilsTests.cs
```diff
@@ -176,6 +176,37 @@ public void WriteNode_DoesNotIndentChildWhenSuppressed()
...
+		[Test]
+		public void WriteNode_PreserveNamespacesArePreserved()
+		{
+			string input = @"<text><span class='bold' xmlns:fw='https://fakefw.com' fw:special='no' xml:space='preserve'> </span></text>";
+			string expectedOutput =
+				"<?xml version=\"1.0\" encoding=\"utf-16\"?>\r\n"
+				+ "<root>\r\n"
+				+ "	<text>\r\n"
+				+ "		<span\r\n"
+				+ "			class=\"bold\" xmlns:fw=\"https://fakefw.com\"\r\n"
```
hahn-kev, line 189: so based on the test failure this line looks like this in net standard
`class=\"bold\"\r\n\t\t\txmlns:fw=\"https://fakefw.com\"\r\n`
so there's a newline between `class="bold"` and xmlns, but not on framework. I debugged this and that's just how the code is written, I don't really see a good way around this. Even reordering the namespace doesn't make a difference.
This is because there's a special case for `xmlns` specifically. 

Looking into the source indenting/formatting works with via overloads. Internally `WriteStartNamespaceDeclaration` is called which is used only for `xmlns` in dotnet standard this method is overloaded in `XmlEncodedRawTextWriterIndent` to indent the line, and in framework it's not. Sadly these are internal so we can't just provide our own version.

I suspect this will be a problem as it will thrash the xml when run on one version of dotnet vs another. I'm not sure how, but if we could use the netstandard version in framework then it would resolve this issue.

hahn-kev (approved):
looks good to me

jasonleenaylor (commented):
> Dismissed @hahn-kev from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1357)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @hahn-kev)

___
*[`SIL.Core.Tests/Xml/XmlUtilsTests.cs` line 189 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1357#gh-1827146157:-OB7VzmfDe81U1Nhvc40:br0ftka) ([raw file](https://github.com/sillsdev/libpalaso/blob/2c919a1f3901c4e99052968fb41c68b9f2ad2b52/SIL.Core.Tests/Xml/XmlUtilsTests.cs#L189)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

so based on the test failure this line looks like this in net standard
`class=\"bold\"\r\n\t\t\txmlns:fw=\"https://fakefw.com\"\r\n`
so there's a newline between `class="bold"` and xmlns, but not on framework. I debugged this and that's just how the code is written, I don't really see a good way around this. Even reordering the namespace doesn't make a difference.
This is because there's a special case for `xmlns` specifically. 

Looking into the source indenting/formatting works with via overloads. Internally `WriteStartNamespaceDeclaration` is called which is used only for `xmlns` in dotnet standard this method is overloaded in `XmlEncodedRawTextWriterIndent` to indent the line, and in framework it's not. Sadly these are internal so we can't just provide our own version. Here's the intended writer overloading [WriteStartAttribute](https://referencesource.microsoft.com/System.Xml/System/Xml/Core/XmlEncodedRawTextWriter.cs.html#b184fc7f471bdcb8), you can see there's no overload of `WriteStartNamespaceDeclaration`

I suspect this will be a problem as it will thrash the xml when run on one version of dotnet vs another. I'm not sure how, but if we could use the netstandard version in framework then it would resolve this issue.
</blockquote></details>

I don't need the xmlns right now, but since it would cause thrashing which is exactly what this utility is trying to prevent I introduced an exception.

jasonleenaylor (commented):
> Dismissed @hahn-kev from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1357)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #1359, opened 2024-11-07
Separated test runs by framework to avoid file-locking conflicts

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1359)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,12 +23,21 @@ jobs:
 
     # there are cases where this will fail and we want to know about it
     # so we don't use continue-on-error, but we still want to publish the results
-    - name: Test project
-      id: test
-      run: dotnet test -bl:test.binlog --filter "TestCategory != SkipOnTeamCity" --blame-hang-timeout 5m --logger:"trx;LogFilePrefix=results" --results-directory ./test-results
+    # Separate test steps for each framework to avoid file lock issues
+    - name: Test project (netcoreapp2.1)
+      id: test_netcoreapp21
+      run: dotnet test -bl:test_netcoreapp21.binlog --framework netcoreapp2.1 --filter "TestCategory != SkipOnTeamCity" --blame-hang-timeout 5m --logger:"trx;LogFilePrefix=results_netcoreapp21" --results-directory ./test-results/netcoreapp2.1
+
+    - name: Test project (net6.0)
+      id: test_net6
+      run: dotnet test -bl:test_net6.binlog --framework net6.0 --filter "TestCategory != SkipOnTeamCity" --blame-hang-timeout 5m --logger:"trx;LogFilePrefix=results_net6" --results-directory ./test-results/net6.0
+
+    - name: Test project (net8.0)
+      id: test_net8
+      run: dotnet test -bl:test_net8.binlog --framework net8.0 --filter "TestCategory != SkipOnTeamCity" --blame-hang-timeout 5m --logger:"trx;LogFilePrefix=results_net8" --results-directory ./test-results/net8.0
 
     - name: Publish test results
-      if: ${{ !cancelled() && steps.test.outcome != 'skipped' }}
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
*[`.github/workflows/build.yml` line 61 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1359#-OB7cDyFDv9H-475c8hW:-OB7cDyFDv9H-475c8hX:b-clmgwq) ([raw file](https://github.com/sillsdev/libpalaso/blob/fbef6b6d6f2e8589a43c5d4e77ff0aa6af8c26a2/.github/workflows/build.yml#L61)):*
> ```YAML
>         files: |
>           ./test-results/netcoreapp2.1/**/*.trx
>           ./test-results/net6.0/**/*.trx
> ```

I don't think this change is needed, the glob should pick up all 3 folders.

jasonleenaylor:
*[`.github/workflows/build.yml` line 25 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1359#-OB7cV7jEz0dbsw1z2cP:-OB7cV7jEz0dbsw1z2cQ:bq2rp0l) ([raw file](https://github.com/sillsdev/libpalaso/blob/fbef6b6d6f2e8589a43c5d4e77ff0aa6af8c26a2/.github/workflows/build.yml#L25)):*
> ```YAML
> 
>     - name: Build project - Debug
>       run: dotnet build -bl:build.dbg.binlog --configuration Debug
> ```

Curious why you wanted to have both a debug & release build?

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1359)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @hahn-kev, @jasonleenaylor, and @josephmyers)

___
*[`.github/workflows/build.yml` line 25 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1359#-OB7cV7jEz0dbsw1z2cP:-OB7oxIUBzpq4P59gBys:b-wb2mib) ([raw file](https://github.com/sillsdev/libpalaso/blob/fbef6b6d6f2e8589a43c5d4e77ff0aa6af8c26a2/.github/workflows/build.yml#L25)):*

Curious why you wanted to have both a debug & release build?
</blockquote></details>

I believe that's what it was doing before. It already did the release build upfront. But then each test run would rebuild it (in debug). But it seems this was maybe part of what was causing file locking issues since the test runs were apparently running in parallel. So now it does the debug build once upfront. Since there is the possibility of it successfully building in release but not in debug or vice versa, it seems safest to try both builds to be sure they both succeed. (Pretty sure most of our builds for other software does this as well.) I'm definitely kind of new to GHA and YAML, so if something I've said seems to contradict reality, feel free to challenge me.
___
*[`.github/workflows/build.yml` line 61 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1359#-OB7cDyFDv9H-475c8hW:-OB7pzwE819sKYEfQ3vQ:b-tiyuer) ([raw file](https://github.com/sillsdev/libpalaso/blob/fbef6b6d6f2e8589a43c5d4e77ff0aa6af8c26a2/.github/workflows/build.yml#L61)):*

I don't think this change is needed, the glob should pick up all 3 folders.
</blockquote></details>

I think you're right. I think the following commit proved to be the magic bullet.

josephmyers (changes requested):
I believe Kevin is going to leave a more complete review, but the most obvious thing here is that two of the frameworks are bogus. The fact that this passes is concerning. Apparently, false positives (silent failures) can happen if the projects are not marked with IsTestProject. I opened a new PR which utilizes the GHA matrix feature to separate things and marks our projects with the flag. (I needed a separate PR to fiddle with GHA silliness)

https://github.com/sillsdev/libpalaso/pull/1362

We also need a check to verify that our three supported frameworks actually have at least a certain number of tests run. We can do this in another task.

josephmyers:
I'm closing this in favor of using the concurrency built into GHA, #1362

PR #1361, opened 2024-11-08
Renamed and reimplemented GetFromRegistryProgramThatOpensFileType

tombogle:
Renamed GetFromRegistryProgramThatOpensFileType to GetDefaultProgramForFileType and implemented it in a way that works on Windows 11, Mono (probably) and MacOS (untested).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1361)


josephmyers (approved):
Note that, given the test project targets .NET8, you could easily test Linux (not Mono) if you have access to it. I think a one-time verification for this PR would be beneficial, if possible. Other than that, the code here looks good to me

tombogle:
I don't have a Linux machine. In some theoretical world, I could probably spin up a VM, but I doubt I'd ever got to it. I agree that it would be ideal.

PR #1363, opened 2024-11-08
Upgraded SIL.BuildTasks used in .proj file. Removed dependency on it from SIL.Windows.Forms

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1363)

### DIFF
--- a/SIL.Core.Desktop.Tests/Progress/Commands/CommandTests.cs
+++ b/SIL.Core.Desktop.Tests/Progress/Commands/CommandTests.cs
@@ -53,7 +53,7 @@ namespace SIL.Tests.Progress.Commands
 		public void CommandCallsFinalize()
 		{
 			TestCommand cmd = new TestCommand();
-			cmd.Finish += new EventHandler(OnFinish);
+			cmd.Finish += OnFinish;
 			cmd.BeginInvoke();
 			WaitOnBool(ref _onFinishCalled);
 		}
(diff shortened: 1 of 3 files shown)


josephmyers (changes requested):
I see a single failing test here that I haven't seen be flaky before. Have you looked into this? I wonder if that would alleviate the problem of all the missing tests.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1363)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tombogle)

tombogle:
> I see a single failing test here that I haven't seen be flaky before. Have you looked into this? I wonder if that would alleviate the problem of all the missing tests.

I didn't even bother looking. I figured I'd wait until the flaky test issue was resolved for good first.

tombogle:
> We can hope it gets fixed, but we're still seeing flaky tests, even when running each framework on its own machine. At the very least, my PR reduces server build time. It should also help with flakiness, but it won't go away. Some of these tests are just inherently flaky, unrelated to framework, e.g. testing the clipboard. In their case, we just need to mark them with `Retry`. I've done this for the one I'm seeing on my PR, but it'll be case by case, whether we can improve it or have to mark it with `Retry`.

Neither this text DLL nor the SUT reference SIL.BuildTasks, and this test passes for me locally, so it feels unlikely that it could be caused by this change. I did a tiny code simplification. Let's see if it magically passes this time.

tombogle:
After merging master, all the tests pass. However, I'm also seeing a similar failure in #1364, so I suspect it is another instance of flakiness. Not 100% sure, but it seems to be coming from tests of LocateInProgramFilesFolder. I would think that if we're running up against a permissions problem, it would be failing consistently. But those tests do seem at least somewhat susceptible to failures on build agents.

josephmyers (approved):

PR #1366, opened 2024-11-13
Net8windows keyboarding and writing systems

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1366)


josephmyers:
Note that there are two (by my count) lingering Framework dependencies in libpalaso, for the libraries that target .net8-windows. These appear to be pretty well ingrained and difficult to upgrade, though we do maintain both. They are Enchant.Net and ibusdotnet. Both of these are .NET wrapper libraries for assemblies we don't control. So, I think our major options are these: rewrite, replace, or ignore.

hahn-kev (approved):

PR #1370, opened 2025-01-06
Fixed problems in GetSafeDirectories and GetDirectoryDistributedWithApplication

tombogle:
In cases where a user does not have access to any one directory under consideration, these methods would fail/crash without considering the remaining directories.

- Made GetSafeDirectories not crash and simply not return any subdirectory the user does not have permission to access.
- In GetDirectoryDistributedWithApplication, prevented a failure in accessing one of the specified subfolders from allowing it to try the others.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1370)


SIL.Core/IO/DirectoryHelper.cs
```diff
@@ -106,21 +108,45 @@ public static bool IsEmpty(string path, bool onlyCheckForFiles = false)
...
+					var dirInfo = new DirectoryInfo(directoryName);
+					if ((dirInfo.Attributes & FileAttributes.System) != FileAttributes.System)
+					{
+						if ((dirInfo.Attributes & FileAttributes.Hidden) != FileAttributes.Hidden)
+							list.Add(directoryName);
+					}
+				}
+				catch (SecurityException e)
+				{
+					Console.WriteLine(e);
```
ermshiperete, line 145: I wonder if it would be better to use `Debug.WriteLine` or `Trace.WriteLine`.

SIL.Core/IO/FileLocationUtilities.cs
```diff
@@ -291,11 +291,24 @@ private static string LocateInProgramFilesFolder(string exeName, SearchOption sr
...
+						try
 						{
-							var tgtPath = GetFiles(exeName, srcOption, subDir);
-							if (!string.IsNullOrEmpty(tgtPath))
-								return tgtPath;
+							subDirectories = DirectoryHelper.GetSafeDirectories(path);
+						}
+						catch (Exception e)
+						{
+							Console.WriteLine(e);
```
ermshiperete, line 301: see above

SIL.Core/IO/FileLocationUtilities.cs
```diff
@@ -336,14 +349,18 @@ private static IEnumerable<string> GetPossibleProgramFilesFolders()
...
+				}
 				yield return "/opt"; // RAMP is installed in the /opt directory by default
 			}
 
-			var pf = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
-			yield return pf.Replace(" (x86)", string.Empty);
-			yield return pf.Replace(" (x86)", string.Empty) + " (x86)";
+			yield return Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
+			if (Environment.Is64BitProcess)
+				yield return Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86);
```
ermshiperete, line 363: :+1:

ermshiperete (approved):

SIL.Core/IO/DirectoryHelper.cs
tombogle: FWIW: There are 156 other instances of Console.Write(Line) in libpalaso and only a tiny handful of Trace.WriteLine calls. Maybe that makes sense, but it might bear a review. In this case, since it is not really a debug-only issue but something that could happen for real end-users, I don't think Debug.WriteLine is appropriate. If we hit this situation and the ende result is something undesirable (either a subsequent crash or failure or something the user reports as a potential bug), I think we would want the log file to reflect the error, so I'm thinking of changing these to Logger.WriteError instead.

SIL.Core/IO/FileLocationUtilities.cs
tombogle: Done.

SIL.Core/IO/FileLocationUtilities.cs
tombogle: Done.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1370)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @ermshiperete and @FoolRunning)

PR #1371, opened 2025-01-06
Added a unit test for XmlSerializationHelper.SerializeToString and did minor code cleanup

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`SIL.Core/Xml/XmlSerializationHelper.cs` line 62 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OIudZAO8EOiErZVUR0N:-OIudZAP9UqXZigGvgpQ:b-e8tkuv) ([raw file](https://github.com/sillsdev/libpalaso/blob/d14fd1bd0d369d1540938187a6eb267aed1980af/SIL.Core/Xml/XmlSerializationHelper.cs#L62)):*
> ```Smalltalk
> 			/// node on which the reader is positioned.
> 			/// </summary>
> 			/// <returns>The namespace URI of the current node; otherwise an empty string.</returns>
> ```

This description says it returns a namespace and the empty string is only a fallback, but it looks to me like it always returns a string.
___
*[`SIL.Core/Xml/XmlSerializationHelper.cs` line 142 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OIudiowCX5-M2v32pLb:-OIudiowCX5-M2v32pLc:b-e0mx5a) ([raw file](https://github.com/sillsdev/libpalaso/blob/d14fd1bd0d369d1540938187a6eb267aed1980af/SIL.Core/Xml/XmlSerializationHelper.cs#L142)):*
> ```Smalltalk
> 		/// </summary>
> 		public static byte[] SerializeToByteArray<T>(T data, bool omitXmlDeclaration)
> ```

I don't see why `[PublicAPI]` wasn't added here (nor on a number of other methods).

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @imnasnainaec and @tombogle)

___
*[`SIL.Core/Xml/XmlSerializationHelper.cs` line 62 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OIudZAO8EOiErZVUR0N:-OIvAnLa2jBjuqEwtB9K:b-sjv501) ([raw file](https://github.com/sillsdev/libpalaso/blob/d14fd1bd0d369d1540938187a6eb267aed1980af/SIL.Core/Xml/XmlSerializationHelper.cs#L62)):*

This description says it returns a namespace and the empty string is only a fallback, but it looks to me like it always returns a string.
</blockquote></details>

My change was merely a simple refactoring. This comment comes from the base class. It's not clear to me on a cursory inspection why this was overridden at all. My guess is that it maybe avoids an exception or other undesirable behavior in the case where we're dealing with a whitespace node that we're keeping but the underlying implementation wouldn't have kept, but that seems only barely plausible, and even then we could catch the exception and return Empty only in that case. Looking at the history, this code technically originated with me *in this file* but it was pulled in from somewhere else. It's an internal (private) class and none of the code that uses it references this property, so unless there's some historical reason that no longer applies, the only possible purpose I can think of for overriding it would be to keep it from crashing or maybe timing out in the debugger's watch window. All that said, I guess it probably makes sense to change the comment to be accurate.
___
*[`SIL.Core/Xml/XmlSerializationHelper.cs` line 142 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OIudiowCX5-M2v32pLb:-OIv9NDCCJ1EAVShssiJ:b-hh1cah) ([raw file](https://github.com/sillsdev/libpalaso/blob/d14fd1bd0d369d1540938187a6eb267aed1980af/SIL.Core/Xml/XmlSerializationHelper.cs#L142)):*

I don't see why `[PublicAPI]` wasn't added here (nor on a number of other methods).
</blockquote></details>

That's just a tip to Resharper not to flag the method as being unused, since it may be (and probably is) used by some client of the API.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @tombogle)

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)* status: 1 of 7 files reviewed, all discussions resolved (waiting on @tombogle)

___
*[`SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs` line 506 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OIw0fwcDbT5_T9CcwbH:-OIza_Bf-rGGJqkXhYjV:b-ddyv8l) ([raw file](https://github.com/sillsdev/libpalaso/blob/810c449bacbe70de19c42803436ea69fd9908983/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs#L506)):*

I find this easier to read, but that's very subjective:
```
			var source = new[] { "List Item 1", "List Item 2", "List Item 3" };
			var data = source.Select(
				s => new XmlTranslation { Reference = $"MAT {s.Last()}:6", PhraseKey = s, Translation = s.ToLower() });
			var result = SerializeToString(data.ToList(), Encoding.UTF8);
			var lines = result.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries).ToList();
```
</blockquote></details>

Me too. I was actually surprised to see in the history that I was the author of those lines because I have always found the from...select syntax in Linq very obtuse, and I make it a practice not to use it.
___
*[`SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs` line 517 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OIvzUti1F-RibxFDLCR:-OIzawYY5Hp2QL26po6P:b5i8kge) ([raw file](https://github.com/sillsdev/libpalaso/blob/810c449bacbe70de19c42803436ea69fd9908983/SIL.Core.Tests/Xml/XmlSerializationHelperTests.cs#L517)):*

Having element `<Translation>` inside type `<Translation [...]>` potentially confuses the example. Is that a realistic layout?
</blockquote></details>

Yes, it's based on some real (albeit unfortunate) data in Transcelerator.

imnasnainaec (commented):
> Reviewed 3 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 188 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-v9d4-s_DY9zhxsr3:-OJ-v9d4-s_DY9zhxsr4:b-ycvprg) ([raw file](https://github.com/sillsdev/libpalaso/blob/1a9e424a4a2979588d3543c292eeca81b8f25b24/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L188)):*
> ```Smalltalk
> 
> 		[Test]
> 		public void SplitLines_NoLineBreaks_ReturnsArrayWithSingleSting()
> ```

"...SingleSting" -> "...SingleString"
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 204 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-v51a2tnjh_P3aYKW:-OJ-v51a2tnjh_P3aYKX:b-rva4hx) ([raw file](https://github.com/sillsdev/libpalaso/blob/1a9e424a4a2979588d3543c292eeca81b8f25b24/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L204)):*
> ```Smalltalk
> 		[TestCase("line1\n\r\n\rline2")]
> 		[TestCase("line1\r\n\r\nline2")]
> 		public void SplitLines_LineBreaks_ReturnsArrayWithSingleSting(string str)
> ```

"...WithSingleSting" => "...WithTwoStrings"
___
*[`CHANGELOG.md` line 23 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-vOyn9XsIvGOK5-lb:-OJ-vOyn9XsIvGOK5-lc:b-fcsif3) ([raw file](https://github.com/sillsdev/libpalaso/blob/1a9e424a4a2979588d3543c292eeca81b8f25b24/CHANGELOG.md#L23)):*
> ```Markdown
> - [SIL.WritingSystems] Added public DownloadLanguageTags method for updating the cached langtags.json from the SLDR repository.
> - [SIL.Core] Add environment variable to disable `GlobalMutex` across processes. Helpful for snap packages in Linux.
> - [SIL.Core] Added String extension method SplitLines.
> ```

It seems to me "string extension" or "StringExtension" would be more accurate than "String extension".

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1371)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @imnasnainaec and @tombogle)

___
*[`CHANGELOG.md` line 23 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-vOyn9XsIvGOK5-lb:-OJ07wBC5Z-XkLrur-jY:b490tm4) ([raw file](https://github.com/sillsdev/libpalaso/blob/1a9e424a4a2979588d3543c292eeca81b8f25b24/CHANGELOG.md#L23)):*

It seems to me "string extension" or "StringExtension" would be more accurate than "String extension".
</blockquote></details>

I was using the uppercase to refer to the `System.String` class, which is essentially equivalent to the reserved word`string` (lowercase). but I'm good either way, so I've changed it to the lowercase version.
___
*[`SIL.Core/Extensions/StringExtensions.cs` line 154 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-MA4q6dBAYVbJsqv6:-OJ08__n7slwrFYghoGT:b-886gsx) ([raw file](https://github.com/sillsdev/libpalaso/blob/1da79a379aeaed4ea346a71a9932fb5f45964906/SIL.Core/Extensions/StringExtensions.cs#L154)):*

"as may parts" -> "as many parts"
</blockquote></details>

Fixed
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 201 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-MMx3-1V85k7glAGn:-OJ08dhr9Y_gRhy-fuGH:bb6un1j) ([raw file](https://github.com/sillsdev/libpalaso/blob/1da79a379aeaed4ea346a71a9932fb5f45964906/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L201)):*

Perhaps add some initial and/or final `\n` and `\r` to a few test cases.
</blockquote></details>

Done
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 188 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-v9d4-s_DY9zhxsr3:-OJ08beb9Z_jU3uh5khO:b-886gsx) ([raw file](https://github.com/sillsdev/libpalaso/blob/1a9e424a4a2979588d3543c292eeca81b8f25b24/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L188)):*

"...SingleSting" -> "...SingleString"
</blockquote></details>

Fixed
___
*[`SIL.Core.Tests/Extensions/StringExtensionTests.cs` line 204 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1371#-OJ-v51a2tnjh_P3aYKW:-OJ08fJk1sHO0hOwDgWA:bteqh0c) ([raw file](https://github.com/sillsdev/libpalaso/blob/1a9e424a4a2979588d3543c292eeca81b8f25b24/SIL.Core.Tests/Extensions/StringExtensionTests.cs#L204)):*

"...WithSingleSting" => "...WithTwoStrings"
</blockquote></details>

Improved and fixed

PR #1374, opened 2025-01-23
Changed build date in SILAboutBox to be computed using the last write time

tombogle:
Probably never matters for production builds, which start clean each time, but using the creation time makes the build date look wrong for local builds where the entry assembly is merely updated by the build and not re-created.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1374)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - [SIL.TestUtilities] Made FluentAssertXml classes use "Assert.That" so they can work in clients that use NUnit 4.
+- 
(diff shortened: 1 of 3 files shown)


josephmyers (approved):
This change makes sense, given the difference between these two properties

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1374#-:-OHLIQWTAPbM_fRHYGBT:b-mnau13)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tombogle)

josephmyers:
I'm going to keep rerunning AppVeyor til it passes. The failures appear to be different each time

PR #1375, opened 2025-01-27
Made FadingMessageWindow implement all UI logic on the UI thread

tombogle:
This fixes crashes like SP-2340
I have marked this as a breaking change because of the removal of protected members, but I know of no derived classes, and in practice, the existing implementation was such that these members would be unlikely to have been meaningful or helpful in a derived class anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1375)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - [SIL.TestUtilities] Made FluentAssertXml classes use "Assert.That" so they can work in clients that use NUnit 4.
+- [SIL.Windows.Forms] Removed protected members from FadingMessageWindow: MsgThread, MsgForm, Text, MsgPoint, and ShowForm. (The underlying implementation needed to be changed, and the existing implementation was such that these members would be unlikely to have been meaningful or helpful in a derived class anyway.)
+- [SIL.Windows.Forms] Added the stated requirement that FadingMessageWindow.Show be called on the UI thread.
 - 
 ### Fixed
 - [SIL.Windows.Forms] Changed build date in SILAboutBox to be computed using the last write time instead of creation time.
+- [SIL.Windows.Forms] Made FadingMessageWindow implement all UI logic on the main UI thread in a thread-safe way. Fixes crashes like SP-2340.
 
 ## [15.0.0] - 2025-01-06
 
--- a/SIL.Windows.Forms/ClearShare/WinFormsUI/FadingMessageForm.cs
+++ b/SIL.Windows.Forms/ClearShare/WinFormsUI/FadingMessageForm.cs
@@ -1,8 +1,8 @@
-﻿using System;
+using System;
 using System.Drawing;
 using System.Drawing.Drawing2D;
-using System.Threading;
 using System.Windows.Forms;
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -18,9 +18,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - [SIL.TestUtilities] Made FluentAssertXml classes use "Assert.That" so they can work in clients that use NUnit 4.
+- [SIL.Windows.Forms] Removed protected members from FadingMessageWindow: MsgThread, MsgForm, Text, MsgPoint, and ShowForm. (The underlying implementation needed to be changed, and the existing implementation was such that these members would be unlikely to have been meaningful or helpful in a derived class anyway.)
+- [SIL.Windows.Forms] Added the stated requirement that FadingMessageWindow.Show be called on the
```
hahn-kev, line 22: don't leave me hanging, should be called on what?

SIL.Windows.Forms/ClearShare/WinFormsUI/FadingMessageForm.cs
```diff
@@ -232,53 +226,50 @@ protected override void OnPaint(PaintEventArgs e)
...
-
-		protected string Text { get; set; }
-
-		protected Point MsgPoint { get; set; }
-
-	    /// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Shows a fading message with the specified text at the specified screen location.
+		/// </summary>
+		/// <remarks>Caller is responsible for ensuring this is called on the UI thread.</remarks>
```
hahn-kev, line 235: it's been a while since I did any winforms dev, but could we just use invoke to call `m_msgForm.Show()` as that's likely what requires the UI thread?

hahn-kev: I played with it and couldn't figure anything out. At least now if you call it from the wrong thread it just won't do anything. If we were to require a parent control in order to call `Show` that would make it impossible to call this incorrectly as we could internally use SafeInvoke on the parent which already exists

hahn-kev (approved):
works for me, though the popup message was not relative to the test app form which was surprising.

tombogle:
Yeah, I probably could have (and maybe should have) gotten it to pop up in a more sensible place, but for the test app I didn't care. I just wanted to make sure it appeared and faded as expected.

SIL.Windows.Forms/ClearShare/WinFormsUI/FadingMessageForm.cs
tombogle: The reason we can't invoke here is that we don't have any UI element (or a Synchroniz ationContext associated with the UI thread). But the called always should. And with a method called FadingMessageForm.Show, I think it is reasonable to just state the requirement. Microsoft has lots of things that must be called on the UI thread. I'm not so sure that failing to do so will do "nothing". On Linux, I think it would throw an exception. On Windows, it might or might not, but hopefully, there would be some deadlock or other annoying indication of the problem.

CHANGELOG.md
tombogle: Done.

hahn-kev (approved):

PR #1376, opened 2025-01-31
set Sldr http client user agent to include dll version

hahn-kev:
this will allow the ldml and langtags apis to determine which version to return based on the client version

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1376)


SIL.WritingSystems.Tests/SldrTests.cs
```diff
@@ -76,6 +76,13 @@ public void Setup()
 				Sldr.Initialize(true, Sldr.DefaultSldrCachePath);
 		}
 
+		[Test]
+		public static void CheckUserAgent()
+		{
+			Assert.That(Sldr.UserAgent, Does.Contain("SIL.WritingSystems Library")
+				.And.Contains(typeof(Sldr).Assembly.GetName().Version.ToString()));
```
imnasnainaec, line 83: This will be the first place in libpalaso with `.And.` in a test assert; would it be better to split this into two asserts for consistent testing style?

tombogle: I don't think that's a problem. As long as the intention of the Assert is one fundamentally related check.

tim-eves (approved):

imnasnainaec (approved):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1376)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

PR #1378, opened 2025-02-04
+semver:minor Add environment variable to disable `GlobalMutex` across processes

lyonsil:
On Linux, snap package confinement prevents reading and writing from many places in the system. That includes `/var/lock` which is used by GlobalMutex. We need a way to disable GlobalMutex in these situations. This PR provides an environment variable that can be set which makes GlobalMutex no longer global. It sill does some locking, though, in case something within the same process relies on a GlobalMutex for local locking.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [SIL.Core] Add environment variable to disable `GlobalMutex` across processes. Helpful for snap packages in Linux.
+
 ### Changed
 - [SIL.TestUtilities] Made FluentAssertXml classes use "Assert.That" so they can work in clients that use NUnit 4.
 - [SIL.Windows.Forms] Removed protected members from FadingMessageWindow: MsgThread, MsgForm, Text, MsgPoint, and ShowForm. (The underlying implementation needed to be changed, and the existing implementation was such that these members would be unlikely to have been meaningful or helpful in a derived class anyway.)
--- a/SIL.Core.Tests/Threading/GlobalMutexTests.cs
+++ b/SIL.Core.Tests/Threading/GlobalMutexTests.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Threading;
+using System;
 using NUnit.Framework;
 using SIL.Threading;
 
--- a/SIL.Core.Tests/Threading/GlobalMutexTests.cs
+++ b/SIL.Core.Tests/Threading/GlobalMutexTests.cs
@@ -93,30 +93,5 @@ namespace SIL.Tests.Threading
 				}
 			}
 		}
-
-		[Test, Timeout(1000)]
-		public void DisposingTheMutexReleasesTheLock()
-		{
-			using (var mutex = new GlobalMutex("test"))
-			{
-				mutex.Initialize();
-				//Explicitly not disposing the release
-				//which Lock returns to ensure that when the mutex is disposed, the lock is released
-				_ = mutex.Lock();
-			}
-
-			var otherThread = new Thread(() =>
-			{
-				using (var mutex = new GlobalMutex("test"))
-				{
-					//Should be able to take the lock again, threading is important here otherwise
-					//if the previous mutex still held the lock,
-					//this would be fine since they were the same thread
-					using (mutex.InitializeAndLock()) { }
-				}
-			});
-			otherThread.Start();
-			otherThread.Join();
-		}
 	}
 }
--- a/SIL.Core.Tests/Threading/GlobalMutexTests.cs
+++ b/SIL.Core.Tests/Threading/GlobalMutexTests.cs
@@ -118,5 +118,33 @@ namespace SIL.Tests.Threading
 			otherThread.Start();
 			otherThread.Join();
 		}
+
+		//this test might seem contrived, but if it was async, with an await inside the using statement,
+		//then the dispose/release may be executed on a different thread which would cause an exception.
+		//I used threads here to make that explicit
+		//as tasks are not guaranteed to be executed on a different thread
+		[Test]
+		public void TakingTheLockFromOneThreadAndReleasingItFromAnotherThreadShouldNotThrow()
+		{
+			using (var mutex = new GlobalMutex("test"))
+			{
+				var release = mutex.InitializeAndLock();
+				Exception exception = null;
+				var otherThread = new Thread(() =>
+				{
+					try
+					{
+						release.Dispose();
+					}
+					catch (Exception e)
+					{
+						exception = e;
+					}
+				});
+				otherThread.Start();
+				otherThread.Join();
+				Assert.That(exception, Is.Null);
+			}
+		}
 	}
 }
--- a/SIL.Core.Tests/Threading/GlobalMutexTests.cs
+++ b/SIL.Core.Tests/Threading/GlobalMutexTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading;
 using NUnit.Framework;
 using SIL.Threading;
 
@@ -91,6 +92,31 @@ namespace SIL.Tests.Threading
 					using (mutex.Lock()) {}
 				}
 			}
+		}
+
+		[Test, Timeout(1000)]
+		public void DisposingTheMutexReleasesTheLock()
+		{
+			using (var mutex = new GlobalMutex("test"))
+			{
+				mutex.Initialize();
+				//Explicitly not disposing the release
+				//which Lock returns to ensure that when the mutex is disposed, the lock is released
+				_ = mutex.Lock();
+			}
+
+			var otherThread = new Thread(() =>
+			{
+				using (var mutex = new GlobalMutex("test"))
+				{
+					//Should be able to take the lock again, threading is important here otherwise
+					//if the previous mutex still held the lock,
+					//this would be fine since they were the same thread
+					using (mutex.InitializeAndLock()) { }
+				}
+			});
+			otherThread.Start();
+			otherThread.Join();
 		}
 	}
 }
--- a/SIL.Core.Tests/Threading/GlobalMutexTests.cs
+++ b/SIL.Core.Tests/Threading/GlobalMutexTests.cs
@@ -1,11 +1,25 @@
-﻿using NUnit.Framework;
+﻿using System;
+using NUnit.Framework;
 using SIL.Threading;
 
 namespace SIL.Tests.Threading
 {
-	[TestFixture]
+	[TestFixture(true)]
+	[TestFixture(false)]
 	public class GlobalMutexTests
 	{
+		public GlobalMutexTests(bool localOnly)
+		{
+			if (localOnly)
+			{
+				Environment.SetEnvironmentVariable("SIL_CORE_MAKE_GLOBAL_MUTEX_LOCAL_ONLY", "true");
+			}
+			else
+			{
+				Environment.SetEnvironmentVariable("SIL_CORE_MAKE_GLOBAL_MUTEX_LOCAL_ONLY", null);
+			}
+		}
+
 		[Test]
 		public void Initialize_CreatedNew_ReturnsTrue()
 		{
@@ -54,7 +68,7 @@ namespace SIL.Tests.Threading
 			}
 		}
 
-		[Test]
+		[Test, Timeout(1000)]
 		public void InitializeAndLock_Reentrancy_DoesNotBlock()
 		{
 			using (var mutex = new GlobalMutex("test"))
@@ -66,7 +80,7 @@ namespace SIL.Tests.Threading
 			}
 		}
 
-		[Test]
+		[Test, Timeout(1000)]
 		public void Lock_Reentrancy_DoesNotBlock()
 		{
 			using (var mutex = new GlobalMutex("test"))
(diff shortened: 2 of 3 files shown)


hahn-kev (changes requested):
I pushed a change to the GlobalMutexTests so that they test this new version. I saw 3 failures. One was `InitializeAndLock_CreatedNew_ReturnsTrue` which only failed when running all the tests, when isolated it would pass.

SIL.Core/Threading/GlobalMutex.cs
```diff
@@ -34,7 +36,12 @@ public class GlobalMutex : DisposableBase
 		public GlobalMutex(string name)
 		{
 			_name = name;
-			if (Platform.IsWindows)
+			if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("SIL_CORE_MAKE_GLOBAL_MUTEX_LOCAL_ONLY")))
```
hahn-kev, line 39: this means if I set `SIL_CORE_MAKE_GLOBAL_MUTEX_LOCAL_ONLY=false` it will still be considered true here.

lyonsil: Yes, I did it that way because if you're setting this obnoxiously long environment variable, I figured you don't really care what the value is. You just want to cause the special behavior because you went to the trouble of setting something.

I updated the code to make it look more explicitly at the value since I think you're implying what I was doing isn't what you would expect.

lyonsil (commented):
Thanks - The tests should be passing now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378#-:-OIMIHqZD66yPu_OtaQu:bqe0glm)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

SIL.Core/Threading/GlobalMutex.cs
jasonleenaylor: We have used a convention before in our libraries that considered anything other than a case insensitive 'false' as on. I think that might be what Kevin is getting at.

https://github.com/sillsdev/libpalaso/blob/d237a40d6251b859cbd2845ead92fe8d2a7149d8/SIL.WritingSystems/Sldr.cs#L402

lyonsil: Thanks - It looks like the line following the one you linked suggests the value must be `true` or `yes` and is case insensitive. I think that aligns with what my code is now (though I also count "1" as `true`).

https://github.com/sillsdev/libpalaso/blob/d237a40d6251b859cbd2845ead92fe8d2a7149d8/SIL.WritingSystems/Sldr.cs#L403

lyonsil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @hahn-kev and @jasonleenaylor)

CHANGELOG.md
```diff
@@ -16,11 +16,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [SIL.Core] Add environment variable to disable `GlobalMutex` across processes. Helpful for snap packages in Linux.
```
tombogle, line 20: In the most technical sense, I suppose this should be regarded as a breaking change, but since it only changes the behavior if that environment variable is set and that seems pretty unlikely, I think it's okay to regard this as non-breaking. But since it is new functionality and not just a bug fix, I think we probably do need to bump up the minor version number. So there needs to be a commit with +semver:minor. (Of course, there's already been a major version bump up on master, so in practice I guess this would only matter if we cherry-picked and did a release of just this change.

SIL.Core/Threading/GlobalMutex.cs
```diff
@@ -34,7 +37,14 @@ public class GlobalMutex : DisposableBase
 		public GlobalMutex(string name)
 		{
 			_name = name;
-			if (Platform.IsWindows)
+			string envVarValue = Environment.GetEnvironmentVariable("SIL_CORE_MAKE_GLOBAL_MUTEX_LOCAL_ONLY");
```
tombogle, line 40: If I understand correctly, this is an intended fix to handle a situation that occurs only on Linux. If that is correct, did you consider perhaps making this special logi Linux-only?

tombogle (changes requested):

CHANGELOG.md
lyonsil: Right - I was going to do a minor bump, but when I saw we're already on a major bump beta, it didn't seem to matter. If you think I should do it anyway, let me know and I can change the commit message.

SIL.Core/Threading/GlobalMutex.cs
lyonsil: It only impacts Linux right now based on our testing. I can't say that it will only ever be Linux. If Windows and/or macOS decides to start sandboxing applications in the same way that snap does, then it could easily impact them. Since macOS is already *nix at its core, I'm guessing it would be more likely to hit there next.

Maybe just thinking about this in terms of desktop OS is not the right way to think, either. We're using a deprecated way of determining the OS, and the [newer way](https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem) has more options for things like Android. Android was built on Linux, but I'm not sure how all these determinations overlap.

Overall, since this is guarded by an environment variable, I think leaving it as OS-agnostic is probably safe and more future proof.

lyonsil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @hahn-kev, @lyonsil, and @tombogle)

CHANGELOG.md
tombogle: I don't feel super strongly about it. Since the SemVer takes care of determining whether/which version actually has to be bumped up, I try to make a practice of pushing the comment with the right +semver cue just as a matter of principle. It does provide a somewhat convenient way to scan the history to see where there were significant changes.

lyonsil: Ok - I added `+semver:minor` to the start of the commit message.

lyonsil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @hahn-kev, @lyonsil, and @tombogle)

tombogle:
*[`SIL.Core/Threading/GlobalMutex.cs` line 22 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1378#-OIRfq8_5IIgazC3F5bF:-OIRfq8a9T5k1O5FL-Mz:b-wgcoyg) ([raw file](https://github.com/sillsdev/libpalaso/blob/5e6831e7391d239862d153f436e9cc1f20171c50/SIL.Core/Threading/GlobalMutex.cs#L22)):*
> ```Smalltalk
> 	/// but even when using the constructors with names, it only works within a single process.
> 	///
> 	/// Note that in .NET 8.0 and later (and perhaps starting sooner than version 8), a named mutex can be used across
> ```

Please review this comment and adjust as needed. It seems as though (at least in the case of snaps) it is probably inaccurate or misleading. If this is true for snaps, then it makes me wonder if might be better to check the environment variable in the GlobalMutex constructor and create a named global mutex. It still wouldn't work with any older software using the file-based approach, but neither does the local-only approach, so going forward, it might be better.

tombogle (changes requested):
> Reviewed 1 of 1 files at r2, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @hahn-kev and @lyonsil)

lyonsil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @hahn-kev, @lyonsil, and @tombogle)

___
*[`SIL.Core/Threading/GlobalMutex.cs` line 22 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1378#-OIRfq8_5IIgazC3F5bF:-OIRnlOy2mEr5ffT5ZCU:bzctozr) ([raw file](https://github.com/sillsdev/libpalaso/blob/5e6831e7391d239862d153f436e9cc1f20171c50/SIL.Core/Threading/GlobalMutex.cs#L22)):*

Please review this comment and adjust as needed. It seems as though (at least in the case of snaps) it is probably inaccurate or misleading. If this is true for snaps, then it makes me wonder if might be better to check the environment variable in the GlobalMutex constructor and create a named global mutex. It still wouldn't work with any older software using the file-based approach, but neither does the local-only approach, so going forward, it might be better.
</blockquote></details>

I just added some more details to comments for this class. I didn't change this paragraph because what it says is true for applications running "normally" in Linux. It's only applications that run in confined environments like snaps where this becomes an issue.

tombogle (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: all files reviewed, 2 unresolved discussions (waiting on @hahn-kev)

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

tombogle (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378#-:-OIRsIQgBHgaIsH-Y_2w:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

SIL.Core.Tests/Threading/GlobalMutexTests.cs
```diff
@@ -78,5 +93,30 @@ public void Lock_Reentrancy_DoesNotBlock()
 				}
 			}
 		}
+
+		[Test, Timeout(1000)]
+		public void DisposingTheMutexReleasesTheLock()
```
hahn-kev, line 98: I wrote this test because I realized that disposing the mutex would not release the lock for the new case. This may be minor since all the lock methods return a disposable to represent the lock which should always be cleaned up.

hahn-kev (approved):
Pushed a test to reflect a failure case and fixed that case.

lyonsil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @hahn-kev and @tombogle)

___
*[`SIL.Core/Threading/GlobalMutex.cs` line 239 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1378#-OIkVALKBc4S8gSL18LY:-OIkVALKBc4S8gSL18LZ:b-r1rukn) ([raw file](https://github.com/sillsdev/libpalaso/blob/cbc8dc1e75e71004eba804365ed62546d2fbb7d9/SIL.Core/Threading/GlobalMutex.cs#L239)):*
> ```Smalltalk
> 			{
> 				Unlink();
> 				if (Monitor.IsEntered(_lock)) Monitor.Exit(_lock);
> ```

Since reentrant locking is required, this is not doing what you want it to do. Arguably changing `if` to `while` would handle that, but core .NET doesn't follow that pattern as it represents bad code. That's why [AbandonedMutexException](https://learn.microsoft.com/en-us/dotnet/api/system.threading.abandonedmutexexception) exists. Perhaps the right answer is to throw that exception here as it represents the same idea.

I will do some testing with different approaches if we want/need to catch this issue, but I'm confident that releasing on Dispose is not a good idea.

SIL.Core.Tests/Threading/GlobalMutexTests.cs
lyonsil: I don't think we should be adding this test case, or at least not as-is. Getting an `IDisposable` and explicitly not calling `Dispose` is an error/mistake in .NET overall, and trying to clean up afterwards for the mistake can just lead to other problems. If you want to keep a test for this coding pattern, then I would argue the correct result of the test would be to expect an `AbandonedMutexException` and fail if one is not received.

As long as `Dispose` is called implicitly (e.g., via `using`) or explicitly on the object returned by calling `GlobalMutex.Lock`/`GlobalMutex.InitializeAndLock`, then there is no problem with the new class. The more I think about this the more I think we probably should add a check for `Monitor.IsEntered` that throws an `AbandonedMutexException` if seen, but that's the extent to which I think we should do anything else with releasing locks here.

hahn-kev:
That makes sense to me, I was just trying to make sure it worked as close as possible to the standard implementation.

hahn-kev (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @tombogle)

tombogle (approved):
> Reviewed 2 of 2 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1378)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lyonsil)

PR #1382, opened 2025-02-06
protect against private use region codes in the regions field of a langtag

hahn-kev:
this should fix an issue that showed up with a new version of the langtags.json

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1382)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1382)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

___
*[`SIL.WritingSystems/LanguageLookup.cs` line 270 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1382#-OIQF-549Ek0ll-d38eU:-OIQF-549Ek0ll-d38eV:b-5r6xju) ([raw file](https://github.com/sillsdev/libpalaso/blob/f74a2c37dca7b1f978c39828910bd28c47b8b7e0/SIL.WritingSystems/LanguageLookup.cs#L270)):*
> ```Smalltalk
> 			}
> ```

I like this extracted function! We probably don't need multiple empty lines before its `return`.
___
*[`SIL.WritingSystems.Tests/LanguageLookupTests.cs` line 486 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1382#-OIQFgjSCNV27Y9cBUzR:-OIQFgjSCNV27Y9cBUzS:b-n4ew5t) ([raw file](https://github.com/sillsdev/libpalaso/blob/f74a2c37dca7b1f978c39828910bd28c47b8b7e0/SIL.WritingSystems.Tests/LanguageLookupTests.cs#L486)):*
> ```Smalltalk
> 			};
> 
> 			new LanguageLookup(entries, false);
> ```

It'd be nice to have an assert or two for checking some of the region/country cases. Though if you just want this test to not throw, then either `Does.Not.Throw`, or at least a comment to that effect, seems called for.

imnasnainaec (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1382)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

SIL.WritingSystems/LanguageLookup.cs
```diff
@@ -258,6 +251,24 @@ private bool AddLanguage(string code, string threelettercode, string full = null
 			return true;
 		}
 
+		private static string RegionToCountry(string region)
+		{
+			string country;
+			if (StandardSubtags.IsPrivateUseRegionCode(region))
+			{
+				country = region == "XK" ? "Kosovo" : "Unknown private use";
```
srl295, line 259: You might also consider leveraging cldr data here/ there are about 8 such codes cldr uses.

imnasnainaec: @srl295 Would you be up for opening an issue for that with a pointer to the relevant CLDR data?

PR #1386, opened 2025-02-07
Moved signed Keyman lib's to their own Nuget package

josephmyers:
This fixes a bug with the .NET8-Windows version, where the Keyman lib's are included in the .exe working directory, but never included in the deps.json. This prevents any app from loading them and launching.

You could theoretically remove the deps.json to allow the default .NET assembly loading behavior (e.g. look in the working directory), but the icu.net reference requires a deps.json in the consuming application. Also that probably isn't a very good solution anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1386)


josephmyers:
Yep, I just took the DLL's from the repo and packed them up into their own space in Nuget. This allows modern .NET dependency resolution to find them, as Reference's don't work

josephmyers:
> Additional thoughts/questions:

> Is the packaging of keyman documented in another Repo or Issue closer to the Keyman Repo? Or is it so legacy that this is not necessary in your opinion and all developers involved would know by looking at other similar pull requests? Are there also other SIL Repos that would benefit from the NuGet Package? Or are those old binaries just used here in this project?

I don't know where these interop DLL's came from. They're not affiliated directly with Keyman, though that is an SIL product. The last time they were touched was [by Eberhard](https://github.com/sillsdev/libpalaso/commit/24c1e9a3f1357784fa7bfe9b95e3614dc5c7e70c), who's since moved on to other projects but is still somewhat reachable. They're legacy enough that I don't want to mess with them any more than I have to -- I doubt there was much in the way of process/automation back when these were relevant. At the same time, replacing them would require a complete reimplementation, and Solid (at least) uses this code path. I don't see any other places in libpalaso that use the old Reference's, but Solid does. So when this is merged, we'll be able to rip out the one Keyman DLL sitting in its repo.

hahn-kev (approved):

PR #1389, opened 2025-02-11
+semver:major [SIL.WritingSystems] Add UI font role

### LINKED ISSUE #1212: Support New LDML Standard
Presently, `SIL.WritingSystems.LdmlDataMapper.ReadLdml` handles a [finite number of font types](https://github.com/sillsdev/libpalaso/blob/89668a769170657356500203c7f887ade3bc66d3/SIL.WritingSystems/LdmlDataMapper.cs#L75). Several LDML files in [SLDR](https://github.com/silnrsi/sldr) (including English) are being updated to include `<sil:font name="Noto Sans" types="ui" [ . . . ]/>`. We should be able to read them properly (at least without crashing—see [LT-21098](https://jira.sil.org/browse/LT-21098)).

@mhosken (2022-08-16) wrote:
[As per the documentation,](https://github.com/silnrsi/sldr/blob/master/doc/sil_namespace.md#fonts) the `@types` attribute may take any values, one of which is explicitly stated as being "ui".

imnasnainaec:
Resolves #1212 (implementing https://github.com/sillsdev/libpalaso/issues/1212#issuecomment-1282380561)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1389)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms.Keyboarding.Windows] In internal interface ITfInputProcessorProfileMgr, renamed method RegisterProfile parameter hklsubstitute to hklSubstitute.
 - [SIL.Windows.Forms.Reporting] In class ProblemNotificationDialog, renamed internal property \_reoccurenceMessage to \_reoccurrenceMessage.
 - [SIL.WritingSystems] In class LanguageLookup changed private method AddLanguage parameter threelettercode to threeLetterCode.
+- [SIL.WritingSystems] Updated langtags.json and ianaSubtagRegistry.txt in Resources.
 - [SIL.Xml] BREAKING CHANGE: In class XmlUtils, renamed method GetIndendentedXml to GetIndentedXml.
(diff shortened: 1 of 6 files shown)


imnasnainaec:
The update to `Roundtrip_LdmlFontRoles` makes it a true "Roundtrip_" test (it was really only a "Write_" test) and covers the new UI case. However, a new "Read_" test will be needed to handle the case "when we get a value that is not in the list".

imnasnainaec:
@ermshiperete For the roundtrip tests to work with XML files that have a font without any type specified, I thought it would make sense to change the code to default to `FontRoles.None` rather than `FontRoles.Default`. However, I'm not sure if there are usages of the library that would reasonably expect or rely on (e.g.) all write-outs some type specified on every font.

ermshiperete:
I don't know the answer to that question and I don't know where that class is used. You might want to ask on Slack.

imnasnainaec:
> I don't know the answer to that question and I don't know where that class is used. You might want to ask on Slack.

No responses in #commonlibraries; should I ask again in LT Software's #public or LT's #general?

ermshiperete (approved):

ermshiperete:
> No responses in #commonlibraries; should I ask again in LT Software's #public or LT's #general?

Everyone who cares should watch #commonlibraries; otherwise it's documented as breaking change in the changelog, so I think we're good.

imnasnainaec (commented):
> Reviewed 5 of 6 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #1400, opened 2025-02-25
Remove useless variable assignments

imnasnainaec:
Fixes roughly half of the 80+ "Useless assignment to local variable" code warnings: https://github.com/sillsdev/libpalaso/security/code-scanning?query=is%3Aopen+branch%3Amaster+rule%3Acs%2Fuseless-assignment-to-local

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)


imnasnainaec (commented):
> Dismissed @github-advanced-security[bot] from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 0 of 33 files reviewed, 1 unresolved discussion

___
*[`SIL.Core/ClearShare/OlacSystem.cs` line 42 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAVWnN-QgNmmh6A8t1:-OLAVWnOAhWEpgKkfgn0:b-eeqj0m) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core/ClearShare/OlacSystem.cs#L42)):*
> ```Smalltalk
> 				var roleCode = contributor.Attributes().Single(a => a.Name == s_nsOlac + "code").Value;
> 				if (!TryGetRoleByCode(roleCode, out var role))
> 					role = GetRoles().ElementAt(0);
> ```

This change is trying to implement what appears to have been the intended function of the code. The current line `Role role = GetRoles().ElementAt(0);` is a useless assignment, because it will be overwritten by `TryGetRoleByCode(roleCode, out role);` even if `TryGetRoleByCode` returns false and has `out` variable `null`.

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms] Made FadingMessageWindow implement all UI logic on the main UI thread in a thread-safe way. Fixes crashes like SP-2340.
 - [SIL.Windows.Forms] Made ContributorsListControl more threadsafe. Possibly fixes crashes like SP-2353 or at least makes them less likely.
 - [SIL.WritingSystems] Added check to Subtag.Equals to ensure two subtags being compared are of same derived type. This could potentially be a subtle breaking change in the unlikely event that someone was intentionally relying on the previous errant behavior.
+- [SIL.Core.ClearShare] BREAKING CHANGE: Public method LoadWorkFromXml could possibly return a contribution with null role; in those cases, now falls back to the first role in the list of roles in the doc.
```
tombogle, line 70: I'm not very sure this is desirable. If a contribution was added but no role was specified, I don't think we can safely default the role.

imnasnainaec: Perhaps not. And if nothing else, that decision is outside the main scope of this pr.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 15 of 35 files reviewed, 4 unresolved discussions (waiting on @imnasnainaec and @tombogle)

___
*[`SIL.Core/ClearShare/OlacSystem.cs` line 42 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAVWnN-QgNmmh6A8t1:-OLBE0V-91kbUUJFaNvF:ba6uoid) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core/ClearShare/OlacSystem.cs#L42)):*

@hatton, since you were the original author of this (in 2011), perhaps you can weigh in as to your actual intentions. I suspect the correct thing to do is remove the initial assignment of role, and then do TryGetRoleByCode(roleCode, out var role);
</blockquote></details>

It looks like his commit (https://github.com/sillsdev/libpalaso/commit/b7277687c74a9206109294b855cb8550510c4245) was migrating the code from SayMore.
___
*[`SIL.Core.Desktop/Settings/CrossPlatformSettingsProvider.cs` line 88 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAg35l8IhOFJRptuK5:-OLB2L4O9mHn8dhnqp5n:b-pod6sk) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core.Desktop/Settings/CrossPlatformSettingsProvider.cs#L88)):*

Seems like this deserves a comment to explain what side-effect we're trying to trigger.
</blockquote></details>

This `CheckForErrorsInSettingsFile` method already has a fair summary:
```
		/// <summary>
		/// This Settings Provider will automatically delete a corrupted settings file, silently.
		/// If you want to control when it does that, and get a message describing the problem
		/// so that you can tell the user, call this before anything else touches the settings.
		/// </summary>
```
Is an additional comment before `_ = SettingsXml` desired?
___
*[`SIL.Core.Desktop.Tests/i18n/StringCatalogTests.cs` line 96 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAgqRt0aJaGN9kClzW:-OLBD8_J6XGS17gbxd-N:b-omyx3v) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core.Desktop.Tests/i18n/StringCatalogTests.cs#L96)):*

Maybe worth adding a comment indicating that the constructor sets the "singleton." In reality, StringCatalog should probably be redesigned so that it throws an exception if the singleton is reassigned. (Then we'd need to add a static Dispose method or something to be used in tests, to reset the singleton to null.) As it is, it seems really bad stuff could happen if someone tried to use this without realizing that it silently trashes the existing singleton. That class probably should just be redesigned, but then it would break a bunch of legacy code that barely gets any love, so probably not worth the tears.
</blockquote></details>

`<remarks>` added to the `StringCatalog` constructors.
___
*[`SIL.Core.Desktop.Tests/Network/RobustNetworkOperation.Tests.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAlULj-yf1VHu-nF52:-OLB7PUc-NywGrOSW3ig:bu0vxu5) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core.Desktop.Tests/Network/RobustNetworkOperation.Tests.cs#L16)):*

http://hg.palaso.org/ no longer exists. It's not clear whether this test is intended to test the "throws" scenario or the "true if proxy is needed" scenario. Makes it kind of hard to run by hand. The name of the test should be improved to make the intention clear, and the URL should adjusted if needed. Maybe add some comments as well. I realize that this PR really isn't about that, so this needn't be considered blocking, but if we can figure out the intention, now would be as good a time as any to improve it.
</blockquote></details>

It throws when I try with `https://rancher.languagetechnology.org/` (which requires a WireGuard tunnel to access). So I renamed the test and added an assert, leaving the old url in case somebody else knows of a suitable replacement for a `ReturnsTrue` test.
___
*[`SIL.DictionaryServices.Tests/Lift/LexEntryFromLiftBuilderTests.cs` line 173 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAnWLs3BeiBI7xVXyL:-OLBDGC5AvAjDoq6-EDj:b-ovwaw7) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.DictionaryServices.Tests/Lift/LexEntryFromLiftBuilderTests.cs#L173)):*

Should we be asserting below that e.Senses[0], Is.EqualTo(s)?
</blockquote></details>

Good idea.
___
*[`SIL.Lift.Tests/Merging/LiftChangeDetectorTests.cs` line 151 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAoEEE12bC_mm5-xwT:-OLBDJEj7M_wFQyzTbsH:b-r9ck87) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Lift.Tests/Merging/LiftChangeDetectorTests.cs#L151)):*

Technically, this test does not check that the constructor does nothing, merely that it does not create the cache directory. In actual fact, it does do something: it creates an object. Might be worth renaming and/or adding verification step(s) to show that the returned object really is a hamstrung object. Since the constructor itself does not throw in this case, how is the caller to know that the object they got back is a clunker?
</blockquote></details>

Renamed and one clunker-ish `Assert` added.

tombogle:
*[`SIL.Core.Desktop.Tests/Network/RobustNetworkOperation.Tests.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAlULj-yf1VHu-nF52:-OLuGkTp6XFA2-AAGKY5:btlr43h) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core.Desktop.Tests/Network/RobustNetworkOperation.Tests.cs#L16)):*

It throws when I try with `https://rancher.languagetechnology.org/` (which requires a WireGuard tunnel to access). So I renamed the test and added an assert, leaving the old url in case somebody else knows of a suitable replacement for a `ReturnsTrue` test.
</blockquote></details>

In that case, this test could probably be renamed: DoHttpGetAndGetProxyInfo_404_ReturnsFalse. No need to run it by hand, I think. I assume the "by hand" case would be the scenario where a proxy is needed, so that it either returns true or throws. Maybe the rancher URL is a good example for that. If you set up the VPN, does it return true?

tombogle (changes requested):
> Reviewed 2 of 2 files at r3, 2 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 15 of 35 files reviewed, 3 unresolved discussions (waiting on @imnasnainaec)

___
*[`SIL.Core.Desktop/Settings/CrossPlatformSettingsProvider.cs` line 88 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAg35l8IhOFJRptuK5:-OLuFBEO7aKxFM2U5B2q:b-2hsivx) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core.Desktop/Settings/CrossPlatformSettingsProvider.cs#L88)):*

This `CheckForErrorsInSettingsFile` method already has a fair summary:
```
		/// <summary>
		/// This Settings Provider will automatically delete a corrupted settings file, silently.
		/// If you want to control when it does that, and get a message describing the problem
		/// so that you can tell the user, call this before anything else touches the settings.
		/// </summary>
```
Is an additional comment before `_ = SettingsXml` desired?
</blockquote></details>

I hadn't paid attention to the context. So I guess the side-effect we're after is setting _lastReadingError. Probably understandable as it, but a little comment to that effect wouldn't hurt.
___
*[`SIL.Lift/Merging/LiftChangeDetector.cs` line 73 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLuHbmhE5SjgSglYVJn:-OLuHbmiDkeA-lxXny_k:b-7qvkdq) ([raw file](https://github.com/sillsdev/libpalaso/blob/314ac8acf75bd639b90760cf89d7f3ebac55ed13/SIL.Lift/Merging/LiftChangeDetector.cs#L73)):*
> ```Smalltalk
> 
> 		/// <summary>
> 		/// Checks if both the Lift file and a cached reference copy exit.
> ```

exist

SIL.Core.Tests/Spelling/WordTokenizerTests.cs
```diff
@@ -13,12 +13,11 @@
...
-				 {
-					 foreach (WordTokenizer.Token t in WordTokenizer.TokenizeText(null))
-					 {
-
-					 }
-				 }
+				{
+					foreach (var _ in WordTokenizer.TokenizeText(null))
+					{
+					}
```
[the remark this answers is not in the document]
imnasnainaec: Done.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 14 of 35 files reviewed, 2 unresolved discussions (waiting on @imnasnainaec and @tombogle)

___
*[`SIL.Core.Desktop.Tests/Network/RobustNetworkOperation.Tests.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAlULj-yf1VHu-nF52:-OM7K6QV-CbSXWxIqgRh:bt0qiwz) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core.Desktop.Tests/Network/RobustNetworkOperation.Tests.cs#L16)):*

When I turn on the VPN it returns false, so we'd need another url for a `ReturnsTrue` scenario.
</blockquote></details>

With `"http://hg.palaso.org/"` it returns false for me locally but throws in our GHA workflows, so I returned that one to `Ignore("Run by hand")` and left the new `ReturnsFalse` test on `"https://sil.org"` without `Ignore`.

tombogle:
*[`SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs` line 344 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMCe-zq0RDBO8DDpp-e:-OMCe-zq0RDBO8DDpp-f:bgpwwwb) ([raw file](https://github.com/sillsdev/libpalaso/blob/527f284f09283c499a65baff283aa4c9d5f4cb4c/SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs#L344)):*
> ```Smalltalk
> 		protected virtual void OnDomDocumentCompleted(object sender, EventArgs ea)
> 		{
> 			_ = new GeckoDocumentCompletedEventArgs(ea);
> ```

Here, also, I think a comment would be helpful to know what side effect we're after.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 33 of 36 files reviewed, 2 unresolved discussions (waiting on @imnasnainaec, @StephenMcConnel, and @tombogle)

___
*[`SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs` line 344 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMCe-zq0RDBO8DDpp-e:-OMI0_Us1wCmL0OQj4Z2:b4o68mw) ([raw file](https://github.com/sillsdev/libpalaso/blob/527f284f09283c499a65baff283aa4c9d5f4cb4c/SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs#L344)):*

Here, also, I think a comment would be helpful to know what side effect we're after.
</blockquote></details>

I cannot see a purpose. @StephenMcConnel Do you know what the `new GeckoDocumentCompletedEventArgs` is for? It's part of 51e275c
___
*[`SIL.WritingSystems/Migration/WritingSystemsLdmlV0To1Migration/XmlHelpersV0.cs` line 243 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMClT5o6fP4kuZxgGEO:-OMHsnPp7kFa63LbfEMB:b-5b9l64) ([raw file](https://github.com/sillsdev/libpalaso/blob/527f284f09283c499a65baff283aa4c9d5f4cb4c/SIL.WritingSystems/Migration/WritingSystemsLdmlV0To1Migration/XmlHelpersV0.cs#L243)):*

I wonder if this check was supposed to be result == 0 instead of name == reader.Name? It would seem that they should be synonymous, but since a comparison method is being passed in, that might not be true. For example, if the passed-in comparison was doing a case-insensitive comparison, the two different methods of determining equality could give different results. In practice, looking at production uses, it appears that the "name" values being passed in are somewhat of a controlled vocabulary, so probably case distinctions would not occur and two strings that were different in any other way should produce a different value. But I'm still not 100% sure what logic would make the most sense.
</blockquote></details>

Given the `comparison(name, reader.Name) > 0` in the `while` loop and the presence of `int result = comparison(name, reader.Name);`, I thought using `result == 0` made more sense, then I found `SIL.Xml.XmlHelpers.FindNextElementInSequence`: https://github.com/sillsdev/libpalaso/commit/4c25530181a9efd067aa9f19d8b6963954ab54b4#diff-d690228068a257662f55483f00f1146212e724bae33cb94af98533e1e4edd7b5

tombogle (changes requested):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs` line 559 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMIAEdo9FQG7kc92UG6:-OMIAEdo9FQG7kc92UG7:b2nhpgn) ([raw file](https://github.com/sillsdev/libpalaso/blob/37d2fe2605bda9931c201f5168ad9a4575374b97/SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs#L559)):*
> ```Smalltalk
> 
> 		/// <summary>
> 		/// Constructor extracts Uri, Window, IsTopLevel properties from the EventArgs's type.
> ```

URI

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel and @tombogle)

___
*[`SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs` line 559 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMIAEdo9FQG7kc92UG6:-OMIZvPWBKdwKEC7XurX:bm4i6go) ([raw file](https://github.com/sillsdev/libpalaso/blob/37d2fe2605bda9931c201f5168ad9a4575374b97/SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs#L559)):*

URI
</blockquote></details>

Usually capitalized, I agree, but what about this case where the property in question is obtained with `.GetProperty("Uri")`?

tombogle:
*[`SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs` line 559 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMIAEdo9FQG7kc92UG6:-OMIbkFz2fYwrs47SskP:b-ajxqm0) ([raw file](https://github.com/sillsdev/libpalaso/blob/37d2fe2605bda9931c201f5168ad9a4575374b97/SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs#L559)):*

Usually capitalized, I agree, but what about this case where the property in question is obtained with `.GetProperty("Uri")`?
</blockquote></details>

Yeah, well, in that case, maybe we should use <see cref="...">

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel and @tombogle)

___
*[`SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs` line 559 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMIAEdo9FQG7kc92UG6:-OMN-dhy-YL518zlj8xb:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/2af4d5fe4c0241c610d96d94fb3e251c1b86e9b9/SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs#L559)):*

Yeah, well, in that case, maybe we should use <see cref="...">
</blockquote></details>

Done.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: all files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs` line 344 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMCe-zq0RDBO8DDpp-e:-OMNXnhc005may8rTYZq:b-1wpm5f) ([raw file](https://github.com/sillsdev/libpalaso/blob/527f284f09283c499a65baff283aa4c9d5f4cb4c/SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs#L344)):*

I cannot see a purpose. @StephenMcConnel Do you know what the `new GeckoDocumentCompletedEventArgs` is for? It's part of 51e275c
</blockquote></details>

I don't have any idea why the generic EventArgs was transformed this way unless the result was used in code that has since been deleted.  (possibly before being submitted 10 years ago)  I also don't understand why a typecast wasn't used instead of a new object creation.
I doubt seriously that there are any side-effects from the constructor.

hatton:
> @hatton, since you were the original author of this (in 2011), perhaps you can weigh in as to your actual intentions.

OK that's the funniest thing I've seen this week.

imnasnainaec:
> > @hatton, since you were the original author of this (in 2011), perhaps you can weigh in as to your actual intentions.
> 
> OK that's the funniest thing I've seen this week.

What? You don't remember all your actions and intentions from 14 years ago? 😄

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 35 of 36 files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs` line 344 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMCe-zq0RDBO8DDpp-e:-OMNecSS7OUAL-nDwILV:b-762dr1) ([raw file](https://github.com/sillsdev/libpalaso/blob/527f284f09283c499a65baff283aa4c9d5f4cb4c/SIL.Windows.Forms.GeckoBrowserAdapter/GeckoBase.cs#L344)):*

I don't have any idea why the generic EventArgs was transformed this way unless the result was used in code that has since been deleted.  (possibly before being submitted 10 years ago)  I also don't understand why a typecast wasn't used instead of a new object creation.
I doubt seriously that there are any side-effects from the constructor.
</blockquote></details>

@tombogle Removed, since there is no identifiable side effect.

tombogle:
*[`SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs` line 559 at r6](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OMIAEdo9FQG7kc92UG6:-OMNhUfF5B03vhJyTSaE:b-gbc57b) ([raw file](https://github.com/sillsdev/libpalaso/blob/6278108a5bd0fbd7dbdf7d72f820b80e946f278f/SIL.Windows.Forms.GeckoBrowserAdapter/ReflectiveGeckoBrowser.cs#L559)):*

Done.
</blockquote></details>

Apparently, my comment (which contained some XML) got obliterated by the do-not-put-literal-XML-in-your-comments-bot. I was suggesting using XML Docs reference elements. Though technically, the things you could reference that way wouldn't be the properties being extracted, but rather the properties to which those values are being assigned. Anyway, I don't think it really matters much.

tombogle (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: 34 of 36 files reviewed, all discussions resolved (waiting on @imnasnainaec)

tombogle (commented):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

tombogle:
*[`SIL.Core/ClearShare/OlacSystem.cs` line 42 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1400#-OLAVWnN-QgNmmh6A8t1:-OMNjiuI-p6UP095ihbh:bey5bax) ([raw file](https://github.com/sillsdev/libpalaso/blob/3d2168f9766adc950022a00a942ccbeed4676048/SIL.Core/ClearShare/OlacSystem.cs#L42)):*

It looks like his commit (https://github.com/sillsdev/libpalaso/commit/b7277687c74a9206109294b855cb8550510c4245) was migrating the code from SayMore.
</blockquote></details>

Studying this a little...

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

imnasnainaec (commented):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1400)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #1406, opened 2025-03-12
Upgraded SIL DLLs to latest stable versions

tombogle:
No longer referencing any beta packages (e.g., L10nSharp)
Upgraded MarkDig.Signed to meet dependency requirements
Updated other packages to latest bug fix patch version for current major/minor version

### DIFF
--- a/ExtractCopyright.Tests/ExtractCopyright.Tests.csproj
+++ b/ExtractCopyright.Tests/ExtractCopyright.Tests.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.2" />
     <PackageReference Include="GitVersion.MsBuild" Version="5.11.1" PrivateAssets="all" />
     <PackageReference Include="NUnit" Version="3.13.3" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.3.2" PrivateAssets="All" />
--- a/SIL.Archiving.Tests/SIL.Archiving.Tests.csproj
+++ b/SIL.Archiving.Tests/SIL.Archiving.Tests.csproj
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.2" />
     <PackageReference Include="GitVersion.MsBuild" Version="5.11.1" PrivateAssets="all" />
     <PackageReference Include="NUnit" Version="3.13.3" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.3.2" PrivateAssets="All" />
--- a/SIL.Archiving/SIL.Archiving.csproj
+++ b/SIL.Archiving/SIL.Archiving.csproj
@@ -21,11 +21,11 @@
 
   <ItemGroup>
     <PackageReference Include="GitVersion.MsBuild" Version="5.11.1" PrivateAssets="all" />
-    <PackageReference Include="JetBrains.Annotations" Version="2024.2.0">
+    <PackageReference Include="JetBrains.Annotations" Version="2024.3.0">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="SIL.ReleaseTasks" Version="2.5.0" PrivateAssets="All" />
+    <PackageReference Include="SIL.ReleaseTasks" Version="3.1.0" PrivateAssets="All" />
     <PackageReference Include="System.IO.Compression" Version="4.3.0" />
     <PackageReference Include="System.IO.Compression.ZipFile" Version="4.3.0" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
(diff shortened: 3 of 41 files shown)


tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1406)* status: 0 of 41 files reviewed, 2 unresolved discussions (waiting on @ermshiperete and @imnasnainaec)

___
*[a discussion](https://reviewable.io/reviews/sillsdev/libpalaso/1406#-OLAyrot1toNJ6IqzBwm:-OLFw6z51XPsVV4C6rj-:b1dss2r) (no related file):*

What determines which `.csproj` files require `Markdig.Signed` and which don't?
</blockquote></details>

The only thing in libpalaso that uses it directly (i.e., has it as a Top-level dependency) is SIL.Windows.Forms. Most of the other SIL.Windows.Forms.* DLLs reference that, so they have it as a transitive dependency. I just noticed that SIL.Media.Tests also references it, but the thing that uses it really should be a Test App, not stuck inside the DLL with the unit tests. I think I'll change that in a separate PR.
___
*[`SIL.Windows.Forms.Scripture.Tests/SIL.Windows.Forms.Scripture.Tests.csproj` line 17 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1406#-OLAxWeT2lPNY6dqMD1u:-OLFyUUUEVlEmDYcaV7t:bwjkfmf) ([raw file](https://github.com/sillsdev/libpalaso/blob/588214ab4e3bcb74cac28cd04e2cfcb359e3bd3d/SIL.Windows.Forms.Scripture.Tests/SIL.Windows.Forms.Scripture.Tests.csproj#L17)):*

Why add `Moq` to `SIL.Windows.Forms.Scripture.Tests` when it has no tests with `using Moq`?
</blockquote></details>

It is a transitive dependency there because SIL.Scripture.Tests references it, and SIL.Windows.Forms.Scripture.Tests\ScrPassageControlTests.cs uses TestScrVers, so it requires a reference to SIL.Scripture.Tests.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1406)* status: 0 of 41 files reviewed, 2 unresolved discussions (waiting on @ermshiperete and @imnasnainaec)

___
*[a discussion](https://reviewable.io/reviews/sillsdev/libpalaso/1406#-OLAyrot1toNJ6IqzBwm:-OLGV24y8NwEdNl571yv:b-2nfcuo) (no related file):*

The only thing in libpalaso that uses it directly (i.e., has it as a Top-level dependency) is SIL.Windows.Forms. Most of the other SIL.Windows.Forms.* DLLs reference that, so they have it as a transitive dependency. I just noticed that SIL.Media.Tests also references it, but the thing that uses it really should be a Test App, not stuck inside the DLL with the unit tests. I think I'll change that in a separate PR.
</blockquote></details>

This is the PR: https://github.com/sillsdev/libpalaso/pull/1408

imnasnainaec (commented):
> Reviewed 15 of 41 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1406)* status: 15 of 41 files reviewed, all discussions resolved (waiting on @ermshiperete)

tombogle:
@ermshiperete , Any idea why this is causing the Appveyor build to fail or what to do about it? The versions of System.Runtime.CompilerServices.Unsafe hasn't changed. I see there are some transitive dependencies on version 4.5.3 for some of our DLLs, though not for some that are experiencing test failures. I'm reticent to turn all these into top-level dependencies, even if I knew that would solve the problem. I could add binding redirects, but we have hitherto not needed any redirects for System.Runtime.CompilerServices.Unsafe and I'd really like to understand how this change could have suddenly introduce so many failures. all may package version changes were upgrades, so how could things suddenly want a much older version of System.Runtime.CompilerServices.Unsafe.

tombogle:
Wouldn't think so, since that has all been done and merged to master.

imnasnainaec:
@tombogle You could try rolling back the "Updated other packages to latest bug fix patch version for current major/minor version" changes...

tombogle:
Yeah, I can troubleshoot it by making the changes incrementally. I was just hoping someone would see the obvious answer and save me that trouble.

tombogle:
Here's what I note:

- L10nSharp is being referenced as Version="8.0.0-beta*", but since 8.00 has been released, the stable version is the version it picks up. So I can try removing the -beta* manually. Should still build and not make any difference.
- If I try to upgrade SIL.ReleaseTasks to Version 3.1.0, it complains about a downgrade of Markdig.Signed. I think that stems from SIL.BuildTasks having a transitive dependency on v. 0.30.3, while SIL.ReleaseTasks has a top-level dependency on 0.37.0. So in order to upgrade that, it seems I have to upgrade Markdig.Signed explicitly. I have created a PR on SIL.BuildTasks to try to consolidate the version there: https://github.com/sillsdev/SIL.BuildTasks/pull/74
- After making the above changes, I get two errors building: Could not copy the file "C:\Projects\libpalaso\output\Debug\net8.0\SIL.WritingSystems.Tests.runtimeconfig.json" because it was not found. and Could not copy the file "C:\Projects\libpalaso\output\Debug\net8.0\SIL.WritingSystems.Tests.deps.json" because it was not found. If I clean and then rebuild, those errors go away. I am vaguely suspicious that those errors might be related to the Appveyor test failures, but I can't draw any obvious connection.

tombogle:
https://github.com/sillsdev/libpalaso/pull/1410 demonstrates that removing -beta* for L10nsharp indeed is not the cause of the Appveyor problem. (That one can be merged independently.)

imnasnainaec (approved):
> Reviewed 3 of 41 files at r1, 1 of 31 files at r3, 39 of 39 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1406)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

PR #1416, opened 2025-03-25
Publish Nuget packages and symbols on a push to master

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1416)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,11 +87,14 @@ jobs:
         name: Test results (${{ matrix.framework }})
         path: ./test-results
 
+    # Limit the packaging to a single matrix job
+    # The dotnet build above does not specify the framework, so all artifacts are available on each job for packaging
     - name: Nuget Packaging
       id: nuget
       run: dotnet pack --include-symbols --no-restore --no-build -p:SymbolPackageFormat=snupkg --property:Platform="Any CPU" --output ./artifacts
       if: matrix.framework == 'net8.0-windows'
 
+    # You might want to test out the PR packages, but the release should be pulled from nuget
     - name: Upload Nuget Packages
       id: nuget-temp
       uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -87,6 +87,26 @@ jobs:
...
+        name: Nuget Packages
+        path: |
+          ./artifacts/*.nupkg
+          ./artifacts/*.snupkg
+      if: github.event_name == 'pull_request' && matrix.framework == 'net8.0-windows'
+          
+    - name: Publish Nuget Packages
+      id: nuget-publish
+      run: dotnet nuget push artifacts/*.nupkg -s https://api.nuget.org/v3/index.json -k ${{secrets.SILLSDEV_PUBLISH_NUGET_ORG}} --skip-duplicate
+      if: github.event_name == 'push' && matrix.framework == 'net8.0-windows'
```
hahn-kev, line 108: rather than `github.event_name == 'push'` and depend on this build only triggering on push for master I would make it explicit here that it should only build for master via `github.ref_name == 'master'`

.github/workflows/build.yml
```diff
@@ -87,6 +87,26 @@ jobs:
...
+
+    - name: Upload Nuget Packages
+      id: nuget-temp
+      uses: actions/upload-artifact@v4
+      with:
+        name: Nuget Packages
+        path: |
+          ./artifacts/*.nupkg
+          ./artifacts/*.snupkg
+      if: github.event_name == 'pull_request' && matrix.framework == 'net8.0-windows'
```
hahn-kev, line 103: is there a good reason not to also upload release packages as artifacts?

hahn-kev (approved):
looks good to me, a couple nitpicks.

.github/workflows/build.yml
jasonleenaylor: Arbitrary decision since I couldn't immediately think of a reason to use them as they would be published to Nuget. It's a 'Why keep something I don't know I need.' vs 'It doesn't cost anything but a bit of build time to keep it'

PR #1423, opened 2025-04-15
Fix invalid string formatting

imnasnainaec:
Fixes https://github.com/sillsdev/libpalaso/security/code-scanning/37
Fixes https://github.com/sillsdev/libpalaso/security/code-scanning/38

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1423)

### DIFF
--- a/SIL.Lift.Tests/Parsing/ParserTests.cs
+++ b/SIL.Lift.Tests/Parsing/ParserTests.cs
@@ -662,17 +662,16 @@ namespace SIL.Lift.Tests.Parsing
 			ExpectGetOrMakeEntry();
 			//ExpectMergeInLexemeForm(Is.Anything);
 			ExpectGetOrMakeSense();
-			string when= new DateTime(2000,1,1).ToUniversalTime().ToString(Extensible.LiftTimeFormatWithUTC);
 			ExpectMergeInTrait(t => new TraitMatcher("dummy", "blah", 2).Equals(t));
 			//ExpectMergeDefinition();
 
-			ParseEntryAndCheck(string.Format(@"
+			ParseEntryAndCheck(@"
 			<entry>
 				<sense>
 					<trait name='dummy' value ='blah'>
 						<annotation name='first'/>
 						<annotation name='second'/>
-				</trait></sense></entry>", when));
+				</trait></sense></entry>");
 		}
 
 		[Test]
--- a/SIL.Media/AlsaAudio/AlsaAudioDevice.cs
+++ b/SIL.Media/AlsaAudio/AlsaAudioDevice.cs
@@ -520,7 +520,7 @@ namespace SIL.Media.AlsaAudio
 			snd_pcm_hw_params_get_buffer_size(_hwparams, ref _bufferSize);
 			if (_chunkSize == _bufferSize)
 			{
-				ShowError(String.Format("Can't use period equal to buffer size (%lu == %lu)", _chunkSize, _bufferSize));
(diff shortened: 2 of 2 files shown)


tombogle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1423)* status: all files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`SIL.Media/AlsaAudio/AlsaAudioDevice.cs` line 523 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1423#-ONvPP_gEbLztJ7zZca9:-ONvPP_gEbLztJ7zZcaA:b-ct4ol6) ([raw file](https://github.com/sillsdev/libpalaso/blob/c352f975885a6a52121aaf0fec6c0cdddbd0cef1/SIL.Media/AlsaAudio/AlsaAudioDevice.cs#L523)):*
> ```Smalltalk
> 			if (_chunkSize == _bufferSize)
> 			{
> 				ShowError($"Can't use period equal to buffer size ({_chunkSize} == {_bufferSize})");
> ```

A clearer and more human-friendly message might be something like:

-   **"Can't use period equal to buffer size (both are 1092)."**
-   **"Chunk size and buffer size must differ (current value: 1092)."**
-   **"Invalid configuration: period and buffer size are both set to 1092."**

It may well be that we should actually be checking _chunkSize >= _bufferSize. ALSA uses the period size to determine how often it will notify the app that data needs to be written or read. If the period size equals the buffer size, then you only get one notification per buffer cycle, which kind of defeats the point of having periods at all. I'm not sure what happens if the period is greater than the buffer, but it might not end well. Still, if we're not sure, we might be best t leave the logic itself as is.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1423)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`SIL.Media/AlsaAudio/AlsaAudioDevice.cs` line 523 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1423#-ONvPP_gEbLztJ7zZca9:-ONyWgC20Brhqz5CVI9u:b-40yvce) ([raw file](https://github.com/sillsdev/libpalaso/blob/c352f975885a6a52121aaf0fec6c0cdddbd0cef1/SIL.Media/AlsaAudio/AlsaAudioDevice.cs#L523)):*

A clearer and more human-friendly message might be something like:

-   **"Can't use period equal to buffer size (both are 1092)."**
-   **"Chunk size and buffer size must differ (current value: 1092)."**
-   **"Invalid configuration: period and buffer size are both set to 1092."**

It may well be that we should actually be checking _chunkSize >= _bufferSize. ALSA uses the period size to determine how often it will notify the app that data needs to be written or read. If the period size equals the buffer size, then you only get one notification per buffer cycle, which kind of defeats the point of having periods at all. I'm not sure what happens if the period is greater than the buffer, but it might not end well. Still, if we're not sure, we might be best t leave the logic itself as is.
</blockquote></details>

Done. (... more comfortable leaving the logic as is.)

tombogle (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1423)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #1432, opened 2025-05-20
Changed build.yml to allow nuget-publish step to publish when build r…

tombogle:
…eferences a tagged version


jasonleenaylor (approved):
We could additionally modify this workflow to trigger a build on a pushed tag as well, but what is important is that we have a process for getting the release packages pushed and this will do that.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1432#-:-OQjAx7w8AJMJPsHnyDE:bd60e8g)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #1436, opened 2025-05-23
Changelog for the Keyman Legacy Bundle dependency

### LINKED ISSUE #1434: KeymanLegacyBundle dependency not documented in CHANGELOG

josephmyers:
Closes #1434

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1436)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Core] BREAKING CHANGE (potentially): Made ParseDateTimePermissivelyWithException into an extension method (on string).
 - [SIL.Core] BREAKING CHANGE (potentially): Changed behavior of DateTimeExtensions.ParseDateTimePermissivelyWithException (now deprecated) to try to interpret the date according to either the Gregorian calendar or the Buddhist calender in order to get the date to fall within a reasonable expected range (from 1/1/1900 through one day in the future). This means that depending on the current culture, dates might be interpreted differently from before. The known places in SIL code where this method is used seems to be for dates in the recent past (modern times); hence the default range. A new overload was added that will allow callers with other needs to specify a different range.
 - [SIL.Windows.Forms] BREAKING CHANGE: ImageToolbox Removed support for Linux due to no longer using DialogAdapters. Affects `OpenFileDialogWithViews`
+- [SIL.Windows.Forms.Keyboarding] BREAKING CHANGE: Moved Keyman legacy libraries to Nuget to fix runtime errors in modern .NET builds.
 
 ### Fixed
 


josephmyers:
I put this in the released version, but I can moved it to Unreleased, if desired. Also, @andrew-polk mentioned that this change was breaking. I either missed or forgot that, and how it was. I am happy to document this, if someone can help me improve the wording.

tombogle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1436)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`CHANGELOG.md` line 77 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1436#-OQvTDhvAyr9ZxoWrSxH:-OQvTDhwC7kjaR1bO_KK:b6cb6a6) ([raw file](https://github.com/sillsdev/libpalaso/blob/85b0749020e6b3f33838fbe847bb70ad849d3c59/CHANGELOG.md#L77)):*
> ```Markdown
> - [SIL.Core] BREAKING CHANGE (potentially): Changed behavior of DateTimeExtensions.ParseDateTimePermissivelyWithException (now deprecated) to try to interpret the date according to either the Gregorian calendar or the Buddhist calender in order to get the date to fall within a reasonable expected range (from 1/1/1900 through one day in the future). This means that depending on the current culture, dates might be interpreted differently from before. The known places in SIL code where this method is used seems to be for dates in the recent past (modern times); hence the default range. A new overload was added that will allow callers with other needs to specify a different range.
> - [SIL.Windows.Forms] BREAKING CHANGE: ImageToolbox Removed support for Linux due to no longer using DialogAdapters. Affects `OpenFileDialogWithViews`
> - [SIL.Windows.Forms.Keyboarding] Moved Keyman legacy libraries to Nuget to fix runtime errors in modern .NET builds.
> ```

This should indicate that this is a breaking change since it requires package references to be updated. According to Semantic Versioning (SemVer), moving DLLs from one package to a separate package is considered a breaking change.
Package consumers may rely on transitive dependencies being directly referenced. When you move the DLL, consumers who had build-time or runtime expectations (e.g., direct assembly references, reflection loading, or copying to output) may now experience failures or missing references if they don’t update their dependencies explicitly.
Restore/build behavior can change. Some consumers may have constraints or tooling that doesn’t handle transitive dependency resolution well, especially in older environments (e.g., .NET Framework projects with packages.config or legacy MSBuild behavior).

josephmyers:
Changed. I'm sure you're right, but keep in mind that the DLL and its output location is the same. Can consumers reference its place in the repo? Not normal ones. Anyone depending on those DLL's, given how they were included, would have had to be referencing where all the libpalaso and related lib's were being resolved to, even for direct referencing, copying to output, etc., and this didn't change.

I guess unless they set up their project to be dependent on the libpalaso source itself, i.e. in a particular folder, and if they happen to be referencing the DLL's where they were checked in, then this would break. I just feel bad for them.

PR #1447, opened 2025-07-11
Fix flaky DateTimeExtension tests

jasonleenaylor:
* Run the test code which changes the calendar in an isolated thread. NUnit can re-use a thread even when [NonParellelizable] is set so this technique should avoid any cross test complications

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1447)


jasonleenaylor:
I fixed up the incomplete commit comment in the automerge squash message.

PR #1458, opened 2025-09-02
Remove tests that only pass some days of the year

### LINKED ISSUE #1456: Failing tests in DateTimeExtensions
There were 10 test cases failing in DateTimeExtensions.
The master build on GHA went from passing to failing without any code changes, but the consistently fail now, including locally.

I commented them out for now.

imnasnainaec:
Closes #1456 

The reason those tests were passing and then were failing with no changes was because they were running from a different `DateTime.Today`. The test case `inputFormat` is not taken into account with the date-time parser, so it doesn't distinguish between d/M and M/d. The tests were only testing that the year was correctly parsed, so on some dates of the year, the test will pass when day is parsed as month and month is parsed as day, but not on all dates.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1458)

### DIFF
--- a/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs
+++ b/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs
@@ -73,7 +73,7 @@ namespace SIL.Tests.Extensions
 				TimeSeparator = "."
 			}
 		};
-		
+
 		[Test]
 		public void ToISO8601TimeFormatWithUTCString_CultureUsesPeriods_OutputsWithColons()
 		{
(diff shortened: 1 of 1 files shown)


tombogle:
I'll make a concerted effort at finding a way forward within the next week.

tombogle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1458)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @imnasnainaec)

___
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 367 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1458#-OZBJQ-uAuYnQozfBWvd:-OZBJQ-uAuYnQozfBWve:blojhig) ([raw file](https://github.com/sillsdev/libpalaso/blob/6683af41aaddc5ba9d7612ddc665e6743aeb9ada/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L367)):*
> ```Smalltalk
> 		{
> 			var reasonableMin = new DateTime(1481, 1, 1);
> 			var reasonableMax = DateTime.Today.AddDays(4);
> ```

What was the purpose in moving these? I wasn't 100% sure at first, but I've confirmed that the behavior is identical regardless of whether the calculation is done on the English/default version of Today or on the Thai version, so moving them is not a problem. (Using AddDays is a good little improvement.)

imnasnainaec (commented):
I comment in #1461 my concern with that approach. We can reopen this if we don't settle on a satisfactory solution.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1458#-:-OZPLs7J4tSa0ga8CuTS:bkpgzze)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 367 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1458#-OZBJQ-uAuYnQozfBWvd:-OZPLVn82dlYbSMj4Uzb:bygqlc6) ([raw file](https://github.com/sillsdev/libpalaso/blob/6683af41aaddc5ba9d7612ddc665e6743aeb9ada/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L367)):*

What was the purpose in moving these? I wasn't 100% sure at first, but I've confirmed that the behavior is identical regardless of whether the calculation is done on the English/default version of Today or on the Thai version, so moving them is not a problem. (Using AddDays is a good little improvement.)
</blockquote></details>

It's to match the location in (e.g.) `ParseDateTimePermissivelyWithException_NearFutureUSDatesWithThaiBuddhistCalendar_ReturnsDateWithFutureYear`. Also, I find it easier to parse the test when the `reasonableMax` and `futureDate` are defined side-by-side.

PR #1461, opened 2025-09-05
SP-2356: Fixed flaky DateTimeExtensionTests

tombogle:
The tests that were previously commented out because of their dependency on certain current dates should now be reliable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1461)

### DIFF
--- a/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs
+++ b/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs
@@ -361,15 +361,15 @@ namespace SIL.Tests.Extensions
 		[TestCase("dd/MM/yyyy")] // Thai/European-style numeric date, zero-padded (e.g. 14/05/2025)
 		[TestCase("d/M/yyyy")] // Thai/European-style numeric date (e.g. 14/5/2025)
 		[TestCase("d-M-yyyy")] // Thai/European-style numeric date with dashes (e.g. 14-5-2025)
-		// Commenting out these test cases for now until they can be fixed.
-		//[TestCase("d")] // Short date pattern (e.g. 14/5/2025)
-		//[TestCase("M/d/yyyy")] // US-style numeric date (e.g. 5/14/2025)
-		//[TestCase("M-d-yyyy")] // US-style numeric date with dashes (e.g. 5-14-2025)
-		//[TestCase("MM/dd/yyyy")] // US-style numeric date, zero-padded (e.g. 05/14/2025)
-		//[TestCase("MM-dd-yyyy")] // US-style numeric date with dashes, zero-padded (e.g. 05-14-2025)
+		[TestCase("d")] // Short date pattern (e.g. 14/5/2025)
+		[TestCase("M/d/yyyy")] // US-style numeric date (e.g. 5/14/2025)
+		[TestCase("M-d-yyyy")] // US-style numeric date with dashes (e.g. 5-14-2025)
+		[TestCase("MM/dd/yyyy")] // US-style numeric date, zero-padded (e.g. 05/14/2025)
+		[TestCase("MM-dd-yyyy")] // US-style numeric date with dashes, zero-padded (e.g. 05-14-2025)
 		public void ParseDateTimePermissivelyWithException_NearFutureDatesWithThaiBuddhistCalendar_ReturnsDateWithPastYear(string inputFormat)
 		{
-			var futureDate = DateTime.Today.AddDays(5);
+			var futureDate = GetNextFutureDateWithPotentiallyAmbiguousDay(5);
+			int maxDateOffset = (futureDate - DateTime.Today).Days - 1;
 			string input = futureDate.ToString(inputFormat);
 			int expectedYear = futureDate.Year - 543;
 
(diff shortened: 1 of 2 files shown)


imnasnainaec (commented):
> @imnasnainaec reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1461)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @tombogle)

___
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 421 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1461#-OZPL8N7CathdG-gw3E0:-OZcoBNCC1kRbrLZ4Fqc:bcafn79) ([raw file](https://github.com/sillsdev/libpalaso/blob/87946aa46d267e0e8d4f6666feb09f652a3d749d/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L421)):*

I assume you mean that it is *still* failing locally even with my proposed fix. If you break after the line that assigns `input`, what does that variable contain? Unless your current culture is set in such a way that it does not resolve to one of the explicit patterns already reflected in the other test cases, I don't see how "d" could fail without the corresponding explicit date pattern case also failing. I have now tweaked the verification logic to account for unexpected input formats due to unusual short date formats, but if its failing for you with a standard US short date, we need to figure out why/how.
</blockquote></details>

I'll try it again when I get to my Windows work computer. It's passing today on my Mac laptop.
___
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 428 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1461#-OZPK4qt0gkhDZP9-M6I:-OZcmZKXCusTDz9HXlrP:b-okhvbk) ([raw file](https://github.com/sillsdev/libpalaso/blob/87946aa46d267e0e8d4f6666feb09f652a3d749d/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L428)):*
That phrase was only referring back to my example that
> Somebody using the US-style numeric date may have future dates in the current year that will be parsed as past dates,

tombogle:
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 479 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1461#-OZdfrtqE14mPjky77rU:-OZdn0Il2JgNWR3yI5dl:b-r3zlrx) ([raw file](https://github.com/sillsdev/libpalaso/blob/85e33efec37f58f574bd7040f45e3c8ee2f0bffc/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L479)):*

I'm not sure this special handling is worthwhile. Do we need to keep the "d" case if it can never cause failure?
</blockquote></details>

I debated removing it, but decided it was probably still nice to have. It would get reported as a new ignored test if it started to "fail" in the CI build, so there's some chance it could get flagged for further investigation. And it's kind of nice to have around for local builds as well. (I thought about adding special logic to have it fail on local builds and ignore only in a CI build, but that seemed like overkill, and probably a local build is more likely to experience a spurious failure. It's usefulness is definitely borderline, but that's how I see it.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1461)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 510 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1461#-OZdxgob0pDj9ZDrQCxg:-OZdxgob0pDj9ZDrQCxh:b-sd1i5b) ([raw file](https://github.com/sillsdev/libpalaso/blob/85e33efec37f58f574bd7040f45e3c8ee2f0bffc/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L510)):*
> ```Smalltalk
> 		{
> 			var futureDate = GetNextFutureDateWithPotentiallyAmbiguousDay(5);
> 			int maxDateOffset = (futureDate - DateTime.Today).Days - 1;
> ```

Is there benefit in calculating `maxDateOffset` to be one day before the future date? Or can we simply have
```
			var maxDateOffset = 4;
			var futureDate = GetNextFutureDateWithPotentiallyAmbiguousDay(maxDateOffset + 1);

```

(Though I think it's more readible to have `reasonableMin`, `reasonableMax`, and `futureDate` all defined together at the start, I can understand conceptually defining `reasonableM*` with the test culture. )

tombogle:
*[`SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs` line 510 at r3](https://reviewable.io/reviews/sillsdev/libpalaso/1461#-OZdxgob0pDj9ZDrQCxg:-OZdztnP8p4LSuMl9eZB:bwm8c4k) ([raw file](https://github.com/sillsdev/libpalaso/blob/85e33efec37f58f574bd7040f45e3c8ee2f0bffc/SIL.Core.Tests/Extensions/DateTimeExtensionTests.cs#L510)):*

Is there benefit in calculating `maxDateOffset` to be one day before the future date? Or can we simply have
```
			var maxDateOffset = 4;
			var futureDate = GetNextFutureDateWithPotentiallyAmbiguousDay(maxDateOffset + 1);

```

(Though I think it's more readible to have `reasonableMin`, `reasonableMax`, and `futureDate` all defined together at the start, I can understand conceptually defining `reasonableM*` with the test culture. )
</blockquote></details>

I think either would be okay. I debated hard-coding `maxDateOffset` like that. For tightest prof that the code is "correct", using the more precise calculation of `maxDateOffset` is better, but if it's less clear then maybe it's worse.

PR #1471, opened 2025-10-13
Fix Chinese tag names in English for .NET 8

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1471)

### DIFF
--- a/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs
+++ b/SIL.WritingSystems.Tests/IetfLanguageTagTests.cs
@@ -948,7 +948,8 @@ namespace SIL.WritingSystems.Tests
 		[TestCase("noh", "en", ExpectedResult = "Nomu")]
 		[TestCase("noh", "noh", ExpectedResult = "Nomu")]
 		// See REVIEW comment in IEtfLanguageTag.FixBotchedNativeName
-		[TestCase("zh-CN", "zh-CN", "中文(中华人民共和国)", ExpectedResult = "中文(中国)")]
+		[TestCase("zh-CN", "zh-CN", "中文(中华人民共和国)", ExpectedResult = "简体中文")]
+		[TestCase("zh-TW", "zh-TW", ExpectedResult = "繁体中文")] // Perhaps this may need an acceptibleFallback
 		// Not sure if this fallback for Dari is truly acceptable. Although it looks the same and
 		// is what Windows supplies as the "Native Name", Wiktionary lists it as a different word
 		// and says that the word we have hardcoded in IetfLanguageTags.GetNativeNameIfKnown is the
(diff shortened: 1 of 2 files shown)


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1471)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`SIL.WritingSystems/IetfLanguageTag.cs` line 1499 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1471#-ObZad6o05OKN4SBx-6v:-ObZad6o05OKN4SBx-6w:b-20jsqr) ([raw file](https://github.com/sillsdev/libpalaso/blob/7ca82a8bb3955f5cb7c9ba2bf6113a8dfceb4473/SIL.WritingSystems/IetfLanguageTag.cs#L1499)):*
> ```Smalltalk
> 			default:
> 				return defaultOtherwise();
> 			}
> ```

Is this a better place to do it, have all the special cases in one place? Or is it silly since we still have 
`	if (ci.Name != ChineseSimplifiedTag && ci.Name != ChineseTraditionalTag)` on line 1433?

hatton:
Above, we have

> For consistency, there should be a newline between `}` and `else`.

Ain't got time for stuff like that. Please determine if prettier is running and fix if not.

andrew-polk (commented):
> Also, we know that the intentional Chinese names output here, "中文(中国)" and "中文 (台湾)" translate to Chinese (China) and Chinese (Taiwan), right? e.g. in "中文 (中国) (Chinese (Simplified))"

No, I don't think I knew that, anyway.
Maybe we should discuss. This has always been this way? We are hard-coding that, too?

> The difference in behavior between the unit test and Bloom is coming from CultureInfo.GetCultureInfo("zh-CN") returning different data in the two situations. I don't understand enough to know why. The tests pass even if I hardcode UseICUForLanguageNames=false. Is it worth looking further into how to simulate the Bloom behavior in unit tests? Am I missing something? I could look into making a test run without ICU for other languages (I think we are short-circuiting it particularly for the Chinese case)

On the one hand, I do wish I understood why the unit tests didn't catch what ended up being a behavior change in Bloom. Clearly the purpose of the unit test was to do so.
On the other hand, I don't want to keep digging into a rabbit hole.
Probably worth a little more investigation; let's discuss.

> @andrew-polk reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1471#-:-Obi8euP-dUXsvOeX4TR:bwz8mix)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @nabalone)

___
*[`SIL.WritingSystems/IetfLanguageTag.cs` line 1499 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1471#-ObZad6o05OKN4SBx-6v:-Obi8Y0EEZI1hbR3yE2i:b-kos91j) ([raw file](https://github.com/sillsdev/libpalaso/blob/7ca82a8bb3955f5cb7c9ba2bf6113a8dfceb4473/SIL.WritingSystems/IetfLanguageTag.cs#L1499)):*
<details><summary><i>Previously, nabalone (Noel) wrote…</i></summary><blockquote>

Is this a better place to do it, have all the special cases in one place? Or is it silly since we still have 
`	if (ci.Name != ChineseSimplifiedTag && ci.Name != ChineseTraditionalTag)` on line 1433?
</blockquote></details>

This is probably better. Though it is all a mess... but I think we've improved things.

nabalone (commented):
I have fixed the Traditional and Simplified chinese autonyms to be consistent with elsewhere in libpalasso.

Just noting that we discussed and decided not to keep digging into the CultureInfo inconsistencies for now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1471#-:-OcHGksx9ensd-tCBhUi:b-mph318)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk (approved):
> @andrew-polk reviewed 1 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1471)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #1478, opened 2025-11-12
Move Metadata and LicenseInfo so they can be accessed independent of Windows Forms

aror92:
Winforms-independent functionality of Metadata, CreativeCommonsLicense, and CustomLicense is moved to the new classes MetadataBare, CreativeCommonsLicenseInfo, and CustomLicenseInfo in SIL.Core.Clearshare.

Metadata, CreativeCommonsLicense, and CustomLicense inherit from the Winforms-free versions.

Make LicenseInfo class independent of Windows Forms and move it to SIL.Core.Clearshare.
	- Move the FromXmp method to LicenseUtils and LicenseWithImageUtils to construct Winforms-independent and Winforms-dependent license types respectively.
	- Move the FromToken method to LicenseWithLogo, so it can return the Winforms-dependent license types. FromToken is only used in libpalaso tests and examples, and a Winforms-independent version of this method is not needed.
	- Move the GetImage method to the ILicenseWithImage interface, which is implemented by CreativeCommonsLicense and CustomLicense.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1478)

### DIFF
--- a/SIL.Core.Tests/ClearShare/MetadataCoreTests.cs
+++ b/SIL.Core.Tests/ClearShare/MetadataCoreTests.cs
@@ -23,6 +23,7 @@ namespace SIL.Tests.ClearShare
 		[SetUp]
 		public void Setup()
 		{
+			_mediaFile = new Image<Rgba32>(10, 10);
 			_tempFile = TempFile.WithExtension("png");
 			_mediaFile.Save(_tempFile.Path, new PngEncoder());
 			_outgoing = MetadataCore.CreateMetadataCoreFromFile(_tempFile.Path);
--- a/SIL.Windows.Forms.Tests/ClearShare/MetadataTests.cs
+++ b/SIL.Windows.Forms.Tests/ClearShare/MetadataTests.cs
@@ -22,6 +22,7 @@ namespace SIL.Windows.Forms.Tests.ClearShare
 		[SetUp]
 		public void Setup()
 		{
+			_mediaFile = new Bitmap(10, 10);
 			_tempFile = TempFile.WithExtension("png");
 			_mediaFile.Save(_tempFile.Path);
 			_outgoing = Metadata.FromFile(_tempFile.Path);


andrew-polk:
Not worth a lot of head time, but worth a question:
Is it better to release 17.0 before this goes in so clients can decide to take the l10nsharp breaking change without this breaking change? Or is it better to include two breaking changes together in one major version release?

Bloom, getting ready to start a full regression test pass before releasing, would probably prefer to do so on a version without this change. (We already have the l10nsharp change.) But we are currently just using the beta version, so I'm not sure it matters much.
Where it could potentially matter is if we needed to hot-fix 17.0. (But we don't really have a build system set up to do that anyway, do we??)

tombogle (changes requested):
> @tombogle reviewed 1 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1478)* status: 6 of 30 files reviewed, 7 unresolved discussions (waiting on @andrew-polk)

___
*[`SIL.Core/ClearShare/LicenseInfo.cs` line 6 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRnYpE-cAU7H1-2nB8:-OeRnYpE-cAU7H1-2nB9:bs5284y) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/LicenseInfo.cs#L6)):*
> ```Smalltalk
> namespace SIL.Core.ClearShare
> {
> 	/// <summary/>
> ```

Please enter a real summary.
___
*[`SIL.Core/ClearShare/CustomLicenseBare.cs` line 7 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRlb0U21Lcndwe9dti:-OeRlb0U21Lcndwe9dtj:b-wnwrel) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CustomLicenseBare.cs#L7)):*
> ```Smalltalk
> {
> 	/// <summary>
> 	/// CustomLicenseBare is a WindowsForms-free base version of the CustomLicense class. In order to be WindowsForms independent,
> ```

Ditto: summary should answer "what is this?"
___
*[`SIL.Core/ClearShare/CustomLicenseBare.cs` line 25 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRljsz-Wqg9sk-Zn1r:-OeRljsz-Wqg9sk-Zn1s:b-7nq3ya) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CustomLicenseBare.cs#L25)):*
> ```Smalltalk
> 		///<summary></summary>
> 		/// <remarks>
> 		/// Currently, we don't know the language of custom license strings, so we the ISO 639-2 code for undetermined, "und"
> ```

missing "use"
___
*[`SIL.Core/ClearShare/CustomLicenseBare.cs` line 54 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRnMaE-l0mwcsgU78f:-OeRnMaE-l0mwcsgU78g:b-ydsol0) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CustomLicenseBare.cs#L54)):*
> ```Smalltalk
> 		{
> 			if (!properties.ContainsKey("rights (en)"))
> 				throw new ApplicationException("A license property is required in order to make a  Custom License from metadata.");
> ```

extra space in "...make a  Custom..."
___
*[`SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs` line 8 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRkK7R9ZX2uDfI8lIh:-OeRkK7R9ZX2uDfI8lIi:b-n1yyf3) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs#L8)):*
> ```Smalltalk
> {
> 	/// <summary>
> 	/// CreativeCommonsLicenseBare is a WindowsForms-free base version of the CustomLicense class. In order to be WindowsForms independent,
> ```

I think this is supposed to say `CreativeCommonsLicense`.
Also, this "Summary" feels more like "Remarks". It gives lots of useful information about the relationship of this class to `CustomLicense` and its WinForms independency, but it fails to answer the "what is this?" question that a would-be consumer would probably expect to see answered in the Summary.
___
*[`SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRm1OP5B9PDGjOMrl2:-OeRm1OP5B9PDGjOMrl3:bk8d55) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs#L12)):*
> ```Smalltalk
> 	/// To include license images, use the CreativeCommonsLicense class in SIL.WindowsForms.Clearshare.
> 	/// </summary>
> 	public class CreativeCommonsLicenseBare : LicenseInfo
> ```

I also don't love "Bare" Since this is a breaking change anyway, could we not perhaps call this CreativeCommonsLicense and name the Winforms version `WindowsCreativeCommonsLicense` or `CreativeCommonsLicenseForWindows' or ``CreativeCommonsLicenseForWinForms', etc. (and similarly with the other "Bare versions). Or maybe better yet, since they bare in different namespaces, keep the class names the *same*. Consumers can just use the appropriate namespace.
___
*[`SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs` line 193 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRlvPY1azULFbua7LT:-OeRlvPZ2IV2hcinFkjL:b-ozttyb) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs#L193)):*
> ```Smalltalk
> 		/// <remarks>
> 		/// REVIEW: (asked by Hatton Oct 2016) Serialization by whom? Why not just use the url, which is the canonical form?
> 		/// Note that this does not include any qualifier (of which at the moment the one one is "igo", but who knows
> ```

typo: one one

tombogle:
*[`SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRm1OP5B9PDGjOMrl2:-OebGQfA2K8LDMFe-LbA:b-xpubm7) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs#L12)):*
<details><summary><i>Previously, aror92 (Ariel Ror.) wrote…</i></summary><blockquote>

What do you think of one of these names instead for the "Bare" version?: "CreativeCommonsLicenseText" or "CreativeCommonsLicenseInfo".

Jason put high priority on keeping things as much the same in the Winforms side as we can. I'm really wanting to keep the names in the Winforms versions as they were in order to minimize pain and confusion of the breaking change, and to break as few things as possible.

I had considered using the same names, but I think it would negatively impact code readability (e.g. which version is being used is harder to see). And it could add confusion, since there are plenty of instances where both namespaces are used in the same code.
</blockquote></details>

`CreativeCommonsLicenseInfo` works

jasonleenaylor:
*[`SIL.Core/ClearShare/LicenseUtils.cs` line 11 at r5](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OebVUYR1IrEV3XCQ2JO:-Oeqflbf0OCAg9KSonFk:beqcr3) ([raw file](https://github.com/sillsdev/libpalaso/blob/7dea421aef2243a781b3b4b5b633871ce30f89dc/SIL.Core/ClearShare/LicenseUtils.cs#L11)):*

I think it might be more intuitive for each concrete subclass of licenseInfo to implement its own static`TryCreateFromMetadata` method, and just call those in succession inside `FromXmp`
</blockquote></details>

@tombogle It has been particularly tricky to get this to work well in consuming classes in an intuitive way. We have tried a number of different static and inheritance options. We put a high value on making this as non-disruptive as possible on the existing consumers who do not need the classes formerly known as 'bare'. Is this a make or break issue for you?

aror92 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1478)* status: 7 of 31 files reviewed, 5 unresolved discussions (waiting on @andrew-polk, @jasonleenaylor, and @tombogle)

___
*[`SIL.Windows.Forms/ClearShare/Metadata.cs` line 76 at r7](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-Of0swnIC0n_lwLXft_r:-OfZq_r5-X_c0uIAk9tn:b-lvzj1p) ([raw file](https://github.com/sillsdev/libpalaso/blob/7ad02b341b235457c50a13fb635af3c0a40414a3/SIL.Windows.Forms/ClearShare/Metadata.cs#L76)):*

If this "new override" is intended to be selected based on the compile-time type of the variable, it really needs a clear comment to that effect. It's extremely subtle behavior, especially since it is textually identical to the method it is replacing. If, on the other hand, that is not the intent, and we really want to choose the correct method based on the runtime type of the object, then the base class should be marked virtual and this should be a normal override. (Even then a comment would probably be in order.)
But if the latter, does LoadProperties actually have to be static? Seems like it would make more sense for it to be a member, since it always gets an instance as its second parameter. Then no override would be needed at all.
</blockquote></details>

Thanks for catching this! Yes, it should be virtual & override. I've fixed that. Will add a comment and look into removing static from LoadProperties.
___
*[`SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-OeRm1OP5B9PDGjOMrl2:-OfZhRO-A6COgEpHxpqB:b-896fix) ([raw file](https://github.com/sillsdev/libpalaso/blob/5edc50c20f4b880399f336665e7019b8c0386bcf/SIL.Core/ClearShare/CreativeCommonsLicenseBare.cs#L12)):*

Agreed. I have offered two possible suggestions.
</blockquote></details>

Done.

tombogle:
*[`SIL.Windows.Forms/ClearShare/WinFormsUI/MetadataEditorControl.cs` line 76 at r10](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-Og30tcqDcHQl6022HdZ:-Og30l3Q9gW-Z_EmseAZ:bbc6myh) ([raw file](https://github.com/sillsdev/libpalaso/blob/d961c94f0c58eac96909a7d5394c7e8af36808d8/SIL.Windows.Forms/ClearShare/WinFormsUI/MetadataEditorControl.cs#L76)):*
> ```Smalltalk
> 
> 				_copyrightBy.Text = _metadata.GetCopyrightBy();
> 				if (_metadata.License!=null && (_metadata.License is ILicenseWithImage))
> ```

Can also be simplified



_Code snippet:_
```Smalltalk
if (_metadata.License is ILicenseWithImage licenseWithImage)
    _licenseImage.Image = licenseWithImage.GetImage();
```

tombogle (approved):
:lgtm:

> @tombogle reviewed 2 of 28 files at r1, 1 of 7 files at r3, 1 of 7 files at r6, 2 of 12 files at r7, 2 of 8 files at r8, 2 of 4 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1478#-:-Og42eqc6ZuCkISURpFG:bnfp4nl)* status: 19 of 31 files reviewed, all discussions resolved (waiting on @andrew-polk and @jasonleenaylor)

PR #1518, opened 2026-06-18
+semver:minor Add FileLocationUtilities.AdditionalSearchDirectories

tombogle:
## Summary

- Added `FileLocationUtilities.AdditionalSearchDirectories` (`IList<string>`) — a public static property, defaulting to an empty list, that apps can populate at startup with full directory paths.
- `GetFileDistributedWithApplication` now iterates those directories after all built-in search locations (`DistFilesFolderPath`, `DirectoriesHoldingFiles`), so they only fire as a fallback and existing behavior is unchanged when the list is empty.
- Added a NUnit test (`GetFileDistributedWithApplication_FileOnlyInAdditionalDirectory_FindsCorrectly`) that places a file exclusively in a temp directory registered via `AdditionalSearchDirectories` and asserts it is found.

**Typical usage at app startup:**
```csharp
FileLocationUtilities.AdditionalSearchDirectories.Add(
    FileLocationUtilities.GetDirectoryDistributedWithApplication("docs"));
```

## Test plan

- [x] New test passes on all three target frameworks (net48, net462, net8.0)
- [x] All existing `FileLocationUtilitiesTests` still pass
- [x] No regressions in `SIL.Core.Tests` build

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1518)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Installer] Added new package for common installer components. Initially, this includes a Privacy dialog and code to access the registry entries so users can opt out of analytics data collection.
 - [SIL.Core] Added PathUtilities.ParentDirectories extension method.
 - [SIL.Core] Added FileLocationUtilities.DistFilesFolderPath property.
+- [SIL.Core] Added FileLocationUtilities.AdditionalSearchDirectories property, allowing apps to register extra directories searched as a fallback by GetFileDistributedWithApplication.
 
 ### Fixed
 - [SIL.Windows.Forms] Updated ImageToolbox UI to consistently use "image" (not "picture").
--- a/SIL.Core.Tests/IO/FileLocationUtilitiesTests.cs
+++ b/SIL.Core.Tests/IO/FileLocationUtilitiesTests.cs
@@ -2,6 +2,7 @@ using System;
 using System.IO;
 using System.Threading.Tasks;
 using NUnit.Framework;
+using SIL.IO;
 using SIL.PlatformUtilities;
 using static SIL.IO.FileLocationUtilities;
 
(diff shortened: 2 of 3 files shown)


tombogle:
I have a use in mind for this, but having made this change, I'm at least a little dubious as to its usefulness vs. just having the application look for the file in the specified location, especially since the need for it (in my case) is only when running in the dev environment, not in the installed application. So if you're reviewing this, feel free to push back.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1518)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on tombogle).

___
*[`SIL.Core/IO/FileLocationUtilities.cs` line 167 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1518#-OvVFmlt-VNrTFqVZR9s:-OvVFmlt-VNrTFqVZR9t:b-4u48qv) ([raw file](https://github.com/sillsdev/libpalaso/blob/5d807cd51fad63fad9dc97137c45a8d14f8856c3/SIL.Core/IO/FileLocationUtilities.cs#L167)):*
> ```Smalltalk
> 			}
> 
> 			foreach (var additionalDir in AdditionalSearchDirectories)
> ```

[Devin](https://app.devin.ai/review/sillsdev/libpalaso/pull/1518) notes:

**AdditionalSearchDirectories not used in GetDirectoryDistributedWithApplication**

The new `AdditionalSearchDirectories` is only consulted in `GetFileDistributedWithApplication` but not in the parallel method `GetDirectoryDistributedWithApplication`. The CHANGELOG entry explicitly scopes this to `GetFileDistributedWithApplication`, however, callers who register additional search directories might reasonably expect directory lookups to also search those paths. Worth confirming this asymmetry is deliberate.

imnasnainaec:
> I have a use in mind for this, but having made this change, I'm not at least a little dubious as to its usefulness vs. just having the application look for the file in the specified location, especially since the need for it (in my case) is only when running in the dev environment, not in the installed application. So if you're reviewing this, feel free to push back.

@tombogle Your commented motivation is using submodules. I asked Claude to consider other scenarios where this fix would be useful.
> **Plugin/extension host apps** — the most compelling one, and the only case that justifies the property being a list rather than a single string. If an app dynamically loads plugins, each plugin can register its own distributed files directory at load time. The existing `DistFilesFolderPath` single override can't handle that; `AdditionalSearchDirectories` can.

tombogle:
*[`SIL.Core/IO/FileLocationUtilities.cs` line 167 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1518#-OvVFmlt-VNrTFqVZR9s:-OvuG58r7P6IKGfAADWt:brwosqi) ([raw file](https://github.com/sillsdev/libpalaso/blob/5d807cd51fad63fad9dc97137c45a8d14f8856c3/SIL.Core/IO/FileLocationUtilities.cs#L167)):*

[Devin](https://app.devin.ai/review/sillsdev/libpalaso/pull/1518) notes:

**AdditionalSearchDirectories not used in GetDirectoryDistributedWithApplication**

The new `AdditionalSearchDirectories` is only consulted in `GetFileDistributedWithApplication` but not in the parallel method `GetDirectoryDistributedWithApplication`. The CHANGELOG entry explicitly scopes this to `GetFileDistributedWithApplication`, however, callers who register additional search directories might reasonably expect directory lookups to also search those paths. Worth confirming this asymmetry is deliberate.
</blockquote></details>

This is deliberate

tombogle:
I'm not feeling great about this change. I think an application-specific solution is probably adequate.

PR #1522, opened 2026-06-30
SettingsProtectionHelper: support always-hidden controls

### LINKED ISSUE #1134: When Settings protections is turned on with password protection, need a way to keep protected controls hidden
Obviously, we need for the user to be able to press Shift+Ctrl to make it possible to get back into the settings dialog to enter the password, but _other_ controls that are hidden via a call to SetSettingsProtection should remain hidden unless the password is entered. Otherwise, the password can really only be used to protect the settings dialog itself. Probably all that's needed is an additional bool parameter to indicate whether a ToolStripMenuItem should remain hidden if the settings are password protected.

imnasnainaec:
## Summary

- Add `SetSettingsProtection(Control, bool, bool keepHidden)` and `SetSettingsProtection(ToolStripItem, bool, bool keepHidden)` overloads so a control or ToolStripItem can be marked always-hidden, staying hidden even while Ctrl+Shift reveals the other protected components. The original 2-parameter overloads are preserved for WinForms designer extender-property discovery and binary compatibility.
- Track always-hidden components in a separate `_alwaysHiddenComponents` set, which `UpdateDisplay()` hides unconditionally after the normal pass. `GetSettingsProtection` reports `true` for them too.
- Apply always-hidden registration immediately at run time instead of leaving the component visible until the next (~100 ms) timer tick. Design time is deliberately excluded: the extender property setter runs there as well, where the timer is off and the component must stay visible on the design surface.
- Apply the normal visibility rule immediately when a component leaves always-hidden, so the two directions are symmetric. The `!NormallyHidden || Ctrl+Shift` rule moved into a property shared with `UpdateDisplay()`, which had been recomputing it once per component.
- Rework `Dispose(bool)` into the standard form used elsewhere in the repo — the missing-Dispose debug warning, an early return that makes it idempotent, managed cleanup only when `disposing`, and `_isDisposed` set *after* `base.Dispose(disposing)`. That ordering matters: `Component.Dispose` removes itself from its container, which reads the `Site` property, and this class overrides `Site` to throw once disposed. The explicit `_checkForCtrlKeyTimer.Dispose()` is gone, since the timer is owned by the `components` container that disposes it.
- Give both new overloads full XML documentation (summary plus a tag for every parameter).
- Add `SettingsProtectionHelperTests` — 11 tests covering normal vs. always-hidden visibility, both transition directions and their immediacy, unprotecting, `GetSettingsProtection`, and disposing while still sited in a container.
- Add CHANGELOG entries under Added and Fixed.

## Scope note on #1134

`keepHidden` is a "never show this" primitive: an always-hidden component is not revealed by Ctrl+Shift, and it is not revealed by entering the settings password either — nothing records that the password was entered, and `UpdateDisplay()` re-hides these components on every tick.

#1134 asks for components that remain hidden *unless the password is entered*. A client can build that on top of this: mark the components `keepHidden: true`, then re-register them with `keepHidden: false` after a successful challenge. This PR does not add that behavior to the library itself.

Fixes #1134

Devin review: https://app.devin.ai/review/sillsdev/libpalaso/pull/1522

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1522)

### DIFF
--- a/SIL.Windows.Forms.Tests/SettingsProtection/SettingsProtectionHelperTests.cs
+++ b/SIL.Windows.Forms.Tests/SettingsProtection/SettingsProtectionHelperTests.cs
@@ -112,5 +112,14 @@ namespace SIL.Windows.Forms.Tests.SettingsProtection
 
 			Assert.That(control.Visible, Is.True);
 		}
+
+		[Test]
+		public void GetSettingsProtection_AlwaysHiddenControl_ReturnsTrue()
+		{
+			var control = new Button();
+			_helper.SetSettingsProtection(control, true, keepHidden: true);
+
+			Assert.That(_helper.GetSettingsProtection(control), Is.True);
+		}
 	}
 }
--- a/SIL.Windows.Forms.Tests/SettingsProtection/SettingsProtectionHelperTests.cs
+++ b/SIL.Windows.Forms.Tests/SettingsProtection/SettingsProtectionHelperTests.cs
@@ -1,0 +1,116 @@
+using System.Reflection;
+using System.Windows.Forms;
+using NUnit.Framework;
+using SIL.Windows.Forms.SettingProtection;
+
+namespace SIL.Windows.Forms.Tests.SettingsProtection
+{
+	[TestFixture]
+	[Apartment(System.Threading.ApartmentState.STA)]
+	public class SettingsProtectionHelperTests
+	{
+		private SettingsProtectionHelper _helper;
+		private bool _savedNormallyHidden;
+
+		[SetUp]
+		public void SetUp()
+		{
+			_savedNormallyHidden = SettingsProtectionSingleton.Settings.NormallyHidden;
+			_helper = new SettingsProtectionHelper(null);
+		}
+
+		[TearDown]
+		public void TearDown()
+		{
+			SettingsProtectionSingleton.Settings.NormallyHidden = _savedNormallyHidden;
+			_helper.Dispose();
+		}
+
+		private static void CallUpdateDisplay(SettingsProtectionHelper helper)
+		{
+			var method = typeof(SettingsProtectionHelper).GetMethod("UpdateDisplay",
+				BindingFlags.NonPublic | BindingFlags.Instance);
+			method.Invoke(helper, null);
+		}
+
+		[Test]
+		public void UpdateDisplay_NormalProtectedControl_IsVisibleWhenNotNormallyHidden()
+		{
+			SettingsProtectionSingleton.Settings.NormallyHidden = false;
+			var control = new Button { Visible = false };
+			_helper.SetSettingsProtection(control, true);
+
+			CallUpdateDisplay(_helper);
+
+			Assert.That(control.Visible, Is.True);
+		}
+
+		[Test]
+		public void UpdateDisplay_NormalProtectedControl_IsHiddenWhenNormallyHidden()
+		{
+			SettingsProtectionSingleton.Settings.NormallyHidden = true;
+			var control = new Button { Visible = true };
+			_helper.SetSettingsProtection(control, true);
+
+			CallUpdateDisplay(_helper);
+
+			Assert.That(control.Visible, Is.False);
+		}
+
+		[Test]
+		public void UpdateDisplay_AlwaysHiddenControl_RemainsHiddenWhenNotNormallyHidden()
+		{
+			// Even when NormallyHidden=false (i.e., normal controls are visible),
+			// an always-hidden control must stay hidden.
+			SettingsProtectionSingleton.Settings.NormallyHidden = false;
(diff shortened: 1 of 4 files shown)


SIL.Windows.Forms.Tests/SettingsProtection/SettingsProtectionHelperTests.cs
```diff
@@ -0,0 +1,125 @@
...
+			var method = typeof(SettingsProtectionHelper).GetMethod("UpdateDisplay",
+				BindingFlags.NonPublic | BindingFlags.Instance);
+			method.Invoke(helper, null);
+		}
+
+		[Test]
+		public void UpdateDisplay_NormalProtectedControl_IsVisibleWhenNotNormallyHidden()
+		{
+			SettingsProtectionSingleton.Settings.NormallyHidden = false;
+			var control = new Button { Visible = false };
```
papeh: use `using` to dispose `control` in each test

SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.Designer.cs
```diff
@@ -15,17 +15,18 @@ partial class SettingsProtectionHelper
 		/// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
 		protected override void Dispose(bool disposing)
 		{
-			if (disposing && (components != null))
+			if (disposing)
 			{
-				components.Dispose();
+				if (components != null)
+					components.Dispose();
+				if (!_isDisposed)
```
papeh: Typically, the form for `Dispose(bool)` is
```
System.Diagnostics.Debug.WriteLineIf(!disposing, "****** Missing Dispose() call for " + GetType().Name + ". ****** ");
if (_isDisposed) return;
if (disposing) { ... }
...
_isDisposed = true;
```
Should this follow the same form?

papeh (changes requested):

imnasnainaec:
Pausing work on this for #1534 to clear.

SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.Designer.cs
imnasnainaec: Done.

SIL.Windows.Forms.Tests/SettingsProtection/SettingsProtectionHelperTests.cs
imnasnainaec: Done.

imnasnainaec (commented):
> @imnasnainaec made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1522)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on imnasnainaec and papeh).

papeh (commented):
> @papeh reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1522)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on imnasnainaec).

___
*[`SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs` line 77 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1522#-OzJ-T0EEolOZqQ1SHpR:-OzJ-T0EEolOZqQ1SHpS:bgo5tah) ([raw file](https://github.com/sillsdev/libpalaso/blob/da28f0222063549710904c378f1395b015972bd6/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L77)):*
> ```Smalltalk
> 			foreach (var component in _componentsUnderSettingsProtection)
> 			{
> 				bool visible = !SettingsProtectionSingleton.Settings.NormallyHidden || ((Control.ModifierKeys & keys) == keys);
> ```

nice improvement!

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1522)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on papeh).

___
*[`SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs` line 77 at r1](https://reviewable.io/reviews/sillsdev/libpalaso/1522#-OzJ-T0EEolOZqQ1SHpR:-OzME0HcCVORMxKYshTY:b-ucfknh) ([raw file](https://github.com/sillsdev/libpalaso/blob/da28f0222063549710904c378f1395b015972bd6/SIL.Windows.Forms/SettingProtection/SettingsProtectionHelper.cs#L77)):*

nice improvement!
</blockquote></details>

Thanks!

papeh (approved):

PR #1524, opened 2026-06-30
Fix BadImageFormatException loading Interop.WIA.dll in 64-bit apps

mark-sil:
## Problem

`build/SIL.Windows.Forms.targets` (shipped in the package's `build/` folder and
auto-imported by consumers) unconditionally copied the 32-bit
`build/Interop.WIA.dll` to every consumer's output directory and explicitly
excluded the AnyCPU `build/x64/` variant.

The 32-bit interop is built with the CorFlags `32BITREQUIRED` bit, so it cannot
load in a 64-bit process. Any 64-bit consumer that exercised the ImageToolbox
scanner/camera support (`AcquireImageControl.GetFromDevice`) crashed with
`System.BadImageFormatException` ("An attempt was made to load a program with an
incorrect format").

## Fix

Make the copy platform-aware:

- **x86** builds get the 32-bit interop from `build/`.
- **x64 / AnyCPU / unspecified** builds get the AnyCPU interop from `build/x64/`.

The AnyCPU assembly loads correctly in both 32- and 64-bit processes.

## Testing

- Verified the deployed `Interop.WIA.dll` is the AnyCPU build (CorFlags `0x09`,
  no `32BITREQUIRED`) in a 64-bit consumer's output.
- Confirmed in a downstream 64-bit application (FieldWorks) that the scanner/camera
  buttons in the ImageToolbox no longer throw `BadImageFormatException`.

## Notes

- Found via FieldWorks LT-22458 (crash clicking Camera/Scanner in the Picture
  Properties dialog of the 64-bit FieldWorks build).
- CHANGELOG.md updated under `[Unreleased]` → `Fixed`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1524)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- [SIL.Windows.Forms] Fixed `Interop.WIA.dll` deployment to provide the AnyCPU build to 64-bit (x64/AnyCPU) consumers instead of the 32-bit-only build.
 - [SIL.Core.ClearShare] Fixed `MetadataCore.GetSummaryParagraph` appending the Creator line twice and conditionally appending `RightsStatement` twice when using `CustomLicenseInfo`.
 - [SIL.Core] Fixed FontAnalytics exception handling to preserve original stack traces in debug mode.
 - [SIL.Windows.Forms.WritingSystems] WritingSystemFromWindowsLocaleProvider no longer crashes when an installed input language has a corrupt Windows installation; it now skips languages whose keyboard layout name cannot be read instead of throwing.
--- a/SIL.Windows.Forms/SIL.Windows.Forms.csproj
+++ b/SIL.Windows.Forms/SIL.Windows.Forms.csproj
@@ -49,12 +49,12 @@
 
   <ItemGroup>
     <None Include="../lib/Interop.WIA.dll">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Pack>true</Pack>
       <PackagePath>build</PackagePath>
     </None>
     <None Include="../lib/x64/Interop.WIA.dll">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <Link>Interop.WIA.dll</Link>
       <Pack>true</Pack>
       <PackagePath>build/x64</PackagePath>
     </None>
--- a/SIL.Windows.Forms/SIL.Windows.Forms.targets
+++ b/SIL.Windows.Forms/SIL.Windows.Forms.targets
@@ -1,7 +1,13 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup>
+  <ItemGroup Condition="'$(Platform)' == 'x86'">
     <None Include="$(MSBuildThisFileDirectory)**\*.dll"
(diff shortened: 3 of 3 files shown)


imnasnainaec:
A couple [Devin](https://app.devin.ai/review/sillsdev/libpalaso/pull/1524) flags:

### SIL.Windows.Forms/SIL.Windows.Forms.targets:9
**Both Interop.WIA.dll variants are still copied to output during library development builds**
The `SIL.Windows.Forms.csproj:51-60` unconditionally copies both `../lib/Interop.WIA.dll` and `../lib/x64/Interop.WIA.dll` to the output directory during development builds of the library itself (both have `CopyToOutputDirectory=Always`). The `.targets` file only governs NuGet package consumers. During local development, both DLLs end up in the output — the 32-bit one at the root and the x64 one in an `x64/` subdirectory. The runtime will load whichever is in the root (the 32-bit one), which means local dev/test of the library on a 64-bit process could still encounter the same `BadImageFormatException` this PR aims to fix for package consumers.

---

### SIL.Windows.Forms/SIL.Windows.Forms.targets:9
**Platform condition relies on solution-level $(Platform), not $(PlatformTarget)**
The condition `'$(Platform)' != 'x86'` checks the solution-level platform, not the project-level `$(PlatformTarget)`. A consumer could theoretically have `<Platform>AnyCPU</Platform>` with `<PlatformTarget>x86</PlatformTarget>` (the old 'Prefer 32-bit' scenario in .NET Framework). In that case, the AnyCPU interop DLL would be deployed. Per the commit message, the AnyCPU assembly is stated to work in both 32-bit and 64-bit processes, so this should be fine in practice — but it's worth being aware of this distinction if that claim about the AnyCPU DLL ever changes.

mark-sil:
Thanks for the review. Addressing both Devin flags:

### `SIL.Windows.Forms.targets:9` — both Interop.WIA.dll variants copied during library development builds

Good catch on the smell here. I checked the actual build output and it was resolving to the AnyCPU DLL, not the 32-bit one — both `<None>` items copied `Interop.WIA.dll` to the same output-root path with `CopyToOutputDirectory=Always`, and the `x64` entry (declared second) overwrote the 32-bit one. So it happened to work, but only by item order, which is fragile.

Fixed in 4990b148: the library now copies only the AnyCPU `lib/x64/Interop.WIA.dll` to its own build output (pinned to the root via `<Link>`), while both variants stay packed into the NuGet (`build/` and `build/x64/`) for consumers. Verified `output/Debug/{net462,net48,net8.0-windows}/Interop.WIA.dll` are all AnyCPU now, with no `x64/` collision.

### `SIL.Windows.Forms.targets:9` — platform condition relies on solution-level `$(Platform)`

Agreed this is safe in practice. The `x64` variant is an AnyCPU COM interop assembly (CorFlags `0x09`, no `32BITREQUIRED`) — it's metadata-only, so it loads correctly in both 32- and 64-bit processes. An `AnyCPU` + `PlatformTarget=x86` ("Prefer 32-bit") consumer getting the AnyCPU interop is therefore harmless, so I've kept the condition on `$(Platform)`. Thanks for flagging the distinction.

aror92 (approved):
:lgtm:

> @aror92 reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1524#-:-OwSyjSg6uuLFdtQMa_k:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mark-sil).

Issue #1275, opened 2023-07-13
Crash in ImageToolbox switching between Crop and Choose


papeh:
* Run the Palaso WinForms Test App
* Click Image Toolbox
* Choose an image from a File (happens for most but not all)
* Click Crop
* Click Choose
* Click Crop
Expected: the same image is available to crop
Actual: crash

The crash in the Palaso WinForms Test App is slightly different from the crash in FLEx ([LT-21467](https://jira.sil.org/browse/LT-21467))

PR #1528, opened 2026-07-10
[SIL.Windows.Forms] Fix ImageCropper resource leaks and cropping correctness bugs

imnasnainaec:
Related to #1275 (does **not** fully resolve it — the tab-switch crash fix is being deferred to a separate PR). This PR lands a set of related `ImageCropper` resource, correctness, and robustness fixes found while investigating that crash.

## Fixes

- **`Image` setter leaked the previous temp file and cropping image on re-set.** Old resources are now disposed and the fields nulled, so re-setting `Image` no longer leaks and no longer leaves disposed-but-non-null references.
- **Tall images were never downscaled before cropping.** The size check read `Width` twice; the second condition now correctly checks `Height`.
- **`GetCroppedImage` threw `NullReferenceException` when `Image` was set directly** (not via `SetImage`). `_originalFormat` is now initialized in the `Image` setter, and the redundant assignment in `SetImage` was removed.
- **JPEG re-encode hardening in `GetCroppedImage`.** The re-encode `MemoryStream` is reset (`Position = 0`) before reading, disposed via `using`, and the cropped bitmap is disposed on the failure path. The result cast was tightened to `(Bitmap)`.
- **`CalculateSourceImageArea` guarded against a null `_croppingImage`.**
- **`ImageCropper` now unsubscribes from `Application.Idle` on dispose.** This fixes a handler leak (the handler could otherwise fire on a disposed cropper). Note: this is a related safety fix and is not, on its own, confirmed to resolve the #1275 crash.

## Tests

Added `ImageCropperTests` covering:
- Double `Dispose` does not throw.
- Cropping a PNG returns a usable bitmap that re-encodes cleanly.
- Cropping a JPEG set directly via the `Image` property returns a JPEG bitmap that re-encodes cleanly (regression guard for the disposed re-encode stream).
- Re-setting the image via `SetImage` updates the format and does not throw.

---
Devin review: https://app.devin.ai/review/sillsdev/libpalaso/pull/1528

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1528)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms] Updated ImageToolbox UI to consistently use "image" (not "picture").
 - [SIL.Windows.Forms.Keyboarding] Removed Timer-based deferred IME conversion status restore from WindowsKeyboardSwitchingAdapter, which disrupted active Chinese Pinyin IME compositions (LT-22442). Added diagnostic tracing for keyboard switching and IME state.
 - [SIL.DictionaryServices] Fix memory leak in LiftWriter
-- [SIL.Windows.Forms] Fixed ImageCropper crash when switching between Crop and Choose tabs: `Application.Idle` handler was never unsubscribed, causing it to fire on a disposed object
+- [SIL.Windows.Forms] ImageCropper now unsubscribes from `Application.Idle` when disposed, fixing a handler leak that could fire on a disposed cropper
 - [SIL.Windows.Forms] Fixed ImageCropper `Image` setter leaking the previous temp file and cropping image on re-set; fields are now nulled after disposal so a mid-setter failure does not leave disposed-but-non-null references
 - [SIL.Windows.Forms] Fixed ImageCropper not downscaling tall images before cropping (height condition was checking width)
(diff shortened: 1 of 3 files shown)


imnasnainaec:
Closing: everything here is now either on master or covered better elsewhere.

- **`Image` setter leak** — fixed in #1536, which builds the new state into locals rather than disposing and nulling up front. That leaves no window where `_croppingImage` is null, and a failed re-set keeps the previous image intact.
- **`CalculateSourceImageArea` null guard** — needed here only because of the dispose-and-null ordering above. Under #1536 the field is never transiently null, so the guard has nothing to guard.
- **Tall-image downscale** — landed as #1531.
- **`Application.Idle` unsubscribe** — landed as #1532.
- **`_originalFormat`** (initializing it in the setter, removing it from `SetImage`) — superseded by #1530, which deletes the field outright along with the JPEG re-encode that reads it.
- **JPEG re-encode hardening** (`stream.Position = 0`, `(Bitmap)` cast) — this keeps the `using` on the `MemoryStream`, so the returned bitmap still refers to a stream that has been disposed. That is the #1275 crash. #1530 removes the re-encode instead of hardening it.

The tests here are covered too: double-dispose and the PNG crop are on master or in #1530, and `SetImage_Reassign_UpdatesFormatAndDoesNotThrow` asserts on `_originalFormat`, which #1530 removes.

Reopen if I've missed something.

PR #1530, opened 2026-07-13
+semver:major Fix unusable ImageCropper crop; RawFormat now MemoryBmp

### LINKED ISSUE #1275: Crash in ImageToolbox switching between Crop and Choose

imnasnainaec:
> [!WARNING]
> **Breaking change.** `GetCroppedImage` no longer returns a JPEG-encoded bitmap for JPEG sources. A real crop now has `RawFormat` `MemoryBmp`; a whole-image selection has `Png`, since the crop is read back from the cropper's PNG temp file. Callers that read `RawFormat`, or that call `Image.Save(path)` on the result and rely on GDI+ picking the JPEG encoder implicitly, must now pass an explicit `ImageFormat` — a `.jpg` path would otherwise receive PNG bytes. Going through `PalasoImage.Save(path)` is unaffected, since it picks the encoder from the file extension. ImageToolbox's own use goes through `PalasoImage`, so nothing in this repo changes behaviour.
>
> `GetImage` is unaffected when nothing was cropped: it now returns the image untouched, so an uncropped JPEG stays a JPEG.
>
> (`.RawFormat` is used in mono: https://github.com/search?q=org%3Asillsdev+.rawformat&type=code)

`GetCroppedImage` re-encodes a JPEG crop through a `MemoryStream` and returns what `Image.FromStream` produces. That stream sits in a `using`, so it is disposed before the bitmap is returned, and GDI+ decodes lazily — the crop refers to a stream that is already gone. The next thing to touch its pixels fails with "A generic error occurred in GDI+".

That is #1275: choose an image from a file, Crop, Choose, Crop. Leaving the Crop tab calls `GetImage`, which stores the stream-backed crop on the `PalasoImage`; returning re-enters the `Image` setter, which calls `value.Image.Save(...)` on that crop and throws. Only JPEG sources take the re-encode path, matching the report's "happens for most but not all".

### The change

Return the cropped bitmap directly instead of the re-encoded one, so nothing holds a lazy reference to a stream. Preserving the JPEG `RawFormat` is what required the re-encode in the first place, so giving it up is what makes the fix possible — hence the breaking change above.

`GetImage` now skips the crop entirely when no grip has moved, returning the `PalasoImage` untouched. That is the common path — open the Crop tab, leave it — and it used to replace the image with a copy of itself round-tripped through the PNG temp file, costing the original's format and bit depth and leaving that temp file locked until the copy was disposed.

With the re-encode gone `_originalFormat` has no remaining use, so it goes too, along with the `Require.That` self-check, the `SIL.Code` import, and the comments describing the old workaround.

Removing `_originalFormat` fixes a second bug on the way. It was assigned only in `SetImage`, so assigning the `Image` property directly left it null, and `GetCroppedImage` threw a `NullReferenceException` reading `_originalFormat.Guid` — swallowed by the catch and surfaced as "Sorry, there was a problem getting the image".

`ShowToolboxWith_PreExisting_EnsureRawFormatUnchanged` is deleted: it is `[Explicit("By hand only")]` so it never ran in CI, and it asserted exactly the JPEG `RawFormat` this change gives up.

### Tests

Five of the added cases fail against `master`:

- `GetCroppedImage_JpegImage_ReturnsUsableBitmap(true)` and `(false)` — the crop is backed by a disposed stream.
- `GetCroppedImage_ImageSetViaPropertyDirectly_ReturnsUsableBitmap` — the null `_originalFormat` NRE.
- `GetImage_NothingCropped_ReturnsImageUntouched` — the image comes back re-encoded.
- `GetImage_ReCropPreviouslyCroppedJpeg_DoesNotThrow` — the #1275 round trip.

`GetImage_ResultOutlivesCropper_SavedOriginalTempFileIsDeleted` pins the temp file against being held open by the result, and `GetCroppedImage_PngImage_ReturnsUsableBitmap` and `GetCroppedImage_OneBitPng_PreservesPixelFormat` cover what the change must not disturb.

`--filter ImageToolbox` is green on net8.0-windows and net48: 56 passed, 1 pre-existing skip.

Fixes #1275

---

Devin review: https://app.devin.ai/review/sillsdev/libpalaso/pull/1530

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1530)


imnasnainaec:
**Claude summary of a Devin flag**

In the `Image` setter (ImageCropper.cs:146-149), the old `_savedOriginalImage` and `_croppingImage` are disposed and nulled before `value.Image.Save(...)` runs. If that save throws, `_image` ends up pointing at a PalasoImage whose backing temp file and cropping bitmap are already gone — leaving the object in an inconsistent state.

**Claude rebuttal**

Reaching that state requires `Save` to throw mid-setter (out-of-memory, disk full, permissions) — already an unrecoverable condition for this control. The only consumer of that state, `GetCroppedImage()`, wraps its work in try/catch and returns `null` via `ErrorReport.NotifyUserOfProblem` on any failure, so a corrupted `_savedOriginalImage` just hits that existing safety net instead of crashing. Reordering to save-before-dispose would mean holding both the old and new temp file/cropping bitmap simultaneously — added complexity for an edge case that already degrades safely, and one that's strictly better than the prior behavior of leaking the old resources unconditionally on every re-set.

imnasnainaec:
@andrew-polk What's your take on the breaking change of this pr? (Don't worry about doing code review... not worthwhile unless and until we decide it's a worthwhile change.)

andrew-polk:
@imnasnainaec 
Sorry; I never saw this until now. I've gotten into the habit of deleting notifications because I've been inundated with them...

Anyway, for Bloom, I don't think this affects us because 6.4 is the last version to use the Image Toolbox from libpalaso (we are replacing it with a React one in 6.5. And hopefully 6.4 will not be getting any more updates from libpalaso.

This bug is certainly real and we even track it, but decided not to fix it.
https://issues.bloomlibrary.org/youtrack/issue/BL-15019

But you should probably make the judgement about the breaking change based on Flex, not Bloom.

Happy to discuss further if helpful.

SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
```diff
@@ -440,21 +435,10 @@ public Image GetCroppedImage()
...
-							cropped = System.Drawing.Image.FromStream(stream) as Bitmap;
-							oldCropped.Dispose();
-							Require.That(ImageFormat.Jpeg.Guid == cropped.RawFormat.Guid, "lost jpeg formatting");
-						}
-					}
-					return cropped;
+					// Clone already copies the pixels out, so the crop outlives originalImage and its
+					// temp file. Don't copy it into a new Bitmap to "detach" it: that widens a 1-bit
+					// PNG to 32bpp (BL-2841) and buys nothing.
+					return originalImage.Clone(selection, originalImage.PixelFormat);
```
jasonleenaylor, line 456: It looks like `Clone(selection, PixelFormat)` only copies the pixels out when `selection` is smaller than the bitmap. When it is the full bounds, which is what unmoved grips produce (lines 432-435 snap to the edge), GDI+ returns a bitmap that still shares `originalImage`'s file-backed data: `RawFormat` is `Png`, the temp file stays locked, and `DisposeImageState` swallows the `IOException` and leaks it. A real crop gives `MemoryBmp`, detached. I think the test `GetCroppedImage_JpegImage_ReturnsUsableBitmap` sees `Png` only because it never moves a grip.

If this is correct maybe update the comments, changelog wording and Test as follows:
- This comment: "Clone copies the pixels out for a partial selection, so a real crop outlives originalImage and its temp file. A full-bounds clone still shares the file-backed source (RawFormat Png, temp file locked until it is disposed). Don't copy into a new Bitmap to detach it: that widens a 1-bit PNG to 32bpp (BL-2841)."
- CHANGELOG: replace "Its `RawFormat` is therefore `Png` (the format of the temp file the cropper crops from) rather than `Jpeg`" with "Its `RawFormat` is no longer `Jpeg`: `MemoryBmp` for a crop, `Png` when the selection is the whole image".
- Test: move a grip before asserting, or assert `Is.Not.EqualTo(ImageFormat.Jpeg.Guid)`.

jasonleenaylor (approved):
From the FieldWorks side, we can take this. FLEx's only consumer of the toolbox is the Picture Properties dialog, which detects a crop by comparing image sizes and saves through `PalasoImage.Save(path)`, so the encoder comes from the file extension and nothing reads `RawFormat`. FieldWorks is already on the 18.0.0 beta line, so the major bump costs us nothing extra. Today a FLEx user who crops a JPEG and re-enters the Crop tab gets the toolbox error dialog, and this fixes that.

One question inline about the comment and CHANGELOG wording.

_This review was assisted by Claude Fable 5.1._

SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
imnasnainaec: Non-crop is now a proper no-op.

Comments and tests updated.

jasonleenaylor (approved):
> @jasonleenaylor reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1530)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #1532, opened 2026-07-13
Unsubscribe ImageCropper's Application.Idle handler on Dispose

imnasnainaec:
ImageCropper subscribed to the static `Application.Idle` event in its constructor but never unsubscribed in `Dispose`, so the handler could still fire after the control was disposed.

Tests: `Dispose_CalledTwiceAfterSettingImage_DoesNotThrow` is a smoke test, and `Dispose_AllowsGarbageCollection` is a regression test that asserts the control is garbage-collected after disposal — it fails without the unsubscribe fix and passes with it.

Devin review: https://app.devin.ai/review/sillsdev/libpalaso/pull/1532/

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1532)

### DIFF
--- a/SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
+++ b/SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
@@ -482,6 +482,8 @@ namespace SIL.Windows.Forms.ImageToolbox.Cropping
 					components = null;
 				}
 
+				Application.Idle -= Application_Idle;
+
 				try
 				{
 					if (_savedOriginalImage != null)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms] Updated ImageToolbox UI to consistently use "image" (not "picture").
 - [SIL.Windows.Forms.Keyboarding] Removed Timer-based deferred IME conversion status restore from WindowsKeyboardSwitchingAdapter, which disrupted active Chinese Pinyin IME compositions (LT-22442). Added diagnostic tracing for keyboard switching and IME state.
 - [SIL.DictionaryServices] Fix memory leak in LiftWriter
+- [SIL.Windows.Forms] Fixed ImageCropper crash caused by its `Application.Idle` handler never being unsubscribed, so it could fire on a disposed instance
 - [SIL.WritingSystems] Fix IetfLanguageTag.GetGeneralCode to handle cases when zh-CN or zh-TW is a prefix and not the whole string.
 - [SIL.WritingSystems] More fixes to consistently use 繁体中文 and 简体中文 for Traditional and Simplified Chinese native language names, and Chinese (Traditional) and Chinese (Simplified) for their English names.
 - [SIL.Windows.Forms] Prevent BetterLabel from responding to OnTextChanged when it has been disposed.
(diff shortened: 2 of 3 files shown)


SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
```diff
@@ -482,6 +482,8 @@ protected override void Dispose(bool disposing)
 					components = null;
 				}
 
+				Application.Idle -= Application_Idle;
```
hahn-kev, line 485: I'm not actually sure how to test this or if this actually does what we want because the subscribe line is `Application.Idle += new EventHandler(Application_Idle);`, There's not really good options to assert this test without more changes. I would probably write a test like this which ensures that the `ImageCropper` gets cleaned up as expected (it would not before this bug fix). 

```C#
[Test]
public void Component_DoesNotLeakMemory_AfterDisposal()
{
    // Arrange & Act
    // We isolate creation/destruction to a separate method to ensure 
    // local strong references go out of scope cleanly before we check GC.
    WeakReference componentReference = CreateAndDisposeComponent();

    // Force full Garbage Collection
    GC.Collect();
    GC.WaitForPendingFinalizers();
    GC.Collect();

    // Assert
    // If it unsubscribed, it should be garbage collected. If it's still alive,
    // the static Application.Idle event is holding a leak path.
    Assert.False(componentReference.IsAlive, 
        "Memory leak detected: Component failed to unsubscribe from Application.Idle.");
}

private WeakReference CreateAndDisposeComponent()
{
    var component = new MyGuiComponent(); // Leverages raw Application.Idle internally
    
    // Simulate the cleanup flow
    component.Dispose(); 
    
    return new WeakReference(component);
}
```

hahn-kev (changes requested):

SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
imnasnainaec: @hahn-kev I added the tests, but Claude and Devin are both concerned of its flakiness

Claude:
> The test uses a WeakReference + forced GC to assert the disposed ImageCropper gets collected. It can flake mainly because GC.Collect() is a hint, not a guarantee — collection timing varies with GC mode, tiered JIT, and Debug-vs-Release local liveness (though your NoInlining helper largely handles the last one), so a genuinely-unrooted instance may occasionally still read IsAlive on some runs or CI configs. It's also imprecise rather than flaky in one respect: it asserts the instance is fully unrooted, so any other hidden root (a lingering WinForms handle registration or a different unremoved static-event subscription) would turn it red for a reason unrelated to the Application.Idle fix. Finalizer timing (Control has one) and untested TFMs (net462/net48) round out the minor risks. If CI ever flakes, retry the collect in a short bounded loop rather than a single pass.

Devin:
> The `Dispose_AllowsGarbageCollection` test asserts the ImageCropper is collected after Dispose. WinForms controls have other roots/finalization paths, and GC timing is non-deterministic. It is not a reliable regression guard for the Application.Idle leak, so future regressions could go undetected.

hahn-kev: yeah that makes sense. Let's leave a note on the test to indicate that it may be flaky and it can be dropped if that turns out to be true, otherwise it at least proved what we wanted for now.

hahn-kev (approved):

SIL.Windows.Forms/ImageToolbox/Cropping/ImageCropper.cs
imnasnainaec: Done

imnasnainaec (commented):
> @imnasnainaec reviewed 3 files and all commit messages, made 1 comment, resolved 1 discussion, and dismissed @hahn-kev from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1532)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #1535, opened 2026-08-12
Fix Replace failing forever after an interrupted writing system update

johnml1135:
**One interrupted writing system update breaks that writing system permanently.** Nothing recovers on its own, and nothing is reported.

`GlobalWritingSystemRepository.Replace` copies the `.ldml` file aside — as `<tag>.ldml.localrepoupdate` — so the old content is still on disk when `Save()` round-trips unknown LDML elements out of it. `WsStasher` makes that copy; its `Dispose` moves it back.

If a process dies part-way through, the copy is left behind, and every later `Replace` for that tag throws in the `WsStasher` **constructor**. That placement is what makes it stick: the `using` body never runs, so the writing system is never updated; `Dispose` never runs, so the debris is never cleared.

It is also slow and silent: `RobustFile` retries ten times 200 ms apart and `OnChangeNotifySharedStore` swallows the exception, so it surfaced only as ~1.8 s added to *every* `LcmCache.Dispose()`, with nothing logged.

## The fix

`base.Replace` removes the file before adding it back, so an interruption leaves one of two states:

1. **`.ldml` still present**, copy stale → copy with `overwrite: true`, making it self-correcting.
2. **`.ldml` already deleted**, copy is the only surviving version → move it back, exactly what the interrupted `Dispose` would have done.

A third route to the same stall: `Replace` on a never-saved writing system had no file to copy. Nothing to preserve, so it no longer tries.

## Tests

Three tests, one per case, asserting the replacement took effect and no debris remains. Each fails before and passes after; 28 green on `net8.0` and `net48`.

## Caveat

`Replace` is the only mutating override here that doesn't take `_mutex.Lock()`. This change makes one narrow concurrent case worse: two racing stashers previously collided harmlessly (the second threw in its constructor, touching nothing); now the second can clobber the first's stash, leaving the first's `Dispose` to fail. No permanent debris. Flagging rather than changing locking semantics here.

https://claude.ai/code/session_01R8yM1naBeYePxhPBgtLj2p

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1535)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.Windows.Forms.Archiving] Fixed formatting of message in ArchivingDlg so that the name of the auxiliary archive upload program (e.g., "RAMP") is displayed.
 - [SIL.Windows.Forms] Fixed SettingsLauncherButton never disposing the SettingsProtectionHelper it creates, which left an enabled timer running after the button was disposed. Also removed the button's own unused visibility timer, which had no handler but was posting timer messages for the life of the control.
 - [SIL.Windows.Forms] Fixed `SettingsProtectionHelper.Dispose` so that it only touches managed resources when disposing, is safe to call more than once, and no longer disposes the Ctrl+Shift timer explicitly (the timer is owned by the component container that disposes it).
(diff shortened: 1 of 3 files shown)


imnasnainaec (approved):
> @imnasnainaec reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1535)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

imnasnainaec (approved):
> @imnasnainaec reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1535)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #1536, opened 2026-08-26
ImageCropper: dispose the state built for the previous image on re-set

imnasnainaec:
The `Image` setter holds two things for the image it is cropping: a temp file with the original at full resolution, and a possibly-downscaled bitmap to display. It mishandles both when given a new image.

It overwrites the two fields without disposing what was there, leaking a temp file and a bitmap on every re-set — the normal path, since re-entering the Crop tab re-sets `Image`. The leaked bitmap is a clone of a file-backed image, so it also holds a lock on the previous source file: disposing that `PalasoImage` trips `Debug.Fail("Not able to delete image temp file")` with `IOException: the process cannot access the file ... because it is being used by another process`.

It also assigns `_savedOriginalImage` before `value.Image.Save` can throw. If it does, the field is left pointing at the new, empty temp file while the cropper still displays the previous image, so `GetCroppedImage` crops from the empty file.

### The change

Build the temp file and the cropping bitmap into locals and store them only once both are ready — disposing the previous pair at that point, or the partial state if anything throws. The BL-2680-tolerant disposal is factored out of `Dispose` so the three call sites share it.

`SetImage` likewise commits `_originalFormat` only after the setter succeeds, so a failure no longer leaves it on the new image while everything else is on the old.

### Tests

Both fail without the change.

- `SetImage_Reassigned_DisposesStateBuiltForPreviousImage` — the previous temp file is still on disk after a re-set.
- `Image_NewImageFailsToLoad_LeavesPreviousImageStateIntact` — after a failed assignment the cropper still holds the same temp file and bitmap.

`--filter ImageToolbox` is green on net8.0-windows and net48: 47 passed, 1 pre-existing skip.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1536)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [SIL.DictionaryServices] Fix memory leak in LiftWriter
 - [SIL.Windows.Forms] Fixed ImageCropper crash caused by its `Application.Idle` handler never being unsubscribed, so it could fire on a disposed instance
 - [SIL.Windows.Forms] Fixed ImageCropper not downscaling tall images before cropping (height condition was checking width)
+- [SIL.Windows.Forms] Fixed ImageCropper leaking the temp file and cropping bitmap built for the previous image each time it is given a new one to crop
 - [SIL.WritingSystems] Fix IetfLanguageTag.GetGeneralCode to handle cases when zh-CN or zh-TW is a prefix and not the whole string.
 - [SIL.WritingSystems] More fixes to consistently use 繁体中文 and 简体中文 for Traditional and Simplified Chinese native language names, and Chinese (Traditional) and Chinese (Simplified) for their English names.
 - [SIL.Windows.Forms] Prevent BetterLabel from responding to OnTextChanged when it has been disposed.
--- a/SIL.Windows.Forms.Tests/ImageToolbox/ImageCropperTests.cs
+++ b/SIL.Windows.Forms.Tests/ImageToolbox/ImageCropperTests.cs
@@ -1,6 +1,7 @@
-using System;
+﻿using System;
 using System.Drawing;
 using System.Drawing.Imaging;
+using System.IO;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Threading;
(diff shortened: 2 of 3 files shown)


tombogle (changes requested):
> @tombogle reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1536)* status: all files reviewed, 1 unresolved discussion (waiting on imnasnainaec and tombogle).

___
*[`SIL.Windows.Forms.Tests/ImageToolbox/ImageCropperTests.cs` line 202 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1536#-P02dUWqB_oHZE8o9HBC:-P02dUWrCYh1PPkO1pFQ:bgj5laz) ([raw file](https://github.com/sillsdev/libpalaso/blob/3a339bf6c29abbbdf2ee18f3fd28c796864403a9/SIL.Windows.Forms.Tests/ImageToolbox/ImageCropperTests.cs#L202)):*
> ```Smalltalk
> 		}
> 
> 		private static object GetPrivateField(ImageCropper cropper, string fieldName)
> ```

I think this could use the existing ReflectionHelper.GetField rather than recreating it.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1536)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on tombogle).

___
*[`SIL.Windows.Forms.Tests/ImageToolbox/ImageCropperTests.cs` line 202 at r2](https://reviewable.io/reviews/sillsdev/libpalaso/1536#-P02dUWqB_oHZE8o9HBC:-P02qtCk-RFv4muHzEiC:b4855pb) ([raw file](https://github.com/sillsdev/libpalaso/blob/3a339bf6c29abbbdf2ee18f3fd28c796864403a9/SIL.Windows.Forms.Tests/ImageToolbox/ImageCropperTests.cs#L202)):*

I think this could use the existing ReflectionHelper.GetField rather than recreating it.
</blockquote></details>

Thanks! Done.

tombogle (commented):
> @tombogle reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1536)* status: all files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

tombogle (approved):
> @tombogle resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1536)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

Issue #1516, opened 2026-06-18
Flaky tests on CI


imnasnainaec:
- [x] Play_InvalidAudioFileThrowsBackgroundException_NonFatalErrorReported (this one fails for me pretty frequently) — #1537
- [x] PlayAndStopPlaying_NonWindows_DoesNotThrow("wav") (this one's failure is pretty rare for me) — #1539
- [x] HotSpotMouseClick_MouseClickOverHotSpot_FiresMouseClick (https://github.com/sillsdev/libpalaso/actions/runs/33097992679/job/98607862662?pr=1537#step:14:1787) — #1540
- [x] TestNAudioWaveOutEvent raises PlaybackStopped while holding its own lock, which can deadlock against the lock WindowsAudioSession holds across Stop — #1538

imnasnainaec:
For example, the first 2 failed here: https://github.com/sillsdev/libpalaso/actions/runs/28372918756/job/84055017433?pr=1520#step:14:1653

PR #1538, opened 2026-08-27
Raise TestNAudioWaveOutEvent.PlaybackStopped outside its lock

imnasnainaec:
`MediaPlaybackStopped` raised `PlaybackStopped` while holding this class's lock.
The handler takes `WindowsAudioSession`'s lock, and `StopPlaying` holds that lock
across its call to `Stop()`, so the two could be acquired in opposite orders and
deadlock. Raise the event after releasing the lock.

The state transition to `Stopped` stays inside the lock, so the event is still
raised exactly once.

CI failure: none known. This deadlock has not been observed; it would show up as a
hang rather than a failure, and would be caught by `--blame-hang-timeout`.

Part of #1516

Devin review: https://app.devin.ai/review/sillsdev/libpalaso/pull/1538

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/libpalaso/1538)

### DIFF
--- a/SIL.Media.Tests/AudioSessionTests.cs
+++ b/SIL.Media.Tests/AudioSessionTests.cs
@@ -380,8 +380,16 @@ namespace SIL.Media.Tests
 			// at least it has been queued to start.
 			Assert.That(session.IsPlaying, Is.True);
 			Assert.DoesNotThrow(() => session.StopPlaying());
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
LGTM

_This review was assisted by Claude Opus 5._

